PR #28, opened 2019-07-26
Fix deployment and Github urls

marcschier:
* Replace all old repo urls and submodule repo urls with new url in properties and documentation
* Fix arm template to work
* Temporarily pointing to dockerhub until mcr images are available
* Remove demo deployment for the time being


docker-compose.yml
```diff
@@ -31,22 +31,8 @@ services:
...
-      - PCS_IOTHUB_CONNSTRING
-      - PCS_IOTHUBREACT_HUB_ENDPOINT
-      - PCS_IOTHUBREACT_HUB_PARTITIONS
-      - PCS_IOTHUBREACT_HUB_NAME
-      - PCS_IOTHUBREACT_HUB_CONSUMERGROUP
-      - PCS_IOTHUBREACT_AZUREBLOB_ACCOUNT
-      - PCS_IOTHUBREACT_AZUREBLOB_KEY
-      - PCS_IOTHUBREACT_AZUREBLOB_ENDPOINT_SUFFIX
+      - PCS_APPINSIGHTS_NAME
+      - PCS_APPINSIGHTS_INSTRUMENTATIONKEY
```
karok2m: I do not see any changes regarding  `PCS_APPINSIGHTS_NAME` and `PCS_APPINSIGHTS_INSTRUMENTATIONKEY` variables outside of `docker-compose.yml` files. Are the variables dumped to `.env` file by deploy script ?

marcschier: Correct.  These are dumped into .env by setup.sh which runs inside on creation of the VM.  And we forgot to update the docker-compose.yml when it was added recently to make use of these.

PR #30, opened 2019-07-29
[opctwin] - fixes to  run as windows edge-module

cristipogacean:
* fixed the connection string for endpoints twin management using the Iot EdgeHub 
* fixed the unix socket based (unix://..) edgelet communication on windows

### DIFF
--- a/common/src/Microsoft.Azure.IIoT.Core/src/Extensions/UriEx.cs
+++ b/common/src/Microsoft.Azure.IIoT.Core/src/Extensions/UriEx.cs
@@ -90,6 +90,10 @@ namespace System {
                 fileDevice = localPath.Substring(0, index + 5);
                 pathAndQuery = localPath.Substring(index + 5);
             }
+            else if ((index = localPath.IndexOf("sock", StringComparison.InvariantCultureIgnoreCase)) != -1) {
+                fileDevice = localPath.Substring(0, index + 4);
+                pathAndQuery = localPath.Substring(index + 4);
+            }
             else {
                 // Find fake port delimiter
                 index = localPath.IndexOf(':');
--- a/common/src/Microsoft.Azure.IIoT.Core/src/Extensions/UriEx.cs
+++ b/common/src/Microsoft.Azure.IIoT.Core/src/Extensions/UriEx.cs
@@ -85,12 +85,8 @@ namespace System {
             };
             string fileDevice;
             string pathAndQuery;
-            var index = localPath.IndexOf(".sock", StringComparison.InvariantCultureIgnoreCase);
-            if (index != -1) {
-                fileDevice = localPath.Substring(0, index + 5);
-                pathAndQuery = localPath.Substring(index + 5);
-            }
-            else if ((index = localPath.IndexOf("sock", StringComparison.InvariantCultureIgnoreCase)) != -1) {
+            var index = localPath.IndexOf("sock", StringComparison.InvariantCultureIgnoreCase);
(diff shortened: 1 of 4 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge/src/Supervisor/SupervisorServices.cs
```diff
@@ -339,8 +339,13 @@ private class TwinHost : IDisposable, IProcessControl, IModuleConfig {
                         throw new InvalidConfigurationException(
                             "Missing IOTEDGE_IOTHUBHOSTNAME variable in environment");
                     }
-                    cs = $"HostName={hostName};DeviceId={endpointId};SharedAccessKey={secret}";
                     var edgeName = Environment.GetEnvironmentVariable("IOTEDGE_GATEWAYHOSTNAME");
+                    if (string.IsNullOrEmpty(edgeName)) {
```
mregen: isn't the if reversed here? edgeName != null etcetc

common/src/Microsoft.Azure.IIoT.Core/src/Extensions/UriEx.cs
```diff
@@ -90,6 +90,10 @@ public static class UriEx {
                 fileDevice = localPath.Substring(0, index + 5);
                 pathAndQuery = localPath.Substring(index + 5);
             }
+            else if ((index = localPath.IndexOf("sock", StringComparison.InvariantCultureIgnoreCase)) != -1) {
```
mregen: better here to define strings (e.g. ".sock")as consts and derive the index modifier from constants

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge/src/Supervisor/SupervisorServices.cs
```diff
@@ -353,7 +358,14 @@ private class TwinHost : IDisposable, IProcessControl, IModuleConfig {
                         throw new InvalidConfigurationException(
                             "Missing HostName in connection string");
                     }
-                    cs = $"HostName={hostName};DeviceId={endpointId};SharedAccessKey={secret}";
+
+                    if (!lookup.TryGetValue("GatewayHostName", out var edgeName) ||
```
mregen: same as above, reversed logic?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge/src/Supervisor/SupervisorServices.cs
cristipogacean: indeed, that is fixed now

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge/src/Supervisor/SupervisorServices.cs
cristipogacean: indeed, this is fixed now

common/src/Microsoft.Azure.IIoT.Core/src/Extensions/UriEx.cs
cristipogacean: the "socket syntax is in linux is ".sock" and in windows is "/sock", therefore the "sock" check is enough for both, not that strict nevertheless...

PR #31, opened 2019-07-29
fix for access policies 

monikavar:
image

**Deployment failed because of above error.** 

Reason:
For AccessPolicy - The Azure Active Directory tenant ID that should be used for authenticating requests to the key vault.


services/deploy/vm/template.json
```diff
@@ -520,7 +521,139 @@
                     "family": "A"
                 },
                 "accessPolicies": []
+<<<<<<< .mine
```
marcschier: This still needs to be fixed

PR #62, opened 2019-08-05
deployment fixes

monikavar:
For deployment type "vm" and "local", automatically select tenantId from the subscription.

### DIFF
--- a/services/deploy/deploy.ps1
+++ b/services/deploy/deploy.ps1
@@ -362,90 +362,10 @@ Function AcquireToken() {
     return $authResult
 }
 
-
-#*******************************************************************************************************
-# Select azure ad tenant available to user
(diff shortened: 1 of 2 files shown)


services/deploy/deploy.ps1
```diff
@@ -437,7 +442,7 @@ Function SelectAzureADTenantId() {
 Function ConnectToAzureADTenant() {
     if ($script:interactive) {
         # Interactive selecting tenant to connect to
-        if (!$script:tenantId) {
+        if (!$script:tenantId -or $script:type -ne "vm") {
```
marcschier: Could you change this for "vm" also?  And then just remove the SelectAzureADTenantId function.   Just use the tenant from the subscription.

PR #105, opened 2019-09-16
Update publisher.md

khilscher:
Need to specify full path to publishednodes json file when using "Binds" otherwise it will attempt to find the file in /app in the container filesystem


barnstee:
How would this work on Linux?

sjai013:
E.g. /appdata/pn.json where /appdata is bound to a directory on the host system.

PR #134, opened 2019-10-10
enable correlation id in the logs

monikavar:


common/src/Microsoft.Azure.IIoT.Core/src/Diagnostics/Extensions/LogEx.cs
```diff
@@ -136,7 +136,7 @@ public static class LogEx {
         public static LoggerConfiguration ConsoleOut(this LoggerConfiguration configuration) {
             return configuration
                 .Enrich.FromLogContext()
-                .WriteTo.Console(outputTemplate: "{Message:lj}{NewLine}")
+                .WriteTo.Console(outputTemplate: "({XCorrelationId}): {Message:lj}{NewLine}")
```
marcschier: I think first best to do it for app insights logging - else the log is so cluttered.

marcschier: Also, it would always be null if not running in a service context, no?  
Could we add correlation for device methods also?

PR #151, opened 2019-11-15
IAI: Added note about resource cleanup on error.

karok2m:


docs/industrial_iot_deployment.md
```diff
@@ -98,7 +98,16 @@ We are working on simplifying this flow, but as of right now these are the steps
     SubscriptionId: XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX, DisplayName: Visual Studio Enterprise
     ```
 
-8. Select Resource Group. You would be presented with options of either using an existing Resource Group or creating a new one. If you choose to use an existing one, you will be presented with a list of all Resource Group within the Subscription to choose from. Otherwise, you would be asked to select a Region for the new Resource Group and provide a name for it. Output for the latter case is below:
+8. Select Resource Group. You would be presented with options of either using an existing Resource Group or
+creating a new one. If you choose to use an existing one, you will be presented with a list of all Resource
+Group within the Subscription to choose from. Otherwise, you would be asked to select a Region for the new
```
monikavar: Groups

docs/industrial_iot_deployment.md
```diff
@@ -98,7 +98,16 @@ We are working on simplifying this flow, but as of right now these are the steps
...
+Resource Group and provide a name for it. Output for the latter case is below.
+
+    > **Note:** If the application encounters an error during execution, it will ask you whether to perform a
+    cleanup or not. Cleanup works by deleting registered Applications and the Resource Group. This means,
+    that if an existing Resource Group has been selected for the deployment and error occurs, then selecting to
+    perform cleanup will also trigger deletion of all resources previously present in the Resource Group. More
```
monikavar: presented

karok2m: @monikavar, I actually think that 'present' is correct here. But I will change the sentence slightly.

docs/industrial_iot_deployment.md
karok2m: Fixed.

PR #166, opened 2019-12-09
IAI: unattended mode

karok2m:
* Introduced interface and extension class for AuthenticationManager.
* Changed "managed-by" tag to "managed_by". …
* Changed KeyVaultMgmtClient.GetCreationParameters() so that it also accepts Service Principal as owner.
* Changed TokenCredentials creation to work with null accounts.
* Minor refactoring of AuthenticationManager and MicrosoftGraphServiceClient.


deploy/src/Microsoft.Azure.IIoT.Deployment/Authentication/IAuthenticationManager.cs
```diff
@@ -0,0 +1,41 @@
...
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    using Microsoft.Azure.Management.ResourceManager.Fluent;
+    using Microsoft.Identity.Client;
+
+    public interface IAuthenticationManager {
+
+        public  Task AuthenticateAsync(
```
marcschier: Would be great to get the public interfaces doc'ed to some degree, at least for it does and how it is supposed to be used.

karok2m: Done.

PR #181, opened 2019-12-19
optimized CDM processor

cristipogacean:


common/src/Microsoft.Azure.IIoT.Cdm/src/Services/CdmMessageProcessor.cs
```diff
@@ -116,49 +127,101 @@ public class CdmMessageProcessor : ICdmClient {
...
+        private async void CacheTimer_ElapesedAsync(object sender) {
+            await PerformWriteCache();
+            Try.Op(() => _cacheUploadTimer.Change(_cacheUploadInterval, Timeout.InfiniteTimeSpan));
+        }
 
-            await _storage.WriteInCsvPartition<SubscriberCdmSampleModel>(
-                partitionLocation,
-                new List<SubscriberCdmSampleModel>() { payload },
-                csvTrait?.Arguments?.FetchValue("delimiter") ?? kCsvPartitionsDelimiter);
+        private async Task ProcessCdmSampleAsync(SubscriberCdmSampleModel payload) {
```
marcschier: Awesome - and ok for now - however - we must make this a full blown IDeviceTelemetryHandler, that way we can do the checkpointing after the batch write.   The current way we would loose the content of the cache if we go down causing data loss.

cristipogacean:
optimized CDM processor

PR #193, opened 2020-01-14
Release 2.6

marcschier:


mregen:
The diff you're trying to view is too large. We only load the first 3000 changed files...

Next time smaller junks please..

PR #205, opened 2020-01-21
getting data after page rendering

dacolgit:
add spinner when loading data
optimized read value when browsing
fix scan status
'about' button redirect to github Industrial-Iot page

### DIFF
--- a/samples/src/Microsoft.Azure.IIoT.App/src/Pages/Applications.razor
+++ b/samples/src/Microsoft.Azure.IIoT.App/src/Pages/Applications.razor
@@ -22,7 +22,7 @@
         </tr>
     </thead>
     <tbody>
-        @foreach (var application in pagedapplicationList.Results) {
+        @foreach (var application in _pagedapplicationList.Results) {
         <tr>
             <td>
                 <a href='endpoints/@RegistryHelper.PathAll/@application.ApplicationId'>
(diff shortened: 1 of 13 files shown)


samples/src/Microsoft.Azure.IIoT.App/src/Pages/Applications.razor
```diff
@@ -54,11 +56,27 @@
...
+    protected override void OnInitialized()
+    {
+        _spinner = "loader-big";
+    }
+
+    /// <summary>
+    /// OnAfterRenderAsync
+    /// </summary>
+    /// <param name="firstRender"></param>
+    protected override async Task OnAfterRenderAsync(bool firstRender) {
```
marcschier: Let's follow consistent formatting.

samples/src/Microsoft.Azure.IIoT.App/src/Pages/Discoverers.razor
```diff
@@ -151,50 +157,67 @@
...
+
+    /// <summary>
+    /// OnAfterRenderAsync
+    /// </summary>
+    /// <param name="firstRender"></param>
+    protected override async Task OnAfterRenderAsync(bool firstRender)
+    {
+        if (firstRender)
+        {
+            discovererList = await RegistryHelper.GetDiscovererListAsync();
```
marcschier: Let's use query with paging.

dacolgit: I'm already using query with paging

marcschier: So where are you calling the API to get the next page?   I can see you use query, but what I am missing is where you query again for the next page.   We should use the paging the API provides, not query all and then page the list, the former is far more efficient - also for all the other API, including browsing.

PR #225, opened 2020-02-03
Fix the kestrel warnings for 2.6

mregen:
- latest OPC stack with new nodeset
- stack now supports DataTypeDefinition on servers
- 2 files removed causing build errors and which had no dependencies
- not officially tagged as release yet -- testing only for early birds


GKAYED:
Can 2.6.0 be tested on a dev environment after this PR is completed?

PR #226, opened 2020-02-03
Fix for swagger

monikavar:


common/src/Microsoft.Azure.IIoT.AspNetCore.OpenApi/src/Extensions/ApplicationBuilderEx.cs
```diff
@@ -74,9 +74,9 @@ public static class ApplicationBuilderEx {
                 foreach (var info in infos) {
                     if (config.WithAuth) {
                         options.OAuthAppName(info.Title);
-                        options.OAuthClientId(config.OpenApiAppId);
-                        if (!string.IsNullOrEmpty(config.OpenApiAppSecret)) {
-                            options.OAuthClientSecret(config.OpenApiAppSecret);
+                        options.OAuthClientId(config.AppId);
```
marcschier: I would you leave the properties as they were called before.   We should have the ability to configure the swagger independent from the client auth..

common/src/Microsoft.Azure.IIoT.AspNetCore.OpenApi/src/Runtime/OpenApiConfig.cs
```diff
@@ -17,27 +17,29 @@ public class OpenApiConfig : ConfigBase, IOpenApiConfig {
...
+            GetBoolOrDefault("PCS_AUTH_REQUIRED", !string.IsNullOrEmpty(AppId)));
         /// <summary>Generate swagger.json</summary>
         public bool UseV2 => GetBoolOrDefault(kOpenApi_UseV2Key,
             GetBoolOrDefault("PCS_SWAGGER_V2", true));
+
         /// <summary>Application id</summary>
-        public string OpenApiAppId => GetStringOrDefault(kOpenApi_AppIdKey,
+        public string AppId => GetStringOrDefault(kAuth_AppIdKey,
             GetStringOrDefault(PcsVariable.PCS_AUTH_CLIENT_APPID,
-            GetStringOrDefault("PCS_OPENAPI_APP_ID")))?.Trim();
```
marcschier: Do not change this.   Just add the legacy environment variable PCS_WEBUI_AUTH_AAD_APPID as "default value" fallback to the original code.

common/src/Microsoft.Azure.IIoT.AspNetCore.OpenApi/src/Runtime/OpenApiConfig.cs
```diff
@@ -17,27 +17,29 @@ public class OpenApiConfig : ConfigBase, IOpenApiConfig {
...
+            GetStringOrDefault("PCS_WEBUI_AUTH_AAD_APPID")))?.Trim();
+
         /// <summary>App secret</summary>
-        public string OpenApiAppSecret => GetStringOrDefault(kOpenApi_AppSecretKey,
+        public string AppSecret => GetStringOrDefault(kAuth_AppSecretKey,
             GetStringOrDefault(PcsVariable.PCS_AUTH_CLIENT_SECRET,
-            GetStringOrDefault("PCS_OPENAPI_APP_KEY")))?.Trim();
+            GetStringOrDefault("PCS_APPLICATION_SECRET")))?.Trim();
```
marcschier: Same as above - do not change - just add the PCS_APPLICATION_SECRET varables as fallback to the PCS_OPENAPI_APP_KEY

PR #230, opened 2020-02-06
Updated security implementation

markusstuhler:


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -48,7 +49,16 @@ public class DefaultSessionManager : ISessionManager {
                     var sessionName = id.ToString();
                     var applicationConfiguration = _clientConfig.ToApplicationConfiguration(true);
                     var endpointConfiguration = _clientConfig.ToEndpointConfiguration();
-                    var endpointDescription = new EndpointDescription(id.Connection.Endpoint.Url);
+                    var endpointDescription = SelectEndpoint(id.Connection.Endpoint.Url, id.Connection.Endpoint.SecurityMode, id.Connection.Endpoint.SecurityPolicy, 15000);
```
cristipogacean: >15000 [](start = 165, length = 5)

would it be possible to replace this with a meaningful constant?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Exceptions/EndpointNotAvailableException.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Microsoft.Azure.IIoT.OpcUa.Core.Models;
+using Opc.Ua;
+using System;
```
marcschier: using convention

PR #231, opened 2020-02-06
Feature/cdm samples

cristipogacean:
* CDM logic changed to create entity and partitions for every Endpoint/nodeID
* added dynamic CDM entity generation
* removed JToken dependency on the MonitoredItemApiModel
* fixed the issue of signalR service stopping the publishing after 5 minutes
* added Value Status code for the published values
* The engineering tool is now capable to display live the value changes on the published nodes
* configurable event hub consumer groups for processors
* small bug-fixes and cosmetics


common/src/Microsoft.Azure.IIoT.Core/src/Hub/Runtime/IoTHubEventConfig.cs
```diff
@@ -45,10 +47,19 @@ public class IoTHubEventConfig : ConfigBase, IEventHubConsumerConfig {
...
         public string ConsumerGroup => GetStringOrDefault(kEventHubConsumerGroupKey,
             GetStringOrDefault("PCS_IOTHUB_EVENTHUBCONSUMERGROUP",
                 GetStringOrDefault("PCS_IOTHUBREACT_HUB_CONSUMERGROUP", "$default")));
+
+        /// <summary> Event hub telemetry consumer group </summary>
+        public string ConsumerGroupTelemetry => GetStringOrDefault(kEventHubConsumerGroupTelemetryKey,
+            GetStringOrDefault(PcsVariable.PCS_IOTHUB_EVENTHUB_CONSUMER_GROUP_TELEMETRY, "telemetry"));
+
+        /// <summary> Event hub events consumer group </summary>
+        public string ConsumerGroupEvents => GetStringOrDefault(kEventHubConsumerGroupEventsKey,
```
marcschier: Let's put this into the services Config class themselves, not into the interface.  Should be just consumer group, and then in the Processor service you overwrite it with the GetStringOrDefault(kEventHubConsumerGroupTelemetryKey,
            GetStringOrDefault(PcsVariable.PCS_IOTHUB_EVENTHUB_CONSUMER_GROUP_TELEMETRY, "telemetry")); 
etc.

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/MonitoredItemMessageApiModel.cs
```diff
@@ -45,15 +64,21 @@ public class MonitoredItemMessageApiModel {
         /// </summary>
         [JsonProperty(PropertyName = "value",
             NullValueHandling = NullValueHandling.Ignore)]
-        public JToken Value { get; set; }
-
+        public ValueApiModel Value {get; set; }
```
marcschier: I am unclear yet why the redundancy between TypeId and Value.Type.   And why the need for ValueApiModel.   Maybe do what we discussed and move completely to DataSetMessage as format, instead of this interim step?

common/src/Microsoft.Azure.IIoT.Core/src/Messaging/EventHub/Services/EventHubDeviceEventHandler.cs
```diff
@@ -36,17 +36,27 @@ public sealed class EventHubDeviceEventHandler : IEventHandler {
...
+                // From a device, but does not have any event schema or message schema
+                await _unknown.HandleAsync(eventData, properties);
             }
         }
 
         /// <inheritdoc/>
         public async Task OnBatchCompleteAsync() {
             foreach (var handler in _used.ToList()) {
-                await Try.Async(handler.OnBatchCompleteAsync);
+                if (handler != null) {
```
marcschier: We should look into why this can be null.   And then also fix that root cause.

common/src/Microsoft.Azure.IIoT.Messaging.SignalR/src/Services/SignalRServiceHost.cs
```diff
@@ -45,6 +45,8 @@ public class SignalRServiceHost : IIdentityTokenGenerator, IEndpoint,
             }).Build();
             Resource = !string.IsNullOrEmpty(config.SignalRHubName) ?
                 config.SignalRHubName : "default";
+            _renewHubTimer = new Timer(RenewHubTimer_ElapesedAsync);
```
marcschier: Please add a TODO: To remove this when signal r service SDK is fixed.

common/src/Microsoft.Azure.IIoT.Messaging.SignalR/src/Services/SignalRServiceHost.cs
```diff
@@ -185,6 +194,23 @@ public class SignalRServiceHost : IIdentityTokenGenerator, IEndpoint,
             return _hub.UserGroups.RemoveFromGroupAsync(client, group, ct);
         }
 
+        /// <inheritdoc/>
```
marcschier: Best to remove the comment or add a full comment for the private method

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Cdm/src/Storage/AdlsCsvStorage.cs
```diff
@@ -103,9 +112,9 @@ public class AdlsCsvStorage : IAdlsStorage {
                     request = _httpClient.NewRequest(
                         $"{partitionUrl}?action=append&position={contentPosition}",
                         kResource);
-                    request.SetContent(content);
+                    request.SetContent(content, UTF8Encoding.UTF8);
                     response = await _httpClient.PatchAsync(request);
-                    contentPosition += content.Length;
+                    contentPosition += UTF8Encoding.UTF8.GetByteCount(content);
```
marcschier: Just "Encoding" - no need to use the subclass here.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Subscriber/Models/MonitoredItemSampleModel.cs
```diff
@@ -4,9 +4,22 @@
 // ------------------------------------------------------------
 
 namespace Microsoft.Azure.IIoT.OpcUa.Subscriber.Models {
-    using Newtonsoft.Json.Linq;
     using System;
 
+    /// <summary>
+    /// the value
```
marcschier: Move to its own file.

services/src/Microsoft.Azure.IIoT.Services.Processor.Events/src/Runtime/Config.cs
```diff
@@ -31,7 +31,7 @@ public class Config : DiagnosticsConfig, IEventProcessorConfig, IEventHubConsume
         /// <inheritdoc/>
         public string EventHubPath => _eh.EventHubPath;
         /// <inheritdoc/>
-        public string ConsumerGroup => "events";
+        public string ConsumerGroup => _eh.ConsumerGroupEvents;
```
marcschier: This is where we should put reading the environment variable (see comment above) - also applies to Config.cs below.

common/src/Microsoft.Azure.IIoT.Core/src/Hub/Runtime/IoTHubEventConfig.cs
cristipogacean: Technically these are properties of the IoTHub Event, respectively the EventHub, that the platform requires being created/configured by the user upfront in the context of configuring the resources. By having all the properties of the resource in one place we make the configuration a bit easier.

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/MonitoredItemMessageApiModel.cs
cristipogacean: Ok, I'll get rid of the ValueApiModel for now and use the TypeId to transport the datatype of the value. then we will take the thorough step into the direction of DataSetMEssage format

common/src/Microsoft.Azure.IIoT.Messaging.SignalR/src/Services/SignalRServiceHost.cs
cristipogacean: done

common/src/Microsoft.Azure.IIoT.Messaging.SignalR/src/Services/SignalRServiceHost.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Subscriber/Models/MonitoredItemSampleModel.cs
cristipogacean: removed completely for now

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Cdm/src/Storage/AdlsCsvStorage.cs
cristipogacean: done

services/src/Microsoft.Azure.IIoT.Services.Processor.Events/src/Runtime/Config.cs
cristipogacean: as mentioned above, I believe this is part of the resource's configuration

common/src/Microsoft.Azure.IIoT.Core/src/Messaging/EventHub/Services/EventHubDeviceEventHandler.cs
cristipogacean: I'll add a bug to track this.

common/src/Microsoft.Azure.IIoT.Core/src/Hub/Runtime/IoTHubEventConfig.cs
marcschier: Imo they are specific to how the service requires the it not generically applicable to all.  We would add more and more for every service we add.

cristipogacean: done

services/src/Microsoft.Azure.IIoT.Services.Processor.Events/src/Runtime/Config.cs
cristipogacean: done

PR #259, opened 2020-02-18
Fix onboarding id and hash calculation due to missing properties

marcschier:
When calling the onboarding API the provided events do not contain all needed information to create an id that matches existing device entries.   This means existing entries get removed first, before same ones are re-added interrupting operation with existing device entities and causing unnecessary twin updates.

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Services/EndpointRegistry.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Services/EndpointRegistry.cs
@@ -388,45 +388,47 @@ namespace Microsoft.Azure.IIoT.OpcUa.Registry.Services {
             var unchanged = 0;
             var removed = 0;
 
-            // Remove or disable an endpoint
-            foreach (var item in remove) {
-                try {
-                    // Only touch applications the discoverer owns.
-                    if (item.DiscovererId == discovererId) {
-                        if (hardDelete) {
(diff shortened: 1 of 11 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Services/EndpointRegistry.cs
```diff
@@ -388,46 +388,48 @@ public sealed class EndpointRegistry : IEndpointRegistry, IApplicationEndpointRe
...
+                        if (item.DiscovererId == discovererId) {
+                            if (hardDelete) {
+                                var device = await _iothub.GetAsync(item.DeviceId);
+                                // First we update any registration
+                                var existingEndpoint = device.ToEndpointRegistration(false);
+                                if (!string.IsNullOrEmpty(existingEndpoint.SupervisorId)) {
+                                    await SetSupervisorTwinSecretAsync(existingEndpoint.SupervisorId,
+                                        device.Id, null);
+                                }
+                                // Then hard delete...
```
cristipogacean: Are we checking somehow if the endpoint is used in an active publishing job? In this case, we should not delete even when is inactive/disabled

marcschier: Yes, that is true.  But this would be a larger work item.  I will add this to our backlog.

PR #263, opened 2020-02-18
Fix following bugs from bugBash:

dacolgit:
- drawer action window should go away once one presses Apply button
- Write Node Cosmetic - truncated nodeId
- Three dots to get Asset info too small
- Home screen in Eng Tool is empty and should be removed
- Asset Info's 'Gateway Server Uri' and 'Capabilities' not filled
- "Eng Tool" should be called "Azure Industrial IoT Engineering Tool"


mregen:
> using System;

At some point there should be copyright headers...

---
Refers to: samples/src/Microsoft.Azure.IIoT.App/src/Common.cs:1 in 1c4f0aa. [](commit_id = 1c4f0aa232a10457b5b8a5646841020cb6305120, deletion_comment = False)

PR #269, opened 2020-02-19
Fixing access to Swagger in Kubernetes deployment.

karok2m:
* Changed Swagger UI generation logic.
  Paths shown by Swagger UI will now include request.PathBase prefix.
* Added configuration for URL path base.
* Added configuration and processing of forwarded headers.
* Added configuration for OpenApiServerHost.

### DIFF
--- a/common/src/Microsoft.Azure.IIoT.AspNetCore.OpenApi/src/Extensions/ApplicationBuilderEx.cs
+++ b/common/src/Microsoft.Azure.IIoT.AspNetCore.OpenApi/src/Extensions/ApplicationBuilderEx.cs
@@ -50,10 +50,10 @@ namespace Microsoft.OpenApi.Models {
 
                     // If request.PathBase exists, then we will prepend it to doc.Paths.
                     if (request.PathBase.HasValue) {
-                        var basePath = request.PathBase.Value;
+                        var pathBase = request.PathBase.Value;
                         var prefixedPaths = new OpenApiPaths();
                         foreach (var path in doc.Paths) {
-                            prefixedPaths.Add(basePath + path.Key, path.Value);
+                            prefixedPaths.Add(pathBase + path.Key, path.Value);
                         }
                         doc.Paths = prefixedPaths;
                     }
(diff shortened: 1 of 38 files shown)


common/src/Microsoft.Azure.IIoT.AspNetCore.OpenApi/src/Extensions/ApplicationBuilderEx.cs
```diff
@@ -32,8 +32,6 @@ public static class ApplicationBuilderEx {
             var addresses = app.ServerFeatures.Get<IServerAddressesFeature>()?.Addresses
                 .Select(a => new Uri(a.Replace("://*", "://localhost")))
                 .ToList() ?? new List<Uri>();
-            var path = addresses.FirstOrDefault()?.PathAndQuery.Trim('/') ??
```
marcschier: Removing this will break the services all container swagger I think.

karok2m: I've tested this for Services.All and it seems to work fine. Swaggers of all components that I've checked came up and showed correct endpoints.

The only problem that I saw there was that a few Swagger pages were empty (there were no endpoint definitions). This happened for the following components:

* configuration
* edge/manage

Is this expected ?

karok2m: @marcschier, the address of `openapi.json` has not changed. It is still served under `/swagger/v2/openapi.json` or `/<component_name>/swagger/v2/openapi.json`.

The difference is that `options.SwaggerEndpoint($"{info.Version}/openapi.json", info.Version)` now specifies it to be on `v2/openapi.json` endpoint **relative** to Swagger UI. Previously we were specifying an **absolute** endpoint, since it was starting with `/`. So do you still think that `swagger.cmd` has to be changed ?

marcschier: No - sounds great 👍

PR #275, opened 2020-02-21
fix the following bugs:

dacolgit:
- no report on re-logon required
- Id Shown in endpoint control
- OPC UA Variable Int32 value is shown as N/A
- display the bottom of "Scan Results" so the user can notice progress


common/src/Microsoft.Azure.IIoT.AspNetCore/src/Auth/Clients/BehalfOfTokenProvider.cs
```diff
@@ -83,13 +83,25 @@ public partial class BehalfOfTokenProvider : ITokenProvider {
...
-                    new ClientCredential(_config.AppId, _config.AppSecret),
-                    new UserAssertion(token, kGrantType, name));
+
+            try {   
+                var result = await ctx.AcquireTokenSilentAsync(resource, _config.AppId);
                 return result.ToTokenResult();
             }
             catch (AdalException ex) {
+                if (ex.ErrorCode == AdalError.FailedToAcquireTokenSilently) {
+                    try {
```
cristipogacean: >       try {         [](start = 13, length = 20)

I would avoid imbricating try ... catch blocks. Just return null in case of failure and take the imbricated try...catch outside the catch

samples/src/Microsoft.Azure.IIoT.App/src/Shared/MainLayout.razor
```diff
@@ -11,6 +11,14 @@
...
 
     <div class="content px-4">
-        @Body
+        <AuthorizeView>
+            <Authorized>
+                @Body
+            </Authorized>
+            <NotAuthorized>
+                <h1>Authentication Failure or expired</h1>
+                <p>Please signed in.</p>
```
marcschier: sign instead of "signed"

samples/src/Microsoft.Azure.IIoT.App/src/Startup.cs
```diff
@@ -150,7 +151,7 @@ public class Startup {
             });
 
             services.AddAntiforgery(options => {
-                options.Cookie.SameSite = SameSiteMode.None;
+                options.Cookie.SameSite = SameSiteMode.Strict;
```
marcschier: Please build and deploy from your branch - run "acrbld <name of deployment>" then try using the app from there.   This is a source of failures in the engineering tool when running in container (i.e. deploy.ps1 script).

common/src/Microsoft.Azure.IIoT.AspNetCore/src/Auth/Clients/BehalfOfTokenProvider.cs
```diff
@@ -83,13 +83,25 @@ public partial class BehalfOfTokenProvider : ITokenProvider {
...
+                        new UserAssertion(token, kGrantType, name));
+                        return result.ToTokenResult();
+                    }
+                    catch (AdalException ex1) {
+                        _handler.Handle(_ctx.HttpContext, new AuthenticationException(
+                        $"Failed to authenticate on behalf of {name}", ex1));
+                        return null;
+                    }
+                }
                 _handler.Handle(_ctx.HttpContext, new AuthenticationException(
```
marcschier: This handler is what we should use.   It was intended to do the non-silent auth.   However, what I would suggest is to change it to remove the _handler entirely and have instead a bool in the constructor to enable this new code.  Set the bool to false by default (bool acquireTokenIfSilentFails = false) to be backcompat, and set it explicitly to true when adding the provider in the Autofac container in Startup.cs.

samples/src/Microsoft.Azure.IIoT.App/src/Shared/MainLayout.razor
dacolgit: changed

samples/src/Microsoft.Azure.IIoT.App/src/Startup.cs
dacolgit: revert to none. I will do more test

common/src/Microsoft.Azure.IIoT.AspNetCore/src/Auth/Clients/BehalfOfTokenProvider.cs
dacolgit: changed

PR #294, opened 2020-03-04
Helm Chart for 2.6.104

karok2m:


deploy/helm/azure-industrial-iot/Chart.yaml
```diff
@@ -14,4 +14,4 @@ home: https://github.com/Azure/Industrial-IoT
 maintainers:
 - name: Karapet Kostandyan
   email: kakostan@microsoft.com
-appVersion: 2.5.2
+appVersion: 2.6.62
```
cristipogacean: shouldn't it be 2.6.10x or so?

deploy/helm/azure-industrial-iot/README.md
```diff
@@ -490,6 +837,191 @@ azure:
...
 ```
+
+## Special Notes
+
+### Resource Requests And Limits
+
+Helm chart does not set any [resource requests or limits](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/)
+for Pod Containers. We recommend to set memory requests to at least `"64Mi"` and cpu requests to at least
+`"20m"`. If you set resource limits, be informed that some components might require up to `"256Mi"` of
+memory, such as `frontend`, `registry`, `publisher` and a few others.
```
cristipogacean: >frontend [](start = 17, length = 8)

I think we use the term "engineering tool" for "frontend" in other places in documentation, I have no clue if this is right or not, but we should be consequent through documentation with naming of our components.

deploy/helm/azure-industrial-iot/values.yaml
```diff
@@ -5,12 +5,12 @@ image:
   # image.registry is URL of Docker registry from where images will be pulled.
   registry: mcr.microsoft.com
   # image.tag defines which version of Docker images to pull.
-  tag: 2.5.2
+  tag: 2.6.104
```
cristipogacean: >104 [](start = 11, length = 3)

can't we use :latest respective :preview here?

deploy/helm/azure-industrial-iot/Chart.yaml
karok2m: Fixed. It should be `2.6.104`.

deploy/helm/azure-industrial-iot/values.yaml
karok2m: We shouldn't. Helm best practices explicitly state not to use 'floating' image tags and to always prefer fixed ones.

We should keep in mind that in Kubernetes environment different components may fail and be restarted independently. Each restart can result in a new image being pulled and used. So having a floating tag may then result in a situation where you have components of several different versions in the same environment. Also, having a floating tag would mean not having reproducible deployments, which should also be avoided.

karok2m: BTW, this does not mean that users cannot change the tag. Quite the opposite, it is very easy to override this value. But then users would have to explicitly do that at their own risk.

deploy/helm/azure-industrial-iot/README.md
karok2m: Good point, thanks. I will change it to engineering tool and `engTool` for configuration parameters. But I will keep `/frontend` as URL path base for engineering tool. Sound good ?

karok2m: Decided to use `engineeringTool` for configuration as well.

PR #298, opened 2020-03-08
Feature/processor networkmessage

cristipogacean:
* added PubSub json and uadp handling in the telemetry processors
* added Display Name processing in the publisher


common/src/Microsoft.Azure.IIoT.Core/src/Messaging/EventHub/Services/EventHubDeviceEventHandler.cs
```diff
@@ -52,7 +56,7 @@ public sealed class EventHubDeviceEventHandler : IEventHandler {
         /// <inheritdoc/>
         public async Task OnBatchCompleteAsync() {
             foreach (var handler in _used) {
-                await Try.Async(_handlers[handler].OnBatchCompleteAsync);
+                await Try.Async(_handlers[handler]!.OnBatchCompleteAsync);
```
marcschier: Is this intentended?

common/src/Microsoft.Azure.IIoT.Core/src/Utils/JsonConvertEx.cs
```diff
@@ -57,7 +57,11 @@ public static class JsonConvertEx {
         /// <returns></returns>
         internal static JsonSerializerSettings GetDefaultSettings() {
             return new JsonSerializerSettings {
-                ContractResolver = new DefaultContractResolver(),
+                ContractResolver = new DefaultContractResolver() {
+                    NamingStrategy = new CamelCaseNamingStrategy {
+                        ProcessDictionaryKeys = false
```
marcschier: QQ: where is this coming into play?

common/src/Microsoft.Azure.IIoT.Core/src/Messaging/EventHub/Services/EventHubDeviceEventHandler.cs
cristipogacean: It was an attempt to get rid of the NullReferenceException we had, I will revert it.

common/src/Microsoft.Azure.IIoT.Core/src/Utils/JsonConvertEx.cs
cristipogacean: this is required for the serialization of the payload dictionary if form "mykey": {"value":"myvalue"}

marcschier: Got it, thanks!

PR #304, opened 2020-03-10
consistent behavior of telemetry processors

cristipogacean:


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesJobConverter.cs
```diff
@@ -126,38 +126,46 @@ public class PublishedNodesJobConverter {
                             DiagnosticsInterval = _config.DiagnosticsInterval
                         },
                         WriterGroup = new WriterGroupModel {
-                            WriterGroupId = null,
+                            WriterGroupId = "LegacyPublisher",
                             DataSetWriters = new List<DataSetWriterModel> {
                                 new DataSetWriterModel {
-                                    DataSetWriterId = Guid.NewGuid().ToString(),
+                                    DataSetWriterId = "LegacyPublisher",//Guid.NewGuid().ToString(),
```
marcschier: Might be good to use the publisher identity here

cristipogacean: You're right, I'll open a work item on this topic, because the change is slightly bigger and we should probably also change the PublisherID currently hardcoded to "LegacyJob".

PR #317, opened 2020-03-16
Add node list filter and re-generate swagger open api

marcschier:
Also fixes uint being exposed as int32 - now exposed as int64.


karok2m:
@marcschier, shouldn't we also generate Swagger JSON files for new services with APIs as well ?

Issue #306, opened 2020-03-11
OPC Twin with large number of nodes returns only 100 nodes - also returns error in error info diagnostics


caschw:
I have 10,000 generated references under one node. When I call browse on the OPC twin for that node, no continuation token is returned (null) and I get a BadNoConinuationPoints errorInfo object returned in the response

```
"errorInfo": {
    "diagnostics": [
      "BadNoContinuationPoints"
    ]
  }
```
What I expected was the first 100 nodes returned and a continuationToken that could be used with BrowseNext.

marcschier:
Hi @caschw, BadNoContinuationPoints is returned by the OPC UA server.  This error indicates that the server does not have enough simultaneously active continuation points and is running at limit.  If not otherwise specified in the request we set MaxReferencesToReturn to 0u, meaning the server should decide the batch sizes.   Maybe try to set it to MaxReferencesToReturn: 10000?.

pcameron-ptc:
@caschw , @marcschier is correct, it's up to the UA server to allocate some max number of continuation points. If, during the browse, the UA server has already used all of its allocated continuation points, then the browse will fail as you've indicated. 

You can observe the limit in any OPC UA server's Server.ServerCapabilities object:
![image](https://user-images.githubusercontent.com/43643709/76422491-acd52b00-637b-11ea-871f-99e72a9cce66.png)

marcschier:
Additional info - there seems to be only 100 items returned when MaxReferenceToReturn > 100 with above error.   Worth looking into the client or UA stack to see if this is an issue on the Twin side.

pcameron-ptc:
Update: Kepware does not read the MaxBrowseContinuationPoints from the object instance, but from the Type node. The max point count is hard-coded to 5 in the product. 

I'm continuing to investigate whether or not there is an issue in the Kepware Continuation Point browsing implementation:

![image](https://user-images.githubusercontent.com/43643709/76523191-5d0b6800-643e-11ea-8cde-c297b2957e6d.png)

pcameron-ptc:
@marcschier , I have some additional information regarding this issue. 

Carl's original description is correct - when we browse a node with 10k children using the Java swagger client (no value assigned to "withMaxReferencesToReturn"), we receive a continuation point and all 10k results (actually 10002 with a couple 'meta' data nodes thrown in). The continuation point in the response is null, and the error info is populated with no status code or error message, but with a diagnostic message that says "BadNoContinuationPoints". 

[...]

Issue #291, opened 2020-03-02
OPC publisher module rPI 4 - "exec format error"


RotateAt60MPH:
I am trying to get the OPC Publisher working on a rPI4 with Raspian Buster.  After the module is deployed, it does not start and has the following error:

> pi@raspberrypi:~ $ sudo iotedge logs publisher -f
> `standard_init_linux.go:211: exec user process caused "exec format error"`

I have verified that the main iotedge modules are running:

> pi@raspberrypi:~ $ sudo iotedge list
> NAME             STATUS           DESCRIPTION      CONFIG
> edgeHub          running          Up 33 minutes              mcr.microsoft.com/azureiotedge-hub:1.0
> publisher        failed           Failed (1) 15 seconds ago  mcr.microsoft.com/iotedge/opc-publisher:latest
> edgeAgent        running          Up 42 minutes              mcr.microsoft.com/azureiotedge-agent:1.0

'sudo iotedge check' runs with no errors.

Docker was installed using the get-docker script.  

> pi@raspberrypi:~ $ curl -fsSL https://get.docker.com -o get-docker.sh
> pi@raspberrypi:~ $ sudo sh get-docker.sh

I've verified that docker is running since I am able to see the containers:

> pi@raspberrypi:~ $ docker container list
> CONTAINER ID        IMAGE                                      COMMAND                   CREATED             STATUS              PORTS                                                                  NAMES
> 2e49ec44dfb3        mcr.microsoft.com/azureiotedge-hub:1.0     "/bin/sh -c 'echo \"$â¦"   20 minutes ago      Up 20 minutes       0.0.0.0:443->443/tcp, 0.0.0.0:5671->5671/tcp, 0.0.0.0:8883->8883/tcp   edgeHub
> 3831e38148ec        mcr.microsoft.com/azureiotedge-agent:1.0   "/bin/sh -c 'echo \"$â¦"   29 minutes ago      Up 29 minutes   

If it was the wrong build of docker (X64 versus armhf),I wouldn't see the edge containers would I?

Any ideas appreciated.

RotateAt60MPH:
Further information.  I stopped the iotedge service, and deleted the publisher container in docker.  I then went back to OPC Publisher 2.5.3 version and the problem went away.

Certainly looks like somebody got broke something in the :latest version.

cristipogacean:
Thanks for reporting this, we will look into it asap!

cristipogacean:
[...]

PR #321, opened 2020-03-17
Fix browse #306 and add arm64 #291

marcschier:
* Fix browse returns at max 100 items even if more is specified and add unit tests (#306 )
* Add Linux-arm64 target to multi arch container build (#291 )
* Fix line break ugliness in deploy script prompts

### DIFF
--- a/deploy/src/Microsoft.Azure.IIoT.Deployment/Infrastructure/SignalRMgmtClient.cs
+++ b/deploy/src/Microsoft.Azure.IIoT.Deployment/Infrastructure/SignalRMgmtClient.cs
@@ -25,10 +25,30 @@ namespace Microsoft.Azure.IIoT.Deployment.Infrastructure {
 
         private readonly SignalRManagementClient _signalRManagementClient;
 
+        /// <summary>
+        /// Defines possible ServiceModes for SignalR service.
+        /// </summary>
+        public class ServiceMode {
+            private ServiceMode(string value) { Value = value; }
+
+            public string Value { get; set; }
+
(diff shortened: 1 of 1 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Twin/src/Control/Services/AddressSpaceServices.cs
```diff
@@ -868,13 +868,18 @@ public sealed class AddressSpaceServices : INodeServices<EndpointModel>,
...
                                 diagnostics, response?.Results?.Select(r => r.StatusCode),
                                 response?.DiagnosticInfos, true);
-                            if (response.Results.Count == 1) {
+                            if (response.Results.Count > 0) {
                                 children = response.Results[0].References.Count != 0;
+                                if (response.Results[0].ContinuationPoint != null) {
+                                    await session.BrowseNextAsync(header, true,
+                                        new ByteStringCollection {
+                                            response.Results[0].ContinuationPoint
+                                        });
```
cristipogacean: Why do we force here a browse next? just to release the provided continuation point? in that case maybe we should do that also when no results are returned (count == 0) but a valid continuation point - would be super weird, but you never know ...

marcschier: Thought the same and had that first - but found that really super weird.   But yes, this is interesting to trace.

Issue #319, opened 2020-03-16
Cannot add Kepware server application when it is remote from the IoT Edge Runtime


pc-avatar-7076:
**Describe the bug**
When Kepware (TKS) is installed on a Windows 10 VM (same network / resource group as IoT Edge RT VM) then "add application" fails. 

**To Reproduce**
1. Deploy the IIoT 2.6 stack, including an edge with 2.6 modules
2. Install Kepware (TKS) on a Windows 10 VM in the same resource group
3. Update the Kepware configuration to expose a discovery URL to the network
4. Use Postman (or the Swagger client) to invoke AddApplication. 
5. The Kepware TKS server never appears in the list of applications. 

**Expected behavior**
The remote TKS server can be added

**Screenshots**

This illustrates a connection config using UA Expert:

![ua-expert-fqdn](https://user-images.githubusercontent.com/43643709/76798281-ac86c680-67a5-11ea-9dcf-0227f7405b70.PNG)

This shows UA Expert connected using this configuration. UA Expert is running on my host in this example (the TKS server is running on an Azure hosted Windows 10 VM):

![ua-expert-connected](https://user-images.githubusercontent.com/43643709/76798204-86f9bd00-67a5-11ea-923a-0257db7bfc74.PNG)

Here is the Kepware discovery URL configuration. Note the hostname is not fully qualified:

![kepware-tks-configuration](https://user-images.githubusercontent.com/43643709/76798325-c45e4a80-67a5-11ea-9f78-7735e855b861.PNG)

Here is Postman initiating the AddApplications call using "no security":

![postman-add-application](https://user-images.githubusercontent.com/43643709/76798407-e952bd80-67a5-11ea-8a32-53f2e807da29.PNG)

Finally, here is the discovery module output:

![discovery-logs-zero-discovery-urls](https://user-images.githubusercontent.com/43643709/76798440-f374bc00-67a5-11ea-8535-15a36c71a366.PNG)

marcschier:
Hi @pcameron-ptc, would it be possible for you to send us the ARM template that you used to deploy the VM and IoT Edge VM or a description of the network configuration for both VM?   If the IoT Edge VM and cannot resolve the name or the resolved IP is not reachable then this might be happening.   We will see if we can replicate such a setup in our simulation environment.

cristipogacean:
[...]

PR #323, opened 2020-03-17
fix discovery to only use IPv4 dns resolved addresses

cristipogacean:
fix for #319


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Discovery/src/Services/DiscoveryServices.cs
```diff
@@ -437,8 +438,11 @@ public class DiscoveryServices : IDiscoveryServices, IScannerServices, IDisposab
         private Task<(IPEndPoint, Uri)> GetHostEntryAsync(
             Uri discoveryUrl) {
             return Try.Async(async () => {
+
                 var entry = await Dns.GetHostEntryAsync(discoveryUrl.DnsSafeHost);
-                foreach (var address in entry.AddressList) {
+                // only pick-up the IPV4 addresses
+                foreach (var address in entry.AddressList
```
mregen: Hmm.. algorithm should be to send ping to every address then wait for the first to return and cancel outstanding requests? Any way hosts don't have to answer ping. But I don't have all the context here...

marcschier: This is to pick an endpoint that can be connected to using the discovery service client.  Ipv4 must be preferred, however - we should make sure that there is at least an item in the list, e.g. if there are no ipv4 addresses we should still use ipv6.

PR #341, opened 2020-03-21
Optimize variable resolution in configuration objects

marcschier:
This is a first cut at optimizing configuration.  Currently every fallback is resolved before reading the actual configuration, that means that heavy computation is executed before a set variable is read.  This provides the minimal changes to only execute the fallback as lambda when the actual value cannot be found.   This looks like larger changes, but in essence it is just making the fallback callbacks.

Also removes dead code.


common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/ModuleHost.cs
```diff
@@ -362,7 +362,7 @@ public sealed class ModuleHost : IModuleHost, ITwinProperties, IEventEmitter,
                     continue;
                 }
                 if (!ProcessEdgeHostSettings(property.Key, value)) {
-                    desired[property.Key] = value;
+                    // desired[property.Key] = value;
```
mregen: comment?

PR #362, opened 2020-03-29
Feature/heartbeat

cristipogacean:


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Extensions/PublishedDataItemsModelEx.cs
```diff
@@ -31,9 +31,9 @@ public static class PublishedDataItemsModelEx {
                     continue;
                 }
                 var monitoredItem = item.ToMonitoredItem();
-                map.Add(monitoredItem.Id ?? Guid.NewGuid().ToString(),
-                    monitoredItem);
-                if (item.HeartbeatInterval == null) {
+                map.Add(monitoredItem.Id ?? Guid.NewGuid().ToString(), monitoredItem);
+                /*  Heartbeat using triggering mode
```
marcschier, line 35: If we never believe this will be using triggering lets just remove the commented out code.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Extensions/MonitoredItemNotificationModelEx.cs
```diff
@@ -35,10 +35,6 @@ public static class MonitoredItemNotificationModelEx {
                 if (message == null) {
                     continue;
                 }
-                if (message.DiagnosticInfo == null && notification.DiagnosticInfos != null &&
```
marcschier, line 38: Why did you remove this, was it causing issues or is it never null?   We want to flow the diagnostics info back to the configuration eventually so having this does not hurt on the way there.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Extensions/PublishedDataItemsModelEx.cs
cristipogacean: let's keep it as sample for generic triggering feature, I think it makes a lot of sense to promote this feature.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Extensions/MonitoredItemNotificationModelEx.cs
cristipogacean: The diagnostics (if any) are already added to the message through: public static MonitoredItemNotificationModel ToMonitoredItemNotification(). Diagnostics are copied from the DataChangeNotification.DiagnosticInfo array to the MonitoredItemNotification.DiagnosticInfo by the stack before the callback is triggered.

PR #376, opened 2020-04-08
add TSI link

dacolgit:


common/src/Microsoft.Azure.IIoT.Core/src/TSI/Runtime/TsiConfig.cs
```diff
@@ -0,0 +1,28 @@
...
+    using Microsoft.Extensions.Configuration;
+    using System;
+
+    /// <summary>
+    /// TSI configuration
+    /// </summary>
+    public class TsiConfig : ConfigBase, ITsiConfig {
+
+        /// <summary>DataAccessFQDN</summary>
+        public string DataAccessFQDN => GetStringOrDefault(PcsVariable.PCS_TSI_URL,
```
karok2m, line 17: We usually have a non-environment variable configuration as first configuration source. For example, for `ADLSg2HostName` parameter in `CdmClientConfig.cs` we first try to get value of `Cdm:ADLSg2HostName` before falling back to `PcsVariable.PCS_ADLSG2_ACCOUNT` or others. Can you please add similar configuration here as well ?

dacolgit: added

common/src/Microsoft.Azure.IIoT.Core/src/TSI/Runtime/TsiConfig.cs
```diff
@@ -0,0 +1,30 @@
...
+
+    /// <summary>
+    /// TSI configuration
+    /// </summary>
+    public class TsiConfig : ConfigBase, ITsiConfig {
+
+        private const string kTsi_DataAccessFQDN = "Tsi:DataAccessFQDN";
+
+        /// <summary>DataAccessFQDN</summary>
+        public string DataAccessFQDN => GetStringOrDefault(kTsi_DataAccessFQDN,
```
marcschier, line 19: It is nice work!  But my preference would be to just add this property to the App config directly - the platform core should not have anything TSI in it.   If you need to inject an interface, feel free to move both this and above file to the engineering tool codebase.   Leave the PcsVariable entry though.

PR #378, opened 2020-04-08
Helm: Changed deployments to use envFrom for env var injection.

karok2m:
Previously we had a script that was generating `env` entries from `env.txt` file. Now I am using `envFrom` to inject all environment variables from a Secret to a Deployment.

This way we:
* do not need to generate one entry per environment variable
* do not need to know about structure of the Secret beforehand

Second feature in particular should enable us to generate and consume much smaller configuration Secret in cases when we want to read environment variables from KeyVault.


karok2m:
Updated components to version `2.6.146`.

cristipogacean:
> {{- if .Values.deployment.microServices.telemetryUxProcessor.enabled }}

I think telemetry_ux was removed in the last commit. perhaps we should remove it here as well? 

---
Refers to: deploy/helm/azure-industrial-iot/templates/40_telemetry_ux_processor_deployment.yaml:1 in f94cb5f. [](commit_id = f94cb5f9366ea425ca103422b8ac3c9f21663e60, deletion_comment = False)

karok2m:
@cristipogacean, those changes refer to `2.6.146` version of components, where `telemetryUxProcessor` still exists. I believe, Marc is going to change the chart for `2.7.x` version of components with reduced number of services.

PR #397, opened 2020-04-27
various fixes for publisher

cristipogacean:
* standalone supports up to 5 sessions
 * sequence number added to the message
 * fixes for CDM handling


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Stack/PubSub/MonitoredItemMessage.cs
```diff
@@ -194,6 +202,9 @@ public class MonitoredItemMessage : IEncodeable {
                 Timestamp = decoder.ReadDateTime(nameof(MonitoredItemMessageContentMask.Timestamp));
             }
             Value = decoder.ReadDataValue("Value");
+            if ((MessageContentMask & (uint)MonitoredItemMessageContentMask.SequenceNumber) == 0) {
+                SequenceNumber = decoder.ReadUInt32(nameof(MonitoredItemMessageContentMask.SequenceNumber));
```
mregen, line 206: shouldn't this be != 0?

PR #405, opened 2020-04-28
IAI: Pushing IIoTEnvironment to KeyVault.

karok2m:


deploy/src/Microsoft.Azure.IIoT.Deployment/Deployment/DeploymentExecutor.cs
```diff
@@ -1202,6 +1206,9 @@ await _computeManagementClient
             // This will be performed in UpdateClientApplicationRedirectUrisAsync() call.
             _applicationURL = aksPublicIp.DnsSettings.Fqdn;
 
+            // Waiting for unfinished tasks.
+            keyVaultConfCreationTask.Wait();
+
```
mregen, line 1211: why not await?

karok2m: Habit, I guess. You are correct, I will change all `.Wait()`s and `.Result`s to use `await`.

mregen: at least don't mix it...it can cause deadlocks

PR #412, opened 2020-05-03
legacy publisher multiple subscriptions fix

cristipogacean:


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -63,13 +70,16 @@ public class LegacyJobOrchestrator : IJobOrchestrator {
         /// <param name="ct"></param>
         /// <returns></returns>
         public Task<JobProcessingInstructionModel> GetAvailableJobAsync(string workerId, JobRequestModel request, CancellationToken ct = default) {
-            _updated = false;
             if (_assignedJobs.TryGetValue(workerId, out var job)) {
                 return Task.FromResult(job);
             }
-            if ((job = _availableJobs.Dequeue()) != null) {
+            if (_availableJobs.Count() > 0 && (job = _availableJobs.Dequeue()) != null) {
```
marcschier, line 76: Use Count property instead of linq as it is faster (reads the count value, vs traversing).   More so though, here you could use TryDequeue which might even be faster as it checks and dequeues in one call.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -85,7 +95,9 @@ public class LegacyJobOrchestrator : IJobOrchestrator {
             HeartbeatResultModel heartbeatResultModel;
 
             if (_updated && heartbeat.Job != null) {
-                _updated = false;
+                if (_availableJobs.Count() == 0) {
```
marcschier, line 98: Use Count, instead of Count()

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
cristipogacean: done

PR #413, opened 2020-05-04
Changing order of configuration providers.

karok2m:
Purpose of this PR is to establish order of configuration providers so that any local configuration provider can override a value from KeyVault. This is required so that deployment through a Helm chart can use configuration deployed through `deploy.ps1` or IAI to KeyVault, but would still be able to override any value if needed. This will also help in local debugging when one wants to re-use most but not all values from KeyVault.

In general this is the order of increasing priority that I propose for configuration provider groups:
1. remote configuration provider (KeyVault)
2. local bulk configuration providers, `appsettings.json` and `.env`
3. local singular configuration providers, environment variables and command line arguments

More specifically, the order would look like this for configuration providers that we use:
1. KeyVault configuration provider
2. `appsettings.json` (if applicable)
3. `.env` file
4. Environment variables
5. User environment variables
6. Command line arguments (if applicable)

It should be noted that there is technical peculiarity with initialization of KeyVault configuration provider. As it itself uses the provided `IConfigurationBuilder` object to extract KeyVault connection details (including credentials). So to make things work I had to first determine the order of configuration providers for each service and then use all except KeyVault configuration provider twice. First  in the intended order before initialization of KeyVault configuration provider (to make sure that intended override order will be applied for configuration of for KeyVault connections details), and then again after initialization of KeyVault configuration provider so that values from local providers override values from KeyVault. The overhead of duplicating the providers would slow down loading of configuration, but that should happen only once when components starts.

### DIFF
--- a/common/src/Microsoft.Azure.IIoT.Auth.ActiveDirectory/src/Extensions/ConfigurationEx.cs
+++ b/common/src/Microsoft.Azure.IIoT.Auth.ActiveDirectory/src/Extensions/ConfigurationEx.cs
@@ -31,9 +31,10 @@ namespace Microsoft.Extensions.Configuration {
         /// <param name="keyVaultUrlVarName"></param>
         /// <returns></returns>
         public static IConfigurationBuilder AddFromKeyVault(
-            this IConfigurationBuilder builder, bool singleton = true,
-            string keyVaultUrlVarName = null) {
-
+            this IConfigurationBuilder builder,
+            bool singleton = true,
+            string keyVaultUrlVarName = null
+        ) {
             var provider = KeyVaultConfigurationProvider.CreateInstanceAsync(
                 singleton, builder.Build(), keyVaultUrlVarName).Result;
             if (provider != null) {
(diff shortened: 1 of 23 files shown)


common/src/Microsoft.Azure.IIoT.Auth.ActiveDirectory/src/Extensions/ConfigurationEx.cs
```diff
@@ -31,9 +31,10 @@ public static class ConfigurationEx {
         /// <param name="keyVaultUrlVarName"></param>
         /// <returns></returns>
         public static IConfigurationBuilder AddFromKeyVault(
-            this IConfigurationBuilder builder, bool singleton = true,
-            string keyVaultUrlVarName = null) {
-
+            this IConfigurationBuilder builder,
```
marcschier, line 34: Only change seems to be formatting, or?

karok2m: Yes, all changes to `ConfigurationEx.cs` are just cosmetic. I can revert those if you want.

karok2m:
@marcschier, well, actually there is away to add a configuration provider to the beginning of `IConfigurationBuilder.Sources` list, thus giving it the lowest priority. In this case the code would look something like this:

```
            new ConfigurationBuilder()
                .AddConfiguration(configuration)
                .AddFromDotEnvFile()
                .AddEnvironmentVariables()
                .AddEnvironmentVariables(EnvironmentVariableTarget.User)
                // Above configuration providers will provide connection
                // details for KeyVault configuration provider.
                .AddFromKeyVaultWithLowestPriority()
                .Build()
```

Here KeyVault configuration provider will use providers configured before it to get KeyVault configuration details then insert itself to the beginning of the providers list.

In my opinion, it is a bit confusing, but it can definitely work. Any naming suggestions for `AddFromKeyVaultWithLowestPriority` ?

marcschier:
Awesome, maybe AddDefaultsFromKeyvault() or so.

karok2m:
@marcschier, I've added `providerPriority` parameter to `AddFromKeyVault()` function which determines where in configuration providers chain KeyVault configuration provider is added.

PR #415, opened 2020-05-04
fix for edge metrics

monikavar:


common/src/Microsoft.Azure.IIoT.Hub.Module.Client/src/Extensions/MetricServerEx.cs
```diff
@@ -17,8 +18,17 @@ public static class MetricServerEx {
...
-                logger.Information("Prometheus metric server started.");
+                try {
+                    server.Start();
+                    logger.Information("Prometheus metric server started.");
+                }
+                catch (HttpListenerException e){
+                    logger.Error(e, "Unable to start metric server. To enable edge metrics collection, please run the following commands in an elevated window:\n" +
+                        "\nnetsh http add urlacl url=http://+:9700/metrics user=Everyone" +
+                        "\nnetsh http add urlacl url=http://+:9701/metrics user=Everyone" +
+                        "\nnetsh http add urlacl url=http://+:9702/metrics user=Everyone\n");
```
cristipogacean, line 29: I think we should not add the commands in the log. they are only relevant for windows. and it is possible that you get a similar exception in linux as well and then the hint would be missleading. I think we should just document this topic in a troubleshooting section in our documentation.

marcschier: +1

monikavar: done

PR #419, opened 2020-05-05
Add codecov and codacy code coverage upload option

mregen:
Codecov: https://codecov.io/gh/Azure/Industrial-IoT
Codacy requires an upload token, at this point the free plan works only on private forks.

### DIFF
--- a/tools/templates/cc.yml
+++ b/tools/templates/cc.yml
@@ -33,6 +33,11 @@ jobs:
       projects: '**/tests/*.csproj'
       arguments: '--configuration Release /p:CollectCoverage=true /p:CoverletOutputFormat=cobertura'
       publishTestResults: false
+  - script: |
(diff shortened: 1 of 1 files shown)


tools/templates/cc.yml
```diff
@@ -33,6 +33,13 @@ jobs:
       projects: '**/tests/*.csproj'
       arguments: '--configuration Release /p:CollectCoverage=true /p:CoverletOutputFormat=cobertura'
       publishTestResults: false
+  - script: |
+      bash <(curl -s https://codecov.io/bash)
```
marcschier, line 37: I am a bit concerned about this - these are IMO top security issues since they allow arbitrary code execution on our build servers and potentially secret exfiltration.   Need a better way.

mregen: right, better not use it here

PR #420, opened 2020-05-05
MSI authentication support / Fix keyvault access issue.

marcschier:


deploy/templates/azuredeploy.standard.json
```diff
@@ -385,88 +411,29 @@
...
-                ]
+                "roleDefinitionId": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', 'ba92f5b4-2d11-453d-a403-e96b0029c9fe')]",
+                "principalId": "[parameters('datalakeServicePrincipalId')]",
+                "principalType": "ServicePrincipal"
             },
             "dependsOn": [
-                "[variables('eventHubResourceId')]"
+                "[variables('datalakeResourceId')]"
             ]
         },
```
cristipogacean, line 438: do we still need this if we changed the  adls storage to use connection string?

marcschier: No, technically not - however I want to keep it around so we can permission a user principal to it, just like we are now doing with the keyvault resource.

PR #427, opened 2020-05-07
Add final newline to .json and .sh files

luiscantero:
Also, use LF as newline char.


luiscantero:
@marcschier Visual Studio will apply the rule to affected files during the next file save. VS Code requires [an extension + npm package](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) for the same effect. Specially .sh files could misbehave if the trailing LF is missing. I will hunt down the files and will submit another PR.

PR #432, opened 2020-05-08
Mabakovi/disabled endpoints

MaidaB:
When endpoint and assets are not active they will be grayed out.


samples/src/Microsoft.Azure.IIoT.App/src/wwwroot/css/site.css
```diff
@@ -270,6 +270,18 @@ a.adhocmargin {
     display: flex;
 }
 
+.activated-true {
+    font-weight: bold;
+}
+
+.seen-true {
```
marcschier, line 277: I would call it .enabled instead of .seen - it makes it clearer

MaidaB: Renamed

PR #434, opened 2020-05-08
Tolerating '/' in the end to service URLs.

karok2m:


api/src/Microsoft.Azure.IIoT.Api/src/Publisher/Clients/PublisherJobServiceClient.cs
```diff
@@ -36,9 +36,12 @@ public class PublisherJobServiceClient : IPublisherJobServiceApi {
         /// <param name="serializer"></param>
         public PublisherJobServiceClient(IHttpClient httpClient, string serviceUri,
             ISerializer serializer = null) {
-            _serializer = serializer ?? new NewtonSoftJsonSerializer();
-            _serviceUri = serviceUri ?? throw new ArgumentNullException(nameof(serviceUri),
```
marcschier, line 40: Another nifty way would be serviceUri?.TrimEnd('/') ?? throw new ArgumentNullException(nameof(serviceUri));

karok2m: Yes, I initially wanted to do that everywhere, but unfortunately it does not catch empty strings.

PR #446, opened 2020-05-12
Set development environment by default

MaidaB:


karok2m:
@MaidaB, how will this be overridden when we deploy the app ?

Issue #430, opened 2020-05-07
OPC Publisher could not issue a valid certificate without deleting PKI folder


SewerynAW:
**Describe the bug**
The OPC publisher could not connect to the OPC UA server because the client certificate has expired and did not renew automatically. After manually removing the PKI folder and restarting Publisher, it forced to create a new folder and issue a valid client certificate for OPC server.

Log with error.
[opc-Publisher.log](https://github.com/Azure/Industrial-IoT/files/4594339/opc-Publisher.log)

**Expected behavior**
We expected the OPC publisher to issue a OPC client certificate automatically after one year, which would ensure that we would continually send OPC data to the Azure center.

**Additional context**
Our deployment configuration for OPC Publisher attached as well
[Deployment-config.txt](https://github.com/Azure/Industrial-IoT/files/4594425/Deployment-config.txt)

cristipogacean:
fixed with #451. Please test it using :preview tagged container

PR #451, opened 2020-05-13
Hotfix/validate app cert

cristipogacean:
* fix for #430
* default certificate store to Directory for windows containers


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Gateway/src/Server/GatewayServer.cs
```diff
@@ -1608,25 +1608,25 @@ public sealed class GatewayServer : SessionServerBase, IServer {
...
+                        StoreType = CertificateStoreType.Directory,
                         StorePath = "pki/issuers",
                     },
                     RejectedCertificateStore = new CertificateTrustList {
-                        StoreType = "Directory",
+                        StoreType = CertificateStoreType.Directory,
                         StorePath = "pki/rejected"
                     },
+                    MinimumCertificateKeySize = 1024,
+                    RejectSHA1SignedCertificates = true,
```
marcschier, line 1628: We should allow sha1 for interop.  Let's add a workitem to warn user in the new cert validation API.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Runtime/Default/ClientServicesConfig.cs
```diff
@@ -63,11 +62,12 @@ public class ClientServicesConfig : ConfigBase, IClientServicesConfig, ISecurity
         /// <inheritdoc/>
         public bool RejectSha1SignedCertificates => _security.RejectSha1SignedCertificates;
```
marcschier, line 63: One could use this option.  But I would default to false and as per discussion minimize config surface.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Runtime/ISecurityConfig.cs
```diff
@@ -50,5 +50,10 @@ public interface ISecurityConfig {
         /// Trusted peer certificates
         /// </summary>
         CertificateStore TrustedPeerCertificates { get; }
+
+        /// <summary>
+        /// Automatically add application certificate to the trusted store
+        /// </summary>
+        public bool AddAppCertToTrustedStore { get; }
```
marcschier, line 57: Public is not needed.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Gateway/src/Server/GatewayServer.cs
cristipogacean: done, DevOps backlog item created as well

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Runtime/Default/ClientServicesConfig.cs
cristipogacean: it is defaulted to false, anywhere except Gateway Server, (previous comment,), where was already set to true and did not questioned the reason :)

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Runtime/ISecurityConfig.cs
cristipogacean: totally! changed

PR #460, opened 2020-05-17
Hotfix/validate app cert

cristipogacean:


modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Twin/cli/Program.cs
```diff
@@ -169,7 +169,7 @@ private enum Op {
                             Console.WriteLine($"Using <deviceId> '{deviceId}'");
                         }
                         if (moduleId == null) {
-                            moduleId = "opctwin";
+                            moduleId = "twin";
```
barnstee, line 172: I thought we agreed to move back to the name opctwin and opcpublisher?

marcschier: We cannot without breaking backwards compatibility as we have found during testing.   Old identities named twin and publisher will now hang around in the UX and look disconnected.   Implementing migration is too costly at this point.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Twin/src/ModuleProcess.cs
```diff
@@ -89,15 +90,17 @@ public class ModuleProcess : IProcessControl {
...
                     logger.Information("Initiating prometheus at port {0}/metrics", kTwinPrometheusPort);
                     var server = new MetricServer(port: kTwinPrometheusPort);
                     try {
-                        server.StartWhenEnabled(config, logger);
+                        server.StartWhenEnabled(moduleConfig, logger);
                         // Start module
                         var product = "OpcTwin_" +
                             GetType().Assembly.GetReleaseVersion().ToString();
-                        kTwinModuleStart.Inc();
+                        kTwinModuleStart.WithLabels(identity.SiteId ?? "",
```
marcschier, line 102: We should add this into the ModuleHost class.   I am not sure if these are even set at this point as I believe Start has to be called first.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Twin/cli/Program.cs
cristipogacean: @Marc, that would kind-of lead to confusion in the setup of the edge device modules, as it happened for me. One could easily end-up with both twin and opctwin and publisher and opcpublisher in the same edge device

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Twin/src/ModuleProcess.cs
cristipogacean: well, DeviceId and ModuleId are set (now, I did some changes to make it happen), the SiteId is set to the default value, but this might be overwritten with the value from the twin's definition. the ModuleHost has a different metric that is kind-of similar

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Twin/cli/Program.cs
barnstee: Can we change the UX, then? Until the last release, we had the old names anyway...

cristipogacean: @barnstee we can do decide separately and change UX in a separate PR, this change was just naming consistency alignment with the existing names.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Twin/src/ModuleProcess.cs
marcschier: I would prefer it to be in the ModuleHost start/stop methods as that is where these values are set.  And yes, site is dynamic and can change  To make sure we have correct numbers I would not add it as label, only deviceid, gw, moduleid which are fixed over the course of the module host running.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Twin/cli/Program.cs
marcschier: Unfortunately we wont be able to change the ux in Azure portal - if that was needed we would need to have a migration step during startup.   We should not show these "identity ids" in our ux anyway, there should be a name that can be changed by the user and that does not affect the id.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Twin/src/ModuleProcess.cs
cristipogacean: ModuleHost has a separate metrics on start/stop methods. I removed the SiteId now from all metrics.

common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/ModuleHost.cs
```diff
@@ -61,6 +65,9 @@ public sealed class ModuleHost : IModuleHost, ITwinProperties, IEventEmitter,
             _router = router ?? throw new ArgumentNullException(nameof(router));
             _settings = settings ?? throw new ArgumentNullException(nameof(settings));
             _serializer = serializer ?? throw new ArgumentNullException(nameof(serializer));
+            DeviceId = _factory.DeviceId;
```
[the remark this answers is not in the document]
cristipogacean: sure, I'll change right away

PR #463, opened 2020-05-18
Generate server models using model compiler

marcschier:
* Fix swagger oauth2 flow in swagger UX
* Update dependencies to fix various component governance reports
* Re-generate test server models using updated model compiler (1.04)


mregen:
>     <uax:DataTypeDefinition>

great to see this DataTypeDefinition now

---
Refers to: components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Testing/src/Servers/TestData/Design/TestData.PredefinedNodes.xml:1135 in eb71116. [](commit_id = eb71116cafcb9494dc1c032724f030fdb8138f0c, deletion_comment = False)

PR #465, opened 2020-05-18
fix method call

dacolgit:


samples/src/Microsoft.Azure.IIoT.App/src/Services/Browser.cs
```diff
@@ -330,17 +330,21 @@ public class Browser {
...
-                var count = 0;
-                foreach (var item in parameters.InputArguments) {
-                    var argument = new MethodCallArgumentApiModel {
-                        Value = parameterValues[count] ?? string.Empty,
-                        DataType = item.Type.DataType
-                    };
-                    argumentsList.Add(argument);
-                    count++;
+                if (parameters.InputArguments != null)
+                {
```
marcschier, line 334: Please do not disable .editorconfig so we have single formatting.  (see line breaks)

dacolgit: my formatting config was screwed up. Fixed

PR #481, opened 2020-05-21
Hotfix/subscription optimization

cristipogacean:


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
```diff
@@ -62,6 +62,7 @@ public class DataFlowProcessingEngine : IProcessingEngine, IDisposable {
         /// <inheritdoc/>
         public void Dispose() {
             _messageTrigger.OnMessage -= MessageTriggerMessageReceived;
+            _messageTrigger.Dispose();
```
marcschier, line 65: Autofac should be calling it.  Otherwise must override onrelease during registration or better, consume it as Owned<>

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/WriterGroupMessageSource.cs
```diff
@@ -61,17 +61,30 @@ public class WriterGroupMessageTrigger : IMessageTrigger {
 
         /// <inheritdoc/>
         public async Task RunAsync(CancellationToken ct) {
-
+/*
```
marcschier, line 64: Could be removed?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -43,39 +43,45 @@ public class DefaultSessionManager : ISessionManager {
...
+
             try {
                 // try to get an existing session
-                if (_sessions.TryGetValue(id, out var wrapper)) {
-                    if (wrapper != null && 
-                        (forceActivation || !wrapper.Session.Connected || 
-                        wrapper.MissedKeepAlives >= wrapper.MaxKeepAlives)) {
+                await _lock.WaitAsync();
+                _sessions.TryGetValue(id, out var wrapper);
+                _lock.Release();
```
marcschier, line 55: Safer would be a try finally here

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -43,39 +43,45 @@ public class DefaultSessionManager : ISessionManager {
...
-                                wrapper.Session.SessionName, ex.Message);
-                        }
-                        finally {
-                            if (dispose) {
+                                _logger.Warning("Failed to reconnect session {sessionName} due to {exception}." +
+                                    " Dispose and try create new.", wrapper.Session.SessionName, e.Message);
                                 //  todo check status codes
+                                await _lock.WaitAsync();
                                 _sessions.Remove(id);
+                                _lock.Release();
```
marcschier, line 84: Same here

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -104,7 +116,19 @@ public class DefaultSessionManager : ISessionManager {
                             wrapper = await CreateSessionAsync(endpointUrl, id);
                             if (wrapper?.Session != null) {
                                 _logger.Information("Connected on {endpointUrl}", endpointUrl);
-                                _sessions.Add(id, wrapper);
+                                await _lock.WaitAsync();
+                                if (!_sessions.TryGetValue(id, out var original)) {
+                                    _sessions.TryAdd(id, wrapper);
+                                }
+                                _lock.Release();
```
marcschier, line 123: Same here

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -226,7 +248,6 @@ public class DefaultSessionManager : ISessionManager {
         /// <inheritdoc/>
         public async Task RemoveSessionAsync(ConnectionModel connection, bool onlyIfEmpty = true) {
             var key = new ConnectionIdentifier(connection);
-            await _lock.WaitAsync();
             try {
                 if (!_sessions.TryGetValue(key, out var wrapper) || wrapper?.Session == null) {
```
marcschier, line 252: Should be under lock

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -235,8 +256,9 @@ public class DefaultSessionManager : ISessionManager {
                 if (onlyIfEmpty && session.SubscriptionCount > 0) {
                     return;
                 }
+                await _lock.WaitAsync();
                 _sessions.Remove(key);
-
+                _lock.Release();
```
marcschier, line 261: Best to call as finally.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -283,42 +305,27 @@ public class DefaultSessionManager : ISessionManager {
         private void Session_KeepAlive(Session session, KeepAliveEventArgs e) {
             _logger.Debug("Keep Alive received from session {name}, state: {state}.",
                 session.SessionName, e.CurrentState);
-            _lock.Wait();
             try {
-                foreach (var entry in _sessions
-                    .Where(s => s.Value.Session.SessionId == session.SessionId).ToList()) {
+                _lock.WaitAsync().ConfigureAwait(false);
```
marcschier, line 309: Must await and call release as finally.

cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/WriterGroupMessageSource.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -226,20 +265,21 @@ public class DefaultSessionManager : ISessionManager {
         /// <inheritdoc/>
         public async Task RemoveSessionAsync(ConnectionModel connection, bool onlyIfEmpty = true) {
             var key = new ConnectionIdentifier(connection);
-            await _lock.WaitAsync();
             try {
+                await _lock.WaitAsync();
```
marcschier, line 269: Nit - i'd suggest doing the acquire outside and let it throw as that prevents the Release() from being called in that case, i.e. when wait did not capture the semaphore you do not want to release it (and end up with a bad semaphore count)

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -280,45 +323,36 @@ public class DefaultSessionManager : ISessionManager {
         /// </summary>
         /// <param name="session"></param>
         /// <param name="e"></param>
-        private void Session_KeepAlive(Session session, KeepAliveEventArgs e) {
+        private async void Session_KeepAlive(Session session, KeepAliveEventArgs e) {
```
marcschier, line 326: Unless you can disable keep alive callbacks firing I would suggest in this case to just use Wait() as you do not want x tasks from being scheduled for every keep alive callback that happens while this task is executing.   

Similar to timers, where you want to stop the timer while the task is executing and restart when the task is done.   (To prevent x tasks queing up on every timer fire, while the thread pool is exhausted).

In general async void should be avoided as much as possible, but in callback case it is not possible.   But we need to make sure that exceptions are caught, or else the entire process is getting killed.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -40,6 +40,67 @@ public class SubscriptionServices : ISubscriptionManager, IDisposable {
...
+            }
+        }
+
+        /// <summary>
+        /// Check connectivity
+        /// </summary>
+        private async void OnCheckAsync(object sender) {
+            var success = true;
+            try {
+                Try.Op(() => _timer.Change(Timeout.InfiniteTimeSpan, Timeout.InfiniteTimeSpan));
```
marcschier, line 63: This is an example re above, a) fully try catch log (even though log could theoratically throw), b) disable firing of timer so this task can first run and then re-arm the timer for the next task. no action at this point.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -54,16 +115,14 @@ public class SubscriptionServices : ISubscriptionManager, IDisposable {
 
         /// <inheritdoc/>
         public void Dispose() {
+            _timer.Change(Timeout.InfiniteTimeSpan, Timeout.InfiniteTimeSpan);
```
marcschier, line 118: Of course there are still issues as you can see here - the timer task above might just be executing while we run this code.   Hence trying to catch whatever we can.   But we need a better async timer task abstraction going forward... - no action at this point.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
marcschier: I did not suggest this was not valid - e.g. if you want to control the order of dispose call then it makes sense to take ownership of the instance passed to you.

cristipogacean: actually this helped on a different bug that was still open. Actually the original problem was that the Message trigger was not implementing the IDisposable...

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
cristipogacean: right

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
cristipogacean: thanks this is really helpful!

PR #483, opened 2020-05-22
Changes to metrics and workbook

monikavar:
- Fixed labels
- Updated queries in workbook correspondingly
- Added module version and restart count
- Fixed published messages count


common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/ModuleHost.cs
```diff
@@ -91,6 +91,9 @@ public sealed class ModuleHost : IModuleHost, ITwinProperties, IEventEmitter,
                     _logger.Error(ce, "Module Host stopping caused exception.");
                 }
                 finally {
+                    kModuleStart.WithLabels(DeviceId ?? "", ModuleId ?? "", ModuleGuid, "",
+                        DateTime.UtcNow.ToString("yyyy-MM-dd'T'HH:mm:ss.FFFFFFFK", 
+                        CultureInfo.InvariantCulture)).Set(0);
                     kModuleStart.WithLabels(DeviceId ?? "", ModuleId ?? "",
                         DateTime.UtcNow.ToString("yyyy-MM-dd'T'HH:mm:ss.FFFFFFFK",
                         CultureInfo.InvariantCulture)).Set(0);
```
cristipogacean, line 99: why is duplicate? btw the guid is missing causing probably   an exception.
Perhaps we should always guard setting the metrics with labels in try ... catch blocks

monikavar: missed it

common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/ModuleHost.cs
```diff
@@ -538,10 +540,11 @@ public sealed class ModuleHost : IModuleHost, ITwinProperties, IEventEmitter,
         private readonly SemaphoreSlim _lock = new SemaphoreSlim(1, 1);
         private readonly Dictionary<string, VariantValue> _reported =
             new Dictionary<string, VariantValue>();
+        private readonly string ModuleGuid = Guid.NewGuid().ToString();
         private static readonly Gauge kModuleStart = Metrics
             .CreateGauge("iiot_edge_module_start", "starting module",
                 new GaugeConfiguration {
-                    LabelNames = new[] {"deviceid", "module", "timestamp_utc" }
+                    LabelNames = new[] {"deviceid", "module", "guid", "version", "timestamp_utc" }
```
cristipogacean, line 547: >"guid" [](start = 62, length = 6)

is this label that is visible to the customer? if yes, perhaps we should find a better name

common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/ModuleHost.cs
```diff
@@ -538,10 +540,11 @@ public sealed class ModuleHost : IModuleHost, ITwinProperties, IEventEmitter,
         private readonly SemaphoreSlim _lock = new SemaphoreSlim(1, 1);
         private readonly Dictionary<string, VariantValue> _reported =
             new Dictionary<string, VariantValue>();
+        private readonly string ModuleGuid = Guid.NewGuid().ToString();
```
cristipogacean, line 543: >ModuleGuid [](start = 32, length = 10)

coding conventions: either _moduleGuid, or make it a property.

Or maybe better have it globally  visible as part of the Instance so that all the metrics can reuse it. Then we would know all thing were generated by a single runtime instance

common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/ModuleHost.cs
```diff
@@ -91,8 +91,8 @@ public sealed class ModuleHost : IModuleHost, ITwinProperties, IEventEmitter,
                     _logger.Error(ce, "Module Host stopping caused exception.");
                 }
                 finally {
-                    kModuleStart.WithLabels(DeviceId ?? "", ModuleId ?? "",
-                        DateTime.UtcNow.ToString("yyyy-MM-dd'T'HH:mm:ss.FFFFFFFK",
+                    kModuleStart.WithLabels(DeviceId ?? "", ModuleId ?? "", ModuleGuid, "",
+                        DateTime.UtcNow.ToString("yyyy-MM-dd'T'HH:mm:ss.FFFFFFFK",
```
cristipogacean, line 95: >    [](start = 21, length = 3)

missed the version?

cristipogacean: I think the version is still missing here

PR #484, opened 2020-05-22
fix paging (activate an endpoint on a page different than the first re…

dacolgit:
…direct to first page)


samples/src/Microsoft.Azure.IIoT.App/src/Pages/Applications.razor
```diff
@@ -141,10 +142,9 @@
     /// </summary>
     /// <param name="applicationId"></param>
     private async Task UnregisterApplicationUIAsync(string applicationId)
-    {
+{
```
MaidaB, line 145: indentation

PR #486, opened 2020-05-22
Make ux work on minimum deployment and fix logs in events service

marcschier:


services/src/Microsoft.Azure.IIoT.Services.OpcUa.Events/src/Runtime/Config.cs
```diff
@@ -75,7 +75,11 @@ public class Config : DiagnosticsConfig, IWebHostConfig, IServiceBusConfig,
         /// <inheritdoc/>
         public string ConsumerGroup => GetStringOrDefault(
             PcsVariable.PCS_EVENTHUB_CONSUMERGROUP_TELEMETRY_UX,
-                () => "telemetryux");
+                () => IsMinimumDeployment ? "$default" : "telemetryux");
```
karok2m, line 78: `azuredeploy.minimum.json` already has the secondary Event Hub. Why not add `telemetryux` and `telemetrycdm` consumer groups there instead of using  `$default` ?

marcschier: My first approach, but I want to keep the min eventhub totally clean of ux or cdm stuff.  Running ux on it is just a debug tool.

PR #492, opened 2020-05-26
Fit content to the window

MaidaB:
Now the tables are fixed to the size of the window. If some text cannot fit in a table cell it will be hidden using ellipsis.
image

When you hover over the text it will expand.
image

There is one issue that I couldn't have solved. When you hover over the text which is smaller than the cell width, the table border disappears. The reason for this is this setting: display: inline-block. Without this setting, the long text cannot overwrite the text from the adjacent cell.
image

Another issue is that the width of all columns is the same, even when their content is just 2-3 characters. That's because of this setting: table-layout: fixed. However, this setting is necessary for keeping the table inside of the window.


samples/src/Microsoft.Azure.IIoT.App/src/wwwroot/css/site.css
```diff
@@ -103,6 +103,33 @@ app {
...
+td {
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+}
+
+td:hover {
+    overflow: visible;
+    position: relative;
+    display: inline-block;
```
dacolgit, line 120: to fix the issue of the table border you can try to use z-index in td:hover
setting a z-index higher than the normal and remove the inline-block

MaidaB: It didn't work. I'll open a new bug for this

PR #493, opened 2020-05-26
Fix duplication of scan events

dacolgit:
Fix navigation back from published nodes page
Fix value update for published variables


samples/src/Microsoft.Azure.IIoT.App/src/Pages/Browser.razor
```diff
@@ -433,6 +437,10 @@
                 SamplingInterval = samplingInterval,
                 HeartbeatInterval = heartbeatInterval
             };
+            if (_publishEvent == null) {
+                _publishEvent = Task.Run(async () => await PublisherServiceEvents.NodePublishSubscribeByEndpointAsync(EndpointId,
```
marcschier, line 441: Why is Task.Run needed?

samples/src/Microsoft.Azure.IIoT.App/src/Pages/Browser.razor
```diff
@@ -307,8 +308,11 @@
             CommonHelper.Spinner = string.Empty;
             CommonHelper.CheckErrorOrEmpty<ListNode>(PagedNodeList, ref _tableView, ref _tableEmpty);
             StateHasChanged();
-            _publishEvent = await PublisherServiceEvents.NodePublishSubscribeByEndpointAsync(EndpointId,
+            if (PublishedNodes.Results.Count > 0)
```
marcschier, line 311: Is it possible that the count is > 0 on first render?  And if so, shouldn't you test for _publishEvent == null?

samples/src/Microsoft.Azure.IIoT.App/src/Pages/Browser.razor
dacolgit: because if you quickly select multiple node to be published in the UI  you will get multiple subscription to the same event. Because the subscription is for endpointId we need just one subscription for all the published nodes for that endpoint

samples/src/Microsoft.Azure.IIoT.App/src/Pages/Browser.razor
dacolgit: the count can be > 0 at the first render. If you have an endpoint with previously published nodes, when you enter on this page the BrowseTreeAsync gives you the list of published nodes.
_publishevent on first render is always null, and if if there are already published nodes is initialized on first render, otherwise when you publish a node

samples/src/Microsoft.Azure.IIoT.App/src/Pages/Browser.razor
marcschier: Not sure I understand why this means you call Task.Run, which just puts the registration on a task in thread pool.   What is the difference to just calling await … Subscribe() other than running outside of the rendering?   Is it blocking?

dacolgit: The key is using Result (that blocks until method has completed), and I'm using Task.run to not block UI rendering

PR #494, opened 2020-05-26
metrics for response time of cloud micro-services API calls

monikavar:
- use in-built middleware to expose response time of all API calls for all services


common/src/Microsoft.Azure.IIoT.AspNetCore/src/ResponseTime/Extensions/ResponseTimeMiddlewareEx.cs
```diff
@@ -0,0 +1,22 @@
...
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+namespace Microsoft.Azure.IIoT.AspNetCore.Monitoring {
+    using Microsoft.AspNetCore.Builder;
+
+    /// <summary>
+    ///Response time middleware extensions
+    /// </summary>
+    public static class ResponseTimeMiddlewareEx {
```
marcschier, line 12: Couldn't we just use UseHttpMetrics() (https://github.com/prometheus-net/prometheus-net/blob/master/Prometheus.AspNetCore/HttpMetricsMiddlewareExtensions.cs) from Prometheus.aspnetcore package which we are already linking?

monikavar: Didn't know about it. Thanks for sharing.

PR #520, opened 2020-06-02
add docfx project to generate C# API Reference

koepalex:
* using docfx to be able to generate full API reference
* updated .gitignore to ignore the build results of docfx
* added reference from docs/readme.md


docs/csharp-api/index.md
```diff
@@ -0,0 +1,24 @@
+# Industrial IoT Platform C# Reference
```
marcschier, line 1: Q: Why not put this under api/ folder?

docs/csharp-api/docfx.json
```diff
@@ -0,0 +1,65 @@
+{
+  "metadata": [
+    {
+      "src": [
+        {
+          "files": [
+            "**.csproj"
```
marcschier, line 7: Isnt this better to just document projects under api/, which are public API to be consumed?

docs/csharp-api/index.md
koepalex: Moved under api folder, which is the better place

docs/csharp-api/docfx.json
koepalex: agreed, updated docfx config to only work with projects under api folder

PR #521, opened 2020-06-03
Fix small issues/typos, reformat diagnostic info

luiscantero:


common/src/Microsoft.Azure.IIoT.Core/tests/Utils/LoggingTests.cs
```diff
@@ -0,0 +1,52 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+namespace Microsoft.Azure.IIoT.Core.Tests {
+    using Xunit;
+    using static Microsoft.Azure.IIoT.Utils.Logging;
```
marcschier, line 8: Great job on the unit test.  But "using static" is not a good pattern to use.   It hides too much and makes the code unmaintainable and difficult to refactor.  Here it should be straightforward to use "Logging." where needed.

common/src/Microsoft.Azure.IIoT.Core/src/Utils/Logging.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// </summary>
+    public static class Logging {
+        /// <summary>
+        /// Extracts server:port from a uri for simpler logging.
+        /// </summary>
+        /// <param name="endpointUrl"></param>
+        /// <returns></returns>
+        public static string ExtractServerPort(string endpointUrl) {
+            string pattern = @":\/\/([^\/_]+)";
+            var match = Regex.Match(endpointUrl, pattern);
```
marcschier, line 21: Why not just "new Url(endpointUrl).Port"?   Saves you all the tests, etc.?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
```diff
@@ -151,36 +153,36 @@ public class DataFlowProcessingEngine : IProcessingEngine, IDisposable {
...
-                _messageEncoder.MessagesProcessedCount,
-                _messageEncoder.AvgNotificationsPerMessage,
-                _messageEncoder.AvgMessageSize,
-                _batchNetworkMessageBlock.OutputCount,
-                _sinkBlock.InputCount,
-                _messageSink.SentMessagesCount,
-                _messageSink.SentMessagesCount / totalDuration,
-                _messageTrigger.NumberOfConnectionRetries);
+            if (_messageTrigger.DataChangesCount > 0 || _messageTrigger.ValueChangesCount > 0 || _messageSink.SentMessagesCount > 0)
+            {
```
marcschier, line 157: I think this was using structured logging for a reason as Cristian only just recently changed it.   

Also please use .editorconfig formatting as rest of code does.  Do not override formatting settings or else we end up with a hodgepodge of formats in our code.

samples/src/Microsoft.Azure.IIoT.App/src/Pages/_DrawerDiscoverer.razor
```diff
@@ -18,11 +18,11 @@
     <div class="list-group-item text-left">
         <form>
             <div class="form-group">
-                <label>Mask address range (CIDR notation)</label>
+                <label>Mask Address Range (CIDR notation)</label>
```
marcschier, line 21: I think Maida just changed it to lowercase - please sync with her and Daniele wrt the changes.

common/src/Microsoft.Azure.IIoT.Core/src/Utils/Logging.cs
luiscantero: I renamed the file to LoggingHelper.cs and the method to ExtractHost. The name of the method was misleading as it doesn't extract the port only. Some input types such as "http://example.com:8080_hash" would not work with the Uri class.

common/src/Microsoft.Azure.IIoT.Core/tests/Utils/LoggingTests.cs
luiscantero: "using static" is used here only to simplify writing of tests cases. Should it be avoided even in test files?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
luiscantero: Fixed

common/src/Microsoft.Azure.IIoT.Core/src/Utils/Logging.cs
marcschier: I think we should have this very "specialized" code somewhere closer to its actual use - and I assume it is only used once, correct?   My preference:  Create real url in publisher as identifier, and then extract what is necessary from it using the existing parsers.   E.g. _ could be just a / and we should be able to use Url(…).

samples/src/Microsoft.Azure.IIoT.App/src/Pages/_DrawerDiscoverer.razor
luiscantero: This fix is in agreement with Maida and Daniele

common/src/Microsoft.Azure.IIoT.Core/src/Utils/Logging.cs
luiscantero: Here it's only used once, but it will be used more once Cristian is done refactoring the session management. I'm waiting for him to avoid merging overhead. This helper also gives us a way to change the formatting centrally in the future.

common/src/Microsoft.Azure.IIoT.Core/src/Utils/LoggingHelper.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// <summary>
+    /// Helper class for logging
+    /// </summary>
+    public static class LoggingHelper {
+        /// <summary>
+        /// Extracts host in the format "server[:port]" from a uri for simpler logging.
+        /// </summary>
+        /// <param name="endpointUrl"></param>
+        /// <returns></returns>
+        public static string ExtractHost(string endpointUrl) {
```
MaidaB, line 19: How often is called this function? Should we use the compiled regex?

common/src/Microsoft.Azure.IIoT.Core/tests/Utils/LoggingTests.cs
marcschier: We can debate this as team but I am truly allergic to it :-)

common/src/Microsoft.Azure.IIoT.Core/src/Utils/Logging.cs
marcschier: Then let's move it into the same package as where it is used as it feels very "publisher-ish".

luiscantero: Done

common/src/Microsoft.Azure.IIoT.Core/tests/Utils/LoggingTests.cs
luiscantero: Done

PR #531, opened 2020-06-04
Hotfix/session reconnect

cristipogacean:


luiscantero:
There seems to be some kind of issue. I left it running from about 0:45 and it crashed at 6:30 in the morning without an error (log file doesn't show anything unusual at the end).
image
The sdk error BadRequestTimeout still shows up a number of times. This is of course an edge case and imho this is a better overall engine than what we had before.

PR #539, opened 2020-06-05
Fix some sporadic NullReferenceExceptions, simplify/reword logs

luiscantero:
* Fix NullReferenceException with semaphoreslim
* Fix NullReferenceException with number of connection retries
* Fix NullReferenceException in MonitoredItemMessage
* Fix NullReferenceException on triggerBatch
* Simplify logging messages by setting some information to verbose
* Use consistent wording and remove period
* Remove unnecessary ' when using structured logging
* Add ConfigureAwait(false) to session/subscription related calls

### DIFF
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
@@ -79,14 +79,14 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
             await _lock.WaitAsync();
             try {
                 if (_cts != null) {
-                    _logger.Warning("Worker already running");
+                    _logger.Verbose("Worker already running");
                     return;
                 }
 
                 _cts = new CancellationTokenSource();
                 _heartbeatTimer.Change(TimeSpan.Zero, Timeout.InfiniteTimeSpan);
 
-                _logger.Information("Worker WorkerId: {WorkerId} {@Capabilities}",
+                _logger.Verbose("Worker {WorkerId}: {@Capabilities}",
                     WorkerId, _agentConfigProvider.Config.Capabilities);
                 _worker = Task.Run(() => RunAsync(_cts.Token));
             }
@@ -189,7 +189,7 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
                     _logger.Information("Worker cancelled...");
                 }
                 catch (Exception ex) {
-                    // TODO: we should notify the exception 
+                    // TODO: we should notify the exception
                     _logger.Error(ex, "Worker: {Id}, exception during worker processing, wait {delay}...",
                         WorkerId, _jobCheckerInterval);
                     kModuleExceptions.WithLabels(AgentId, ex.Source, ex.GetType().FullName, ex.Message, ex.StackTrace, "Exception during worker processing").Inc();
@@ -209,7 +209,7 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
                 // Stop worker heartbeat to start the job heartbeat process
                 _heartbeatTimer.Change(-1, -1); // Stop worker heartbeat
 
-                _logger.Information("Worker: {WorkerId}, start processing new job: {JobId}, mode: {ProcessMode}",
+                _logger.Information("Worker: {WorkerId} processing job: {JobId}, mode: {ProcessMode}",
                     WorkerId, jobProcessInstruction.Job.Id, jobProcessInstruction.ProcessMode);
 
                 // Execute processor
@@ -234,7 +234,7 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
                         _jobProcess = null;
                         break;
                     }
-                    _logger.Information("Processing job continuation...");
+                    _logger.Verbose("Processing job continuation...");
                 }
             }
             finally {
@@ -324,7 +324,7 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
                     _logger.Information("Job {job} completed.", Job.Id);
                 }
                 catch (OperationCanceledException) {
-                    _logger.Information("Job {job} cancelled.", Job.Id);
+                    _logger.Verbose("Job {job} cancelled.", Job.Id);
                 }
                 catch (Exception ex) {
                     _logger.Error(ex, "Error processing job {job}.", Job.Id);
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
@@ -79,14 +79,14 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
             await _lock.WaitAsync();
             try {
                 if (_cts != null) {
-                    _logger.Verbose("Worker already running");
+                    _logger.Warning("Worker already running");
                     return;
                 }
 
                 _cts = new CancellationTokenSource();
                 _heartbeatTimer.Change(TimeSpan.Zero, Timeout.InfiniteTimeSpan);
 
-                _logger.Verbose("Worker {WorkerId}: {@Capabilities}",
+                _logger.Information("Worker {WorkerId}: {@Capabilities}",
                     WorkerId, _agentConfigProvider.Config.Capabilities);
                 _worker = Task.Run(() => RunAsync(_cts.Token));
             }
@@ -234,7 +234,7 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
                         _jobProcess = null;
                         break;
                     }
-                    _logger.Verbose("Processing job continuation...");
+                    _logger.Information("Processing job continuation...");
                 }
             }
             finally {
@@ -324,7 +324,7 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
                     _logger.Information("Job {job} completed.", Job.Id);
                 }
                 catch (OperationCanceledException) {
-                    _logger.Verbose("Job {job} cancelled.", Job.Id);
+                    _logger.Information("Job {job} cancelled.", Job.Id);
                 }
                 catch (Exception ex) {
                     _logger.Error(ex, "Error processing job {job}.", Job.Id);
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
@@ -110,12 +110,12 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
             while (true) {
                 var workers = _agentConfigProvider.Config?.MaxWorkers ?? kDefaultWorkers;
                 while (_instances.Count < workers) {
-                    _logger.Verbose("Creating new worker...");
+                    _logger.Information("Creating new worker...");
                     var worker = await CreateWorker();
                 }
 
                 foreach (var stoppedWorker in _instances.Keys.Where(s => s.Status == WorkerStatus.Stopped)) {
-                    _logger.Verbose("Starting worker {workerId}...", stoppedWorker.WorkerId);
+                    _logger.Information("Starting worker '{workerId}'...", stoppedWorker.WorkerId);
                     workerStartTasks.Add(stoppedWorker.StartAsync());
                 }
                 await Task.WhenAll(workerStartTasks);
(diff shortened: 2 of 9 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Stack/Extensions/NodeIdEx.cs
```diff
@@ -27,12 +27,15 @@ public static class NodeIdEx {
...
+
+            ushort namespaceIndex = nodeId.NamespaceIndex;
+            if (namespaceIndex > 0 && namespaces == null) {
+                namespaceIndex = 0;
+            }
+            return new ExpandedNodeId(nodeId.Identifier, namespaceIndex,
+                namespaceIndex > 0
+                ? namespaces.GetString(namespaceIndex)
+                : null, 0);
         }
```
cristipogacean, line 39: We need to revert this, the namespace 0 has a special meaning, being reserved for the OPC Foundation spec nodes. A namespace name must be provided  if the namespace index is != 0. if the table is not provided is a fatal error and we cannot continue, therefore the exception is required here. Changing all namespaces indexes to 0 in case the table is not present would mess up everything

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -107,9 +107,9 @@ internal sealed class SubscriptionWrapper : ISubscription {
 
             /// <inheritdoc/>
             public async Task CloseAsync() {
-                await _lock.WaitAsync();
+                await _lock.WaitAsync().ConfigureAwait(false);
                 try {
-                    _logger.Information("Closing subscription {subscription}", Id);
```
cristipogacean, line 112: >Information [](start = 28, length = 11)

I would generally prefer to keep these kind of logs more visible. They will only happen once during runtime so they are not that chatty. On the other hand, if something goes wrong and everything is verbose you have no chance follow any hind of flow and to help the customer.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -195,12 +195,12 @@ internal sealed class SubscriptionWrapper : ISubscription {
...
                         ResolveDisplayNameAsync(_subscription.MonitoredItems);
                     }
-                    await ReapplyAsync(session, false);
+                    await ReapplyAsync(session, false).ConfigureAwait(false);
                     Enabled = true;
                     Active = false;
                 }
                 catch (Exception e) {
-                    _logger.Error(e, "Failed to enable subscription");
+                    _logger.Verbose(e, "Failed to enable subscription");
```
cristipogacean, line 203: >Verbose [](start = 28, length = 7)

generally for exceptions I think is a good idea to make such exceptions visible. In this particular case we do not expect exceptions and if they are caught something went seriously bad and we want to know ... the logic here was totally changed and we are no longer relying on exceptions

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -643,7 +643,7 @@ internal sealed class SubscriptionWrapper : ISubscription {
                     }
 
                     if (_currentlyMonitored == null) {
-                        _logger.Information("DataChange for subscription: {Subscription} having no monitored items yet",
+                        _logger.Verbose("DataChange for subscription: {Subscription} having no monitored items yet",
```
cristipogacean, line 646: >Verbose [](start = 32, length = 7)

this should not happen normally, therefore we need to know it happens. Again the logic was changed and this should never appear

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -163,7 +163,7 @@ public class DefaultSessionManager : ISessionManager, IDisposable {
                     return;
                 }
                 if (wrapper._subscriptions.TryRemove(subscription.Id, out _)) {
-                    _logger.Information("Subscription '{subscriptionId}' unregistered from session '{sessionId}' in state {state}",
+                    _logger.Verbose("Subscription {subscriptionId} unregistered from session {sessionId} in state {state}",
```
cristipogacean, line 166: >Verbose [](start = 28, length = 7)

this is typically is happening maximum once during runtime if the cleanup works properly, I would recommend to keep it at least at debug level

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
```diff
@@ -110,12 +110,12 @@ public class WorkerSupervisor : IWorkerSupervisor, IDisposable {
             while (true) {
                 var workers = _agentConfigProvider.Config?.MaxWorkers ?? kDefaultWorkers;
                 while (_instances.Count < workers) {
-                    _logger.Information("Creating new worker...");
+                    _logger.Verbose("Creating new worker...");
```
cristipogacean, line 113: >Verbose [](start = 28, length = 7)

this is typically only happening once during runtime, I think we can keep it to Information level, same below

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
```diff
@@ -110,12 +110,12 @@ public class WorkerSupervisor : IWorkerSupervisor, IDisposable {
...
                     var worker = await CreateWorker();
                 }
 
                 foreach (var stoppedWorker in _instances.Keys.Where(s => s.Status == WorkerStatus.Stopped)) {
-                    _logger.Information("Starting worker '{workerId}'...", stoppedWorker.WorkerId);
```
cristipogacean, line 118: >'{ [](start = 57, length = 2)

I guess generally these quotes mean the parameter is a string. I don't exactly know if this is more than a convention, it is at least a convention. Do you have a reason to change this? you know the saying, "When in Rome, do as romans do" :)

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
```diff
@@ -79,14 +79,14 @@ public class Worker : IWorker, IDisposable {
             await _lock.WaitAsync();
             try {
                 if (_cts != null) {
-                    _logger.Warning("Worker already running");
+                    _logger.Verbose("Worker already running");
```
cristipogacean, line 82: >Verbose [](start = 28, length = 7)

that's more like an error condition, when it happens, we should know it happened.

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
```diff
@@ -79,14 +79,14 @@ public class Worker : IWorker, IDisposable {
...
                     return;
                 }
 
                 _cts = new CancellationTokenSource();
                 _heartbeatTimer.Change(TimeSpan.Zero, Timeout.InfiniteTimeSpan);
 
-                _logger.Information("Worker WorkerId: {WorkerId} {@Capabilities}",
+                _logger.Verbose("Worker {WorkerId}: {@Capabilities}",
```
cristipogacean, line 89: >Verbose [](start = 24, length = 7)

that's something that happens typically once during runtime. if happens more often might be a god indication for debugging

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
luiscantero: In a test run with 75 servers/90 nodes, in less than 10 min it generated this amount:
image
I can revert if you insist :)

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -142,7 +142,7 @@ public class DefaultSessionManager : ISessionManager, IDisposable {
                     _sessions.Add(id, wrapper);
                 }
                 wrapper._subscriptions.AddOrUpdate(subscription.Id, subscription);
-                _logger.Information("Subscription '{id}' registered in session '{id}' in state {state}",
+                _logger.Verbose("Subscription {subscriptionId} registered in session {id} in state {state}",
```
cristipogacean, line 145: >Verbose [](start = 24, length = 7)

Typically this should only happen once during runtime. I think is a good idea to keep it as information so that we can follow the flow in a debugging case.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
luiscantero: I will revert this, only 10 hits in 10 min

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
luiscantero: Will revert, only 14 hits

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
luiscantero: 2825 hits in 10 min, do you insist?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
cristipogacean: typically when subscriptions/sessions get closed you're loosing data. If we do not trace such things the customer will only notice he lost data, but will have no clue why. and we will have to answer why ...

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
luiscantero: The reasoning here is that it's a bit redundant:
1. Creating new worker...
2. Starting worker 'StandalonePublisher_1'...
3. Worker: StandalonePublisher_1, start processing new job: Standalone_test_publisher, mode: "Active"

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
luiscantero: Quotes are usually used as a delimiter when there is no structured logging. It's also not used consistently. I can bring it back consistently if you insist :)

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
luiscantero: 306 hits in 10 min, is it so informative that we need it? for an error condition, it happens a lot

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
luiscantero: 36 hits, will revert

cristipogacean:
Please don't mind the share number of comments, I think the PR is very good contribution and truly appreciate it! you did a great job testing the publisher to the limits! My main concern is that we will face very often communication and interoperability issues with misbehaving servers and if we don't have traces we will not be able to help and debug.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
luiscantero: 5530 hits in 10 min...

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
cristipogacean: ok, that's wrong we need to investigate why ...

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
cristipogacean: totally agree about consistency ... I don't mind how we do it as long as is consistent :). As I said, I don't know why it was used like that, I was curious what is the reason to change it.

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
cristipogacean: then we need to find out why and not hide it :)

cristipogacean:
sorry about that:  cristipogacean closed this 22 minutes ago, pressed the wrong button :(

PR #540, opened 2020-06-07
Hotfix/publisher fixes

cristipogacean:
* fix for bad formatted heartbeat interval
* fixed trusted certificates
* fixed publisher graceful cleanup

### DIFF
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
@@ -115,6 +115,9 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
                 _logger.Information("Worker stopped.");
             }
             catch (OperationCanceledException) { }
+            catch (Exception e) {
+                _logger.Error(e, "Stopping worker failed.");
+            }
             finally {
                 _cts?.Dispose();
                 _cts = null;
@@ -218,7 +221,7 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
                     ct.ThrowIfCancellationRequested();
                     using (_jobProcess = new JobProcess(this, jobProcessInstruction,
                         _lifetimeScope, _logger)) {
-                        await _jobProcess.WaitAsync(); // Does not throw
+                        await _jobProcess.WaitAsync(ct); // Does not throw
                     }
 
                     // Check if the job is to be continued with new configuration settings
@@ -236,6 +239,10 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
                     }
                     _logger.Information("Processing job continuation...");
                 }
+            }
+            catch (OperationCanceledException) {
+                _logger.Information("Processing cancellation received ...");
+                _jobProcess = null;
             }
             finally {
                 _logger.Information("Worker: {WorkerId}, Job: {JobId} processing completed ... ",
@@ -294,8 +301,8 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
             /// Wait till completion or heartbeat cancelling
             /// </summary>
             /// <returns></returns>
-            public Task WaitAsync() {
-                return _processor.ContinueWith(_ => Task.CompletedTask);
+            public Task WaitAsync(CancellationToken ct) {
+                return _processor.ContinueWith(_ => Task.CompletedTask, ct);
             }
 
             /// <inheritdoc/>
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
@@ -221,7 +221,7 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
                     ct.ThrowIfCancellationRequested();
                     using (_jobProcess = new JobProcess(this, jobProcessInstruction,
                         _lifetimeScope, _logger)) {
-                        await _jobProcess.WaitAsync(ct); // Does not throw
+                        await _jobProcess.WaitAsync(ct).ConfigureAwait(false); // Does not throw
                     }
 
                     // Check if the job is to be continued with new configuration settings
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
@@ -43,14 +43,16 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
 
         /// <inheritdoc/>
         public async Task StopAsync() {
+            _logger.Information("Stopping worker supervisor");
             var stopTasks = new List<Task>();
             _ensureWorkerRunningTimer.Stop();
 
-            foreach (var instance in _instances) {
+            foreach (var instance in _instances.ToList()) {
                 stopTasks.Add(instance.Key.StopAsync());
             }
 
             await Task.WhenAll(stopTasks);
+            _logger.Information("Worker supervisor successfully stopped");
         }
 
         /// <inheritdoc/>
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
@@ -56,13 +56,11 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine {
             if (_config.MaxMessageSize.HasValue && _config.MaxMessageSize.Value > 0) {
                 _maxEncodedMessageSize = _config.MaxMessageSize.Value;
             }
-            
-            _diagnosticInterval = _config.DiagnosticsInterval.GetValueOrDefault(TimeSpan.MaxValue);
-            _batchTriggerInterval = _config.BatchTriggerInterval.GetValueOrDefault(TimeSpan.MaxValue);
-            _diagnosticsOutputTimer = new Timer(DiagnosticsOutputTimer_Elapsed, null,
-                Timeout.Infinite, Timeout.Infinite);
-            _batchTriggerIntervalTimer = new Timer(BatchTriggerIntervalTimer_Elapsed, null,
-                Timeout.Infinite, Timeout.Infinite);
+
+            _diagnosticInterval = _config.DiagnosticsInterval.GetValueOrDefault(TimeSpan.Zero);
+            _batchTriggerInterval = _config.BatchTriggerInterval.GetValueOrDefault(TimeSpan.Zero);
+            _diagnosticsOutputTimer = new Timer(DiagnosticsOutputTimer_Elapsed);
+            _batchTriggerIntervalTimer = new Timer(BatchTriggerIntervalTimer_Elapsed);
         }
 
         /// <inheritdoc/>
@@ -89,6 +87,9 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine {
                 IsRunning = true;
                 _encodingBlock = new TransformManyBlock<DataSetMessageModel[], NetworkMessageModel>(
                     async input => {
+                        if (_batchTriggerInterval > TimeSpan.Zero) {
+                            _batchTriggerIntervalTimer.Change(_batchTriggerInterval, Timeout.InfiniteTimeSpan);
+                        }
                         try {
                             if (_dataSetMessageBufferSize == 1) {
                                 return await _messageEncoder.EncodeAsync(input, _maxEncodedMessageSize).ConfigureAwait(false);
@@ -120,13 +121,16 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine {
 
                 _sinkBlock = new ActionBlock<NetworkMessageModel[]>(
                     async input => {
-                        await _messageSink.SendAsync(input).ConfigureAwait(false);
-                        if (_batchTriggerInterval > TimeSpan.Zero) {
-                            _batchTriggerIntervalTimer.Change(_batchTriggerInterval, _batchTriggerInterval);
+                        if (input != null && input.Any()) {
+                            await _messageSink.SendAsync(input).ConfigureAwait(false);
+                        }
+                        else {
+                            _logger.Information("Sink block in engine {Name} triggered with empty input",
+                                Name);
                         }
                     },
                     new ExecutionDataflowBlockOptions {
-                        CancellationToken = cancellationToken,
+                        CancellationToken = cancellationToken
                     });
                 _batchDataSetMessageBlock.LinkTo(_encodingBlock);
                 _encodingBlock.LinkTo(_batchNetworkMessageBlock);
@@ -145,11 +149,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine {
                 _messageTrigger.OnMessage -= MessageTriggerMessageReceived;
                 _diagnosticsOutputTimer.Change(Timeout.Infinite, Timeout.Infinite);
                 _batchTriggerIntervalTimer.Change(Timeout.Infinite, Timeout.Infinite);
-                await Task.WhenAll(
-                    _sinkBlock.Completion,
-                    _batchDataSetMessageBlock.Completion,
-                    _batchNetworkMessageBlock.Completion,
-                    _encodingBlock.Completion);
+                await _sinkBlock.Completion;
             }
         }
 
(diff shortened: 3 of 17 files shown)


modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/ModuleProcess.cs
```diff
@@ -119,12 +121,18 @@ public class ModuleProcess : IProcessControl {
                         logger.Error(ex, "Error during module execution - restarting!");
                     }
                     finally {
-                        kPublisherModuleStart.WithLabels(
-                            identity.DeviceId ?? "", identity.ModuleId ?? "").Set(0);
                         await workerSupervisor.StopAsync();
+                        if (hostScope.Resolve<ISessionManager>() is IDisposable sessionManager){
```
marcschier, line 125: Container will call dispose on both.  If the intend is to call dispose earlier, then it will be called twice which is not good.   I would remove this or we should fix the lifetime scoping.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -188,21 +192,24 @@ public class DefaultSessionManager : ISessionManager, IDisposable {
...
                 _logger.Error(ex, "Unexpected exception stopping processor thread");
             }
-
-            foreach (var session in _sessions.ToList()) {
-                if (!session.Value.Processing.IsCompleted) {
-                    await session.Value.Processing.ConfigureAwait(false);
-                }
-                await HandleDisconnectAsync(session.Key, session.Value).ConfigureAwait(false);
+            finally {
+                Task.WaitAll(processingTasks.ToArray());
```
marcschier, line 212: await WhenAll()

cristipogacean: done

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/ModuleProcess.cs
cristipogacean: Dispose is not being called by the container. Both are singletons and this is intended here. The only way I could clean them up them was by manually call Dispose.

cristipogacean: I moved the cleanup in a StopAsync() method. The Dispose issue with singletons still remains

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
```diff
@@ -126,18 +119,37 @@ public class DataFlowProcessingEngine : IProcessingEngine, IDisposable {
                     });
 
                 _sinkBlock = new ActionBlock<NetworkMessageModel[]>(
-                    async input => await _messageSink.SendAsync(input).ConfigureAwait(false),
+                    async input => {
+                        await _messageSink.SendAsync(input).ConfigureAwait(false);
+                        if (_batchTriggerInterval > TimeSpan.Zero) {
+                            _batchTriggerIntervalTimer.Change(_batchTriggerInterval, _batchTriggerInterval);
```
marcschier, line 125: Shouldn't this be done upon sending?   And why change it to always trigger, isn't it better to only trigger once?   Then reset it when it was send to retrigger?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
```diff
@@ -126,18 +119,37 @@ public class DataFlowProcessingEngine : IProcessingEngine, IDisposable {
...
+                        }
+                    },
                     new ExecutionDataflowBlockOptions {
-                        CancellationToken = cancellationToken
+                        CancellationToken = cancellationToken,
```
marcschier, line 129: Remove comma

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
```diff
@@ -224,6 +236,7 @@ public class DataFlowProcessingEngine : IProcessingEngine, IDisposable {
         /// </summary>
         /// <param name="state"></param>
         private void BatchTriggerIntervalTimer_Elapsed(object state) {
+            _batchTriggerIntervalTimer.Change(Timeout.Infinite, Timeout.Infinite);
```
marcschier, line 239: Should not be needed if you only trigger once

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
```diff
@@ -50,20 +50,24 @@ public class DataFlowProcessingEngine : IProcessingEngine, IDisposable {
...
             if (_config.BatchSize.HasValue && _config.BatchSize.Value > 1) {
                 _dataSetMessageBufferSize = _config.BatchSize.Value;
             }
             if (_config.MaxMessageSize.HasValue && _config.MaxMessageSize.Value > 0) {
                 _maxEncodedMessageSize = _config.MaxMessageSize.Value;
             }
+            
+            _diagnosticInterval = _config.DiagnosticsInterval.GetValueOrDefault(TimeSpan.MaxValue);
+            _batchTriggerInterval = _config.BatchTriggerInterval.GetValueOrDefault(TimeSpan.MaxValue);
+            _diagnosticsOutputTimer = new Timer(DiagnosticsOutputTimer_Elapsed, null,
```
marcschier, line 62: I think there is a constructor that just takes the callback and nothing else.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
```diff
@@ -126,18 +119,37 @@ public class DataFlowProcessingEngine : IProcessingEngine, IDisposable {
...
+
                 await _messageTrigger.RunAsync(cancellationToken).ConfigureAwait(false);
+
             }
             finally {
                 IsRunning = false;
+                _messageTrigger.OnMessage -= MessageTriggerMessageReceived;
+                _diagnosticsOutputTimer.Change(Timeout.Infinite, Timeout.Infinite);
+                _batchTriggerIntervalTimer.Change(Timeout.Infinite, Timeout.Infinite);
+                await Task.WhenAll(
```
marcschier, line 148: Isnt it enough to wait on the sink block?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
cristipogacean: thanks!

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
cristipogacean: right ...

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
cristipogacean: most probably probably it is

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
cristipogacean: send triggering is performed in case of "queue full" or "timer expired". I  wanted to avoid time triggering in case the queue is (quasi) empty right after a "queue full" trigger. The idea is to send out at least one message per batch trigger interval, but to keep them as little and full as possible. I'm stopping the timer before triggering since I don't know how long sending will take and I want to avoid double time triggering.

cristipogacean: done - thanks!

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
luiscantero: I would argue that using a trailing comma is considered best practice. If we decide to add a new item, adding a comma and the new item in another line will result in two lines having changed.

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
```diff
@@ -218,7 +221,7 @@ public class Worker : IWorker, IDisposable {
                     ct.ThrowIfCancellationRequested();
                     using (_jobProcess = new JobProcess(this, jobProcessInstruction,
                         _lifetimeScope, _logger)) {
-                        await _jobProcess.WaitAsync(); // Does not throw
+                        await _jobProcess.WaitAsync(ct); // Does not throw
```
luiscantero, line 224: When adding new or changing awaitable calls, we could always insert ConfigureAwait(false). At least until we get a global solution :)

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
```diff
@@ -43,14 +43,16 @@ public class WorkerSupervisor : IWorkerSupervisor, IDisposable {
 
         /// <inheritdoc/>
         public async Task StopAsync() {
+            _logger.Information("Stopping worker supervisor");
             var stopTasks = new List<Task>();
             _ensureWorkerRunningTimer.Stop();
 
-            foreach (var instance in _instances) {
+            foreach (var instance in _instances.ToList()) {
```
luiscantero, line 50: Why?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/MonitoredItemMessageEncoder.cs
```diff
@@ -312,8 +316,17 @@ public class MonitoredItemMessageEncoder : IMessageEncoder {
...
         /// <param name="encoding"></param>
+        /// <param name="context"></param>
         private IEnumerable<MonitoredItemMessage> GetMonitoredItemMessages(
-            IEnumerable<DataSetMessageModel> messages, MessageEncoding encoding) {
+            IEnumerable<DataSetMessageModel> messages, MessageEncoding encoding,
+            ServiceMessageContext context) {
+            if (context?.NamespaceUris == null) {
+                // declare all notifications in messages dropped 
+                foreach (var message in messages) {
+                    NotificationsDroppedCount += (uint)message?.Notifications?.Count();
```
luiscantero, line 326: If message or Notifications is null, it will result in null and that cannot be added to the count. Should be:
NotificationsDroppedCount += (uint)message?.Notifications?.Count() ?? 0;

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
```diff
@@ -312,10 +317,18 @@ public class NetworkMessageEncoder : IMessageEncoder {
...
         /// <returns></returns>
         private IEnumerable<NetworkMessage> GetNetworkMessages(
-            IEnumerable<DataSetMessageModel> messages,
-            MessageEncoding encoding) {
+            IEnumerable<DataSetMessageModel> messages, MessageEncoding encoding,
+            ServiceMessageContext context) {
+            if (context?.NamespaceUris == null) {
+                // declare all notifications in messages dropped 
+                foreach (var message in messages) {
+                    NotificationsDroppedCount += (uint)message?.Notifications?.Count();
```
luiscantero, line 328: Add ?? 0

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -187,22 +191,25 @@ public class DefaultSessionManager : ISessionManager, IDisposable {
...
         /// </summary>
         /// <returns></returns>
-        private async Task StopAsync() {
-
-            Try.Op(() => _cts?.Cancel());
+        public async Task StopAsync() {
+            var processingTasks = new List<Task>();
             try {
+                _logger.Information("Stopping all sessions");
+                foreach (var session in _sessions.ToList()) {
```
luiscantero, line 198: ToList() adds execution time, is it really necessary?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -187,22 +191,25 @@ public class DefaultSessionManager : ISessionManager, IDisposable {
...
                 _logger.Error(ex, "Unexpected exception stopping processor thread");
             }
-
-            foreach (var session in _sessions.ToList()) {
-                if (!session.Value.Processing.IsCompleted) {
-                    await session.Value.Processing.ConfigureAwait(false);
-                }
-                await HandleDisconnectAsync(session.Key, session.Value).ConfigureAwait(false);
+            finally {
+                await Task.WhenAll(processingTasks.ToArray());
```
luiscantero, line 212: Why an array?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -498,9 +505,9 @@ public class DefaultSessionManager : ISessionManager, IDisposable {
         /// </summary>
         /// <param name="id"></param>
         /// <param name="wrapper"></param>
-        /// <returns>continue processing</returns>
+        /// <returns></returns>
         private async Task HandleDisconnectAsync(ConnectionIdentifier id, SessionWrapper wrapper) {
-            _logger.Debug("Removing idle Session '{id}'", id);
+            _logger.Debug("Removing Session '{id}'", id);
```
luiscantero, line 510: Better "session" to be consistent?

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
cristipogacean: _instances are changing during sttoping

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/MonitoredItemMessageEncoder.cs
cristipogacean: thanks!

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
cristipogacean: thanks!

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
cristipogacean: thanks!

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
cristipogacean: yes, the _session is going to change during foreach

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
cristipogacean: good point

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
cristipogacean: done

PR #545, opened 2020-06-08
Adding site to the discoverer page

MaidaB:
Site is added to the discovery page for the consistency (Site is shown on the Asset page as well). In the future we should think about changing this term because it's being confused with SiteId which is set by users.

### DIFF
--- a/api/src/Microsoft.Azure.IIoT.Api/src/Registry/Extensions/ModelExtensions.cs
+++ b/api/src/Microsoft.Azure.IIoT.Api/src/Registry/Extensions/ModelExtensions.cs
@@ -410,6 +410,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Api.Registry.Models {
             return new DiscovererApiModel {
                 Id = model.Id,
                 SiteId = model.SiteId,
+                SiteOrGatewayId = model.SiteOrGatewayId,
                 LogLevel = (TraceLogLevel?)model.LogLevel,
                 RequestedMode = (DiscoveryMode?)model.RequestedMode,
                 RequestedConfig = model.RequestedConfig.ToApiModel(),
--- a/api/src/Microsoft.Azure.IIoT.OpcUa.Api.Registry/src/Models/DiscovererApiModel.cs
+++ b/api/src/Microsoft.Azure.IIoT.OpcUa.Api.Registry/src/Models/DiscovererApiModel.cs
@@ -83,5 +83,12 @@ namespace Microsoft.Azure.IIoT.OpcUa.Api.Registry.Models {
         [DataMember(Name = "version", Order = 9,
             EmitDefaultValue = false)]
         public string Version { get; set; }
+
+        /// <summary>
+        /// Site/gateway of the discoverer
+        /// </summary>
+        [DataMember(Name = "siteOrGatewayId", Order = 10,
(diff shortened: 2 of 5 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Registry/Models/DiscovererModel.cs
```diff
@@ -20,6 +20,12 @@ public class DiscovererModel {
         /// </summary>
         public string SiteId { get; set; }
 
+        /// <summary>
+        /// Gateway of the discoverer
+        /// TODO: Stop using term Site for the gateway id
```
marcschier, line 25: Call it GatewayId please.

api/src/Microsoft.Azure.IIoT.Api/src/Registry/Extensions/ModelExtensions.cs
```diff
@@ -410,6 +410,7 @@ public static class ModelExtensions {
             return new DiscovererApiModel {
                 Id = model.Id,
                 SiteId = model.SiteId,
+                SiteOrGatewayId = model.SiteOrGatewayId,
```
marcschier, line 413: Please change the SiteOrGatewayId to "GatewayId" in the service model.   Update the ToServiceModel() extension method also, to set Gateway id from api model (for testing and adapter support).   

Then make sure that the GatewayId is set as part of the "registration" extension (trace where SiteOrGatewayId is set).   instead of reading it from the registration, use the "deviceid" property, not "SiteOrGatewayId" property.   The latter is an "or", but the gatewayId is always the "deviceid".

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Registry/src/Models/DiscovererApiModel.cs
```diff
@@ -83,5 +83,12 @@ public class DiscovererApiModel {
         [DataMember(Name = "version", Order = 9,
             EmitDefaultValue = false)]
         public string Version { get; set; }
+
+        /// <summary>
+        /// Site/gateway of the discoverer
+        /// </summary>
+        [DataMember(Name = "siteOrGatewayId", Order = 10,
+            EmitDefaultValue = false)]
+        public string SiteOrGatewayId { get; set; }
```
marcschier, line 92: Rename this to GatewayId please.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Registry/Models/DiscovererModel.cs
marcschier, line 25: Remove the todo after changing it to GatewayId.

samples/src/Microsoft.Azure.IIoT.App/src/Pages/Discoverers.razor
```diff
@@ -49,6 +49,7 @@
     <thead>
         <tr>
             <th class="width-large">Discovery Module Id</th>
+            <th class="width-medium">Site</th>
```
marcschier, line 52: This should be called Gateway Id.   But if the idea is to show the site, then below should refer to the SiteId value in the API model.

PR #558, opened 2020-06-15
Using 3.1.300 version of .Net Core SDK for build and test.

karok2m:


karok2m:
Closing the PR as .Net Core SDK version was not the root cause of mac test failures.

PR #569, opened 2020-06-23
Fixes/events test

cristipogacean:


services/src/Microsoft.Azure.IIoT.Services.OpcUa.Events/tests/Api/RegistryServiceEventsTests.cs
```diff
@@ -46,7 +46,7 @@ public class RegistryServiceEventsTests {
             })) {
 
                 await bus.PublishAsync(expected);
-                await Task.WhenAny(result.Task, Task.Delay(1000));
+                await Task.WhenAny(result.Task, Task.Delay(5000));
```
luiscantero, line 49: It would be better to create a constant that can be used everywhere and changed once, instead of a magic number

cristipogacean: well, everything is hardcoded in the unit tests... I won't change this

luiscantero: afaik magic numbers are fine in unit tests if they are used to compare to return values. But this are just delays. I would be interested to know what @marcschier's opinion is.

cristipogacean: Sure, he probably wrote the original code. I just changed the 1 to 5 after noticing too often build failures due to awaited operations that take more than 1 second to be executed in these tests. Increasing the timeout helps a lot. I would be interested to learn, **for this particular blocked PR**: what are the improvements in code complexity, extensibility, execution speed and especially maintainability and readability when constants will be used instead of hardcoded values? Please note, for your argumentation, that there are tests using different hardcoded timeout values in the very same files, therefore, in order to be truly compliant, we will need a different contextually meaningful constants for every test case here.

luiscantero: Sorry, didn't mean to block the pr. I thought we could just have a quick discussion, but nobody seems to be available. This is why we should meet at the office every now and then :)

PR #576, opened 2020-06-24
show error message instead of exception in UI

dacolgit:


samples/src/Microsoft.Azure.IIoT.App/src/Services/Browser.cs
```diff
@@ -127,10 +127,9 @@ public class Browser {
                 pageResult.Error = "Unauthorized access: Bad User Access Denied.";
             }
             catch (Exception e) {
-                // skip this node
-                _logger.Error(e, "Can not browse node '{id}'", id);
-                var errorMessage = string.Concat(e.Message, e.InnerException?.Message ?? "--", e?.StackTrace ?? "--");
-                pageResult.Error = errorMessage;
+                var message = $"Can not browse node '{id}'";
```
luiscantero, line 130: I think we should be consistent and use "cannot" instead of "can not" everywhere https://brians.wsu.edu/2016/05/31/cannot/

dacolgit: changed

PR #578, opened 2020-06-25
Simplify razor output

luiscantero:
* Remove some unused assignments
* Reformat some ternary operators for easier interpretation
* Remove redundant placeholders


samples/src/Microsoft.Azure.IIoT.App/src/Pages/PublishedNodes.razor.cs
```diff
@@ -34,7 +34,7 @@ public partial class PublishedNodes {
             new PagedResult<ListNode>();
         private string _tableView = "visible";
         private string _tableEmpty = "displayNone";
-        private IAsyncDisposable _publishEvent { get; set; }
+        private IAsyncDisposable PublishEvent { get; set; }
```
marcschier, line 37: Why even use a property, why not just a variable?

luiscantero: Agree. There are a number of properties that should be changed to variables. My suggestion would be to only use properties for data-binding, as unused properties are not shown in VS. I will create a work item for this.

Issue #551, opened 2020-06-10
Root level deploy.cmd/deploy.sh scripts cannot be invoked with -version


jleger-ptc:
**Describe the bug**
Because the root level deployment scripts act as a pass-through to the `./deploy/scripts/deploy.ps1` file and hard code "**-version preview**", end users are unable to leverage them if/when they would like to specify a specific version of the stack to deploy

deploy.sh

- Line 69: `pwsh -File $curdir/deploy/scripts/deploy.ps1 "$@" -version preview`

deploy.cmd

- Line 33: `%PWSH% -ExecutionPolicy Unrestricted ./deploy.ps1 %* -version preview`

We were hoping to use these top level deployment scripts in CI/CD to stage the Azure Powershell environment, but need to specify a version

**To Reproduce**
1. Go to the root of the workspace
2. Invoke the deploy.sh or deploy.cmd script with the -version flag (e.g. `./deploy.cmd -version 2.7.105`)

**Expected behavior**
Users should be able to provide any/all input parameters that they would by directly invoking the `deploy.ps1` script

Instead of hard coding the `-version preview` at this level, could we specify a default value in `deploy.ps1` similar to `$type` and `$environmentName`?

**Screenshots**
![deploy_version_failure](https://user-images.githubusercontent.com/61740697/84300549-7183fd00-ab20-11ea-8907-ed96110ef9cd.PNG)

monikavar:
Hi @jleger-ptc , thanks for reporting the issue. It has been fixed now.

PR #579, opened 2020-06-25
fix for hard-coded version(preview) in deploy.ps1

monikavar:
This PR solves [this](https://github.com/Azure/Industrial-IoT/issues/551) github issue.

This will check if version is passed in deploy script and use that version, otherwise use preview.


marcschier:
Resolves #551

PR #589, opened 2020-06-30
restored telemetry sample message mode

cristipogacean:
* telemetry messages in sample mode are no longer converted to DataSet messages for TSI compatibility
* sequence number added to the sample messages


services/src/Microsoft.Azure.IIoT.Services.Processor.Onboarding/src/Properties/launchSettings.json
```diff
@@ -0,0 +1,27 @@
+{
```
marcschier, line 1: Please remove

cristipogacean: done, no idea why was this created, by bad that I did not notice when adding it

luiscantero: Sometimes this file is wanted/needed: https://stackoverflow.com/a/60383204/569998

PR #590, opened 2020-07-01
Small cosmetic changes to diag info

luiscantero:
* Move estimated chunks per day right below chunk usage stats
* Small change to indentation
* Add thousand separator to outgress buffer count


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
```diff
@@ -175,28 +175,28 @@ public class DataFlowProcessingEngine : IProcessingEngine, IDisposable {
             _logger.Debug("Identity {deviceId}; {moduleId}", _identity.DeviceId, _identity.ModuleId);
 
             var diagInfo = new StringBuilder();
-            diagInfo.Append("\n   DIAGNOSTICS INFORMATION for         : {host}\n");
```
marcschier, line 178: Could we use AppendLine, rather than Append with \n`?

luiscantero: Done

PR #591, opened 2020-07-01
fix discovery Url scan

dacolgit:


samples/src/Microsoft.Azure.IIoT.App/src/Models/DiscovererInfoRequested.cs
```diff
@@ -66,6 +66,15 @@ public void RemoveDiscoveryUrl(string url)
             RequestedDiscoveryUrls.Remove(url);
         }
 
+        /// <summary>
+        /// Clear url list
+        /// </summary>
+        public void ClearDiscoveryUrlList(List<string> list) {
+            if (list != null) {
+                list.Clear();
+            }
```
luiscantero, line 75: list?.Clear();

samples/src/Microsoft.Azure.IIoT.App/src/Validation/DiscovererInfoValidator.cs
```diff
@@ -86,5 +91,13 @@ public class DiscovererInfoValidator : AbstractValidator<DiscovererInfoRequested
 
             return false;
         }
+
+        private bool BeAValidDiscoveryUrl(List<string> value) {
+            if (value != null) {
+                return !(value.Contains(null) || value.Contains(string.Empty));
+            }
+
+            return true;
```
luiscantero, line 100: When value == null, it will return true. Is this expected?

common/src/Microsoft.Azure.IIoT.Serializers.NewtonSoft/src/Default/NewtonSoftJsonSerializer.cs
```diff
@@ -231,8 +231,17 @@ internal class JsonVariantValue : VariantValue {
...
             public override object ConvertTo(Type type) {
                 try {
-                    return Token.ToObject(type,
-                        JsonSerializer.CreateDefault(_serializer.Settings));
+                        if (type != typeof(List<string>)) {
+                            return Token.ToObject(type,
+                                JsonSerializer.CreateDefault(_serializer.Settings));
+                        }
+                        else {
+                            List<string> output = new List<string>();
```
luiscantero, line 239: C# guidelines recommend using "var" when type is apparent (e.g. when using "new"). It would be great to reach agreement on this so that we can all be consistent :) https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions#implicitly-typed-local-variables

marcschier: var please!

samples/src/Microsoft.Azure.IIoT.App/src/Validation/DiscovererInfoValidator.cs
dacolgit: null means empty list, no discovery Urls set

common/src/Microsoft.Azure.IIoT.Serializers.NewtonSoft/src/Default/NewtonSoftJsonSerializer.cs
dacolgit: changed

samples/src/Microsoft.Azure.IIoT.App/src/Models/DiscovererInfoRequested.cs
dacolgit: changed

common/src/Microsoft.Azure.IIoT.Serializers.NewtonSoft/src/Default/NewtonSoftJsonSerializer.cs
```diff
@@ -231,8 +231,17 @@ internal class JsonVariantValue : VariantValue {
...
+                            foreach(var item in Token) {
```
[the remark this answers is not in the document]
dacolgit: In the original code this part was never hit due to the missing DiscoveryUrl property in DiscoverySettingsController.cs.

With the other changes this part is now executed when setting a discoveryUrls, but the original code raises an exception:
  Cannot deserialize the current JSON object (e.g. {"name":"value"}) into type ' 
  System.Collections.Generic.List`1[System.String]' because the type requires a JSON array (e.g. [1,2,3]) to deserialize 
  correctly.

The reason is that Token content is something like that: {{"0": "opc.tcp://dacoldell5520:51510/UA/DemoServer"}} and doesn't correspond to type.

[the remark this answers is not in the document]
marcschier: The fix must then be different, because IoT Hub twin does not support array we encode all arrays in a dictionary with index as key.   To make this work make DiscoveryUrl setting on controller a Dictionary<string, string>.  Search and you will find other settings that are doing the same.   There is a helper on the dictionary called something like DecodeAsList(), use that to set the property on the discovery service.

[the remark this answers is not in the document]
dacolgit: changed DiscoveryUlrs to Dictionary

common/src/Microsoft.Azure.IIoT.Serializers.NewtonSoft/src/Default/NewtonSoftJsonSerializer.cs
```diff
@@ -232,7 +232,7 @@ internal class JsonVariantValue : VariantValue {
             public override object ConvertTo(Type type) {
                 try {
                     return Token.ToObject(type,
-                        JsonSerializer.CreateDefault(_serializer.Settings));
+                        JsonSerializer.CreateDefault(_serializer.Settings));
```
marcschier, line 235: Undo this file - (or remove added whitespace)

modules/src/Microsoft.Azure.IIoT.Modules.Discovery/src/Controllers/DiscoverySettingsController.cs
```diff
@@ -8,8 +8,10 @@ namespace Microsoft.Azure.IIoT.Modules.Discovery.Controllers {
     using Microsoft.Azure.IIoT.OpcUa.Edge;
     using Microsoft.Azure.IIoT.OpcUa.Registry.Models;
     using System;
+    using System.Collections.Generic;
     using System.Threading.Tasks;
-
+    using Microsoft.Azure.IIoT.Hub;
```
marcschier, line 13: Nit: Move new using up above System;

modules/src/Microsoft.Azure.IIoT.Modules.Discovery/src/Controllers/DiscoverySettingsController.cs
```diff
@@ -98,6 +100,14 @@ public class DiscoverySettingsController : ISettingsController {
             get => _discovery.Configuration.IdleTimeBetweenScans;
         }
 
+        /// <summary>
+        /// Discovery Urls to scan
+        /// </summary>
+        public Dictionary<string, string> DiscoveryUrls {
+            set => _discoveryUrls = value;
```
marcschier, line 107: Consider doing it like above - _config.DiscoveryUrls = value.DecodeAsList(), get=> _config.DiscoveryUrls.EncodeAsList().   Then remove the newly added member and logic below.

modules/src/Microsoft.Azure.IIoT.Modules.Discovery/src/Controllers/DiscoverySettingsController.cs
dacolgit: done

common/src/Microsoft.Azure.IIoT.Serializers.NewtonSoft/src/Default/NewtonSoftJsonSerializer.cs
dacolgit: done

modules/src/Microsoft.Azure.IIoT.Modules.Discovery/src/Controllers/DiscoverySettingsController.cs
dacolgit: changed.
I did that way because I used as a sample DiscoverySettingController.cs

PR #597, opened 2020-07-02
exposing prometheus endpoints for non-api services

monikavar:
- Exposing Prometheus endpoints for non-api services:
    - Microsoft.Azure.IIoT.Services.Processor.Events
    - Microsoft.Azure.IIoT.Services.Processor.Onboarding
    - Microsoft.Azure.IIoT.Services.Processor.Telemetry.Cdm
    - Microsoft.Azure.IIoT.Services.Processor.Telemetry
    - Microsoft.Azure.IIoT.Services.Processor.Tunnel
    - Microsoft.Azure.IIoT.Services.OpcUa.Registry.Sync

- Updating helm charts to add pod annotations for the same
- Added events processor metrics


services/src/Microsoft.Azure.IIoT.Services.OpcUa.Registry.Sync/src/HostStarterService.cs
```diff
@@ -69,6 +72,7 @@ ILogger logger
             ServiceInfo = serviceInfo ?? throw new ArgumentNullException(nameof(serviceInfo));
             _hostProcesses = hostProcesses?.ToList() ?? throw new ArgumentNullException(nameof(hostProcesses));
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _metricServer = new KestrelMetricServer(port: kSyncPrometheusPort);
```
karok2m, line 75: Please wrap creation, start and stop of `KestrelMetricServer` in an `IHostProcess` interface similar to other processes that we have. That way you would only need to register it with `ContainerBuilder` in `Program.cs` and the same code would be used for all processors. Port can be passes as a configuration.

common/src/Microsoft.Azure.IIoT.Hub.Processor/src/Services/EventProcessorFactory.cs
```diff
@@ -299,6 +306,11 @@ private class EventProperties : IDictionary<string, string> {
             private readonly long? _interval;
             private readonly Stopwatch _sw;
             private readonly PartitionContext _partitionContext;
+            private static readonly Gauge kEventProcessorDetails = Metrics
+                .CreateGauge("iiot_event_processor_info", "details about event processor",
+                    new GaugeConfiguration {
+                        LabelNames = new[] { "id", "partition", "status" }
```
karok2m, line 312: Can you please use `partitionId` instead of `partition` to align this metric key with diagnostics keys that are delivered with logs.

Also please change logs on line `159` and `165` to use `{partitionId}` instead of `{partition}`. This one is my mistake.

monikavar: done

common/src/Microsoft.Azure.IIoT.Core/src/Diagnostics/Default/MetricServerHost.cs
```diff
@@ -0,0 +1,46 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+namespace Microsoft.Azure.IIoT.Messaging.Default {
```
marcschier, line 6: Namespace name.  And empty line after.   

However, as below, please move this to AspNetCore project which is already being referenced by the processors.

common/src/Microsoft.Azure.IIoT.Core/src/Diagnostics/Default/MetricServerHost.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// Start and stop metric server
+    /// </summary>
+    public class MetricServerHost : IHostProcess {
+
+        /// <summary>
+        /// Auto registers metric server
+        /// </summary>
+        /// <param name="port"></param>
+        /// <param name="logger"></param>
+        public MetricServerHost(int? port, ILogger logger) {
```
marcschier, line 23: Ideally has a configuration interface and object that can be injected.

common/src/Microsoft.Azure.IIoT.Core/src/Microsoft.Azure.IIoT.Core.csproj
```diff
@@ -10,6 +10,7 @@
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
     <PackageReference Include="prometheus-net" Version="3.6.0" />
+    <PackageReference Include="prometheus-net.AspNetCore" Version="3.6.0" />
```
marcschier, line 13: Remove and move the code to the AspNetCore package.   It is already referenced and avoids polluting the core package with AspNetCore app model.

common/src/Microsoft.Azure.IIoT.Hub.Processor/src/Services/EventProcessorFactory.cs
```diff
@@ -17,6 +17,7 @@ namespace Microsoft.Azure.IIoT.Hub.Processor.Services {
     using System.Diagnostics;
     using Autofac;
     using Microsoft.Azure.IIoT.Exceptions;
```
marcschier, line 19: Please move the Exceptions using over using Serilog while you are touching the file.

services/src/Microsoft.Azure.IIoT.Services.Processor.Events/src/Program.cs
```diff
@@ -128,6 +128,11 @@ IConfiguration configuration
             builder.RegisterType<EventProcessorFactory>()
                 .AsImplementedInterfaces();
 
+            // Prometheus metric server
+            builder.RegisterType<MetricServerHost>()
+                .AsImplementedInterfaces().SingleInstance()
+                .WithParameter("port", 9500);
```
marcschier, line 134: This is brittle, the parameter name must now never change in the "implementation".   A new implementation must use exactly the same name for the parameter, etc.   Please add interface or use an existing interface to define the metricsserver port on the Config.cs class which is already injected.    See below for the other occurrences.

services/src/Microsoft.Azure.IIoT.Services.Processor.Telemetry/src/Program.cs
```diff
@@ -24,6 +24,7 @@ namespace Microsoft.Azure.IIoT.Services.Processor.Telemetry {
     using Microsoft.Extensions.DependencyInjection;
     using Autofac.Extensions.DependencyInjection;
     using Microsoft.Extensions.Hosting;
+    using Microsoft.Azure.IIoT.Messaging.Default;
```
marcschier, line 27: Move using up to be aligned with our other IIoT usings.

PR #601, opened 2020-07-06
added camel case annotations for json encoding

cristipogacean:


luiscantero:
Newtonsoft has a global setting CamelCasePropertyNamesContractResolver and JsonSerializer has JsonNamingPolicy.CamelCase. The above annotations are usually unnecessary unless you want granular control as to what should be serialized using camel case. Just saying ;)

PR #609, opened 2020-07-09
fix IotHubQuotaExceeded exception

dacolgit:


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Services/DiscovererRegistry.cs
```diff
@@ -221,14 +221,25 @@ public sealed class DiscovererRegistry : IDiscovererRegistry {
...
+                return new DiscovererListModel {
+                    ContinuationToken = queryResult.ContinuationToken,
+                    Items = queryResult.Items
                     .Select(t => t.ToDiscovererRegistration())
                     .Select(s => s.ToServiceModel())
                     .ToList()
-            };
+                };
+            }
+            catch (Exception e) {
```
marcschier, line 234: This is not a good fix - let's fix the exception status serialization problem instead.

dacolgit: I'm not sure it's a serialization issue. I put breakpoints in ExceptionConverter.cs and in MessagePackSerializer.cs in ExceptionResolver and were never hit.
The error is returned in private async Task<IHttpResponse> SendAsync(IHttpRequest httpRequest, HttpMethod httpMethod, CancellationToken ct) in HttpClient.cs:
image

The client.SendAsync return a right response with StatusCode='InternalServerError' and Content empty. 
Content field is what then cause an exception message empty. I was not able to find any serialization issue.

dacolgit: image

the line highlighted in blue is the POST request that causes the error. The answer is empty, and it is what I see in the code

dacolgit: image

this is the same POST message with the right response

PR #614, opened 2020-07-15
IAI: Updated docs to include links to compiled binaries.

karok2m:


docs/deploy/howto-deploy-aks.md
```diff
@@ -29,16 +29,21 @@
...
 Kubernetes Service (AKS) cluster, while `deploy.ps1` runs the entire platform as a web application.
 
-`Microsoft.Azure.IIoT.Deployment` deploys `2.7.170` version of Azure Industrial IoT microservices.
-
 Source codes of `Microsoft.Azure.IIoT.Deployment` can be found in the following directory:
 [deploy/src/Microsoft.Azure.IIoT.Deployment/](../../deploy/src/Microsoft.Azure.IIoT.Deployment/)
 
+Latest compiled binaries of `Microsoft.Azure.IIoT.Deployment` (version `2.8.5`) can be found here:
+
+* [linux-x64](https://azureiiot.blob.core.windows.net/binaries/master/2.8.5/linux-x64/Microsoft.Azure.IIoT.Deployment)
```
marcschier, line 43: We should add a aka.ms link to point to the binaries.  Could even update the aka.ms redirector link whenever we change location, so the doc stays the same.

karok2m: OK. Let me add a separate task for that in our backlog.

PR #615, opened 2020-07-15
Disable batching in publisher regardless of job configuration

marcschier:


karok2m:
Not sure if it is relevant or not given that this change will not go to `master`, but we do already have `2.7.171` version of containers and binaries. That was commit that bumped components version to `2.7.170` in the Helm chart. Rebasing onto that commit will effectively make this `2.7.172` version.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Extensions/WriterGroupModelEx.cs
```diff
@@ -20,8 +20,8 @@ public static class WriterGroupModelEx {
         public static IWriterGroupConfig ToWriterGroupJobConfiguration(
             this WriterGroupJobModel model, string publisherId) {
             return new WriterGroupJobConfig {
-                BatchSize = model.Engine?.BatchSize,
-                BatchTriggerInterval = model.Engine?.BatchTriggerInterval,
+                BatchSize = 1, // 2.7 disabled: model.Engine?.BatchSize,
```
cristipogacean, line 23: I would keep these values so we have a chance to enable batching manually from the file when needed. Defaulting EngineConfigurationModel to batch size to 1 respectively interval to null in enough to change the default behavior.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Services/PublisherJobService.cs
```diff
@@ -229,6 +229,13 @@ public sealed class PublisherJobService : IPublishServices<string> {
                 var publishJob = (WriterGroupJobModel)_serializer.DeserializeJobConfiguration(
                     job.JobConfiguration, job.JobConfigurationType);
                 if (publishJob != null) {
+
+                    if (publishJob.Engine != null) {
+                        publishJob.Engine.BatchTriggerInterval = null;
```
cristipogacean, line 234: why is this needed?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Extensions/WriterGroupModelEx.cs
marcschier: I thought about it - however, the problem is existing jobs will have 50 and 10 seconds configured, so they will not get fixed.   This is a bit of a drastic step, but it is simpler than rewriting the jobs database when this is the end of 2.7 anyway, and most users will use 2.7.170.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Services/PublisherJobService.cs
marcschier: The goal is that any update to existing jobs will write the job back to the database with the trigger empty and batchsize 1, migrating existing jobs as they are upgraded.   It is layer 2 of the 3 layer safety net:  1. create new jobs without batching, 2. update existing jobs as they are updated, 3. disregard any batch size or interval in publisher 2.7.171.   The 3rd step is needed if 1 or 2 are not happening.

marcschier:
release/2.7.170 is the target for this change.  2.7.171 tag would be overridden with the build from this branch once checked in.   Alternative is pull a dedicated release branch from master @ the commit of the helm change and get a 2.7.172.   I prefer the overriding of 2.7.171 container at this point as that image has no meaning in our release history anyway.

PR #613, opened 2020-07-13
IAI: Upgraded IAI to deploy 2.7.170 version of Azure IIoT components.

karok2m:
Changes:

IAI: Using `azure-industrial-iot` Helm chart to deploy components.
IAI: Added two more IoT Hub Event Hub consumer groups.
IAI: Using connection string with data lake specific endpoint suffix.
IAI: Added environment variables for Log Analytics Workspace.
IAI: Updated service URLs for `2.7`.
IAI: Added `PCS_SIGNALR_MODE`.
IAI: Upgraded Helm chart to version `0.3.1` and components to `2.7.170`.
IAI: Removed IIoTK8SClient.
IAI: Removed Kubernetes resource definition YAML files.
IAI: Added configuration for Helm chart.
IAI: Upgraded Kubernetes version in AKS to `1.16.10`.
IAI: Minor renaming of ApplicationURL configuration parameter.
IAI: Updated documentation.

PR #619, opened 2020-07-16
Bugfix: Reverting removal of PCS_STORAGE_KEY as it used later for TSI.

karok2m:
`PCS_STORAGE_KEY ` was removed (#613) as it was seemingly not used in code or deployment scripts. What I previously missed was that while `PCS_STORAGE_KEY ` is not used, its representation in Azure Key Vault is, `pcs-storage-key` secret.


karok2m:
A deployment with this fix finished successfully.

PR #621, opened 2020-07-17
Fix broken links due to backslash

luiscantero:
* Fix getting started doc links

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -38,7 +38,7 @@ On Linux:
 ./deploy.sh
 ```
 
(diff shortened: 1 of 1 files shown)


luiscantero:
> BTW, I found this VS Code extension to be useful for checking links in markdown files: [HTTP/s and relative link checker](https://marketplace.visualstudio.com/items?itemName=blackmist.LinkCheckMD)

Unfortunately it doesn't detect the links with backslash as errors, as they are not broken on Windows :/

PR #625, opened 2020-07-20
fix Discovery with Url 

dacolgit:


api/src/Microsoft.Azure.IIoT.OpcUa.Api.Registry/src/Models/DiscoveryMode.cs
```diff
@@ -18,6 +18,12 @@ public enum DiscoveryMode {
         [EnumMember]
         Off,
 
+        /// <summary>
```
marcschier, line 21: Please add to the end of the enum, or this will break backcompat since many enums are specified as int and this shifts the following enum values.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Registry/Models/DiscoveryMode.cs
```diff
@@ -15,6 +15,11 @@ public enum DiscoveryMode {
         /// </summary>
         Off,
 
+        /// <summary>
```
marcschier, line 18: For consistency, please also add to end of the enum.   Even though technically this is only required for API models

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Registry/src/Models/DiscoveryMode.cs
dacolgit: added

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Registry/Models/DiscoveryMode.cs
dacolgit: done

PR #649, opened 2020-07-30
Allow environment variable configuration of batch size and interval.

marcschier:
The following environment variables can be set to configure the batch trigger interval and batch size.
The environment variable PCS_DEFAULT_PUBLISH_JOB_BATCH_INTERVAL can be set as milliseconds from 100 to 3,600,000 ms.  Other values (e.g. <= 0) will result in the default value of 10,000 ms (10 seconds).
The environment variable PCS_DEFAULT_PUBLISH_JOB_BATCH_SIZE can be set as a value between 2 and 1000.   Every other value will result in the default of 50 subscription notifications per batch being set.
The environment variables must be set on the Publisher container Pod to affect the settings in all job configurations from this point on that are performed using the Publisher REST API.  (iot/opc-publisher-service).
The change is in the PublisherJobService Adapter, which adapts from 2.7 REST API to internal Publisher configuration API.  The engine configuration that had hardcoded settings in them in 2.7 release will be updated to read both values out of the environment and set them in the configuration.


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
```diff
@@ -163,124 +140,94 @@ public class DataFlowProcessingEngine : IProcessingEngine, IDisposable {
...
-            diagInfo.AppendLine("  # Encoder avg IoT Chunk (4 KB) usage : {chunkSizeAverage,14:0.#}");
-            diagInfo.AppendLine("  # Estimated IoT Chunks (4 KB) per day: {estimatedMsgChunksPerDay,14:n0}");
-            diagInfo.AppendLine("  # Outgress Batch Block buffer size   : {batchNetworkMessageBlockOutputCount,14:0}");
-            diagInfo.AppendLine("  # Outgress input buffer count        : {sinkBlockInputCount,14:n0}");
-
-            string sentMessagesPerSecFormatted = _messageSink.SentMessagesCount > 0 && totalDuration > 0 ? $"({sentMessagesPerSec:0.##}/s)" : "";
-            diagInfo.AppendLine("  # Outgress IoT message count         : {messageSinkSentMessagesCount,14:n0} {sentMessagesPerSecFormatted}");
-            diagInfo.AppendLine("  # Connection retries                 : {connectionRetries,14:0}");
+            diagInfo.AppendLine("  # Encoder avg IoT Unit (4 KB) usage  : {chunkSizeAverage,14:0.#}");
+            diagInfo.AppendLine("  # Estimated IoT Units (4 KB) per day : {estimatedMsgChunksPerDay,14:n0}");
```
luiscantero, line 176: The documentation mentions "chunks" and also "meter size" for this, not units. Why introduce a new term? (I think it might be confusing)

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/MonitoredItemMessageEncoder.cs
```diff
@@ -76,77 +68,85 @@ public class MonitoredItemMessageEncoder : IMessageEncoder {
...
+            if (values > 0) {
+                // Write remaining values
+                yield return CreateNetworkMessage(values, content);
+            }
+
+            // Helper to write out content of the writer
+            NetworkMessageModel CreateNetworkMessage(int values, StringBuilder content) {
+                // Write out existing buffer
+                var body = content.Replace(',', ']', content.Length - 1, 1).ToString();
+              //  System.Diagnostics.Debug.Assert(Newtonsoft.Json.JsonConvert.DeserializeObject(body) != null);
```
luiscantero, line 135: Test code?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
marcschier: Ah - got it - will change it back.

PR #650, opened 2020-08-03
Allow environment variable configuration of batch size and interval. 

mregen:
The following environment variables can be set to configure the batch trigger interval and batch size.
The environment variable PCS_DEFAULT_PUBLISH_JOB_BATCH_INTERVAL can be set as milliseconds from 100 to 3,600,000 ms.  Other values (e.g. <= 0) will result in the default value of 10,000 ms (10 seconds).
The environment variable PCS_DEFAULT_PUBLISH_JOB_BATCH_SIZE can be set as a value between 2 and 1000.   Every other value will result in the default of 50 subscription notifications per batch being set.
The environment variables must be set on the Publisher container Pod to affect the settings in all job configurations from this point on that are performed using the Publisher REST API.  (iot/opc-publisher-service).
The change is in the PublisherJobService Adapter, which adapts from 2.7 REST API to internal Publisher configuration API.  The engine configuration that had hardcoded settings in them in 2.7 release will be updated to read both values out of the environment and set them in the configuration.

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
   "version": "2.7",
-  "versionHeightOffset": -1,
+  "versionHeightOffset": 180,
   "publicReleaseRefSpec": [
     "^refs/heads/master$",
     "^refs/heads/release/\\d+\\.\\d+"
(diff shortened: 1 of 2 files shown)


version.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
   "version": "2.7",
-  "versionHeightOffset": -1,
+  "versionHeightOffset": 180,
```
karok2m, line 4: @Martin, why isn't the branch based on `release/2.7.171` instead of `2.7.170` ? Hard-coding of `180` value seem to be a necessary workaround because of that.

mregen: we just focus on the least amount of necessary changes for this hotfix, just enable the batching per environment

PR #652, opened 2020-08-03
Drop iot hub messages if exceed the maximum limit

koepalex:
* limit for outgoing ioT Hub messages set to 200 (hardcoded) which equals _~51 MB memory_
* extend diagnostics output with number of dropped messages

New diagnostic output:
image

/cc @luiscantero


koepalex:
I've reverted the wrongly merge from master into release/2.7.180, now even if more commits are listed, only one file has changed and only contains the expected changes.

luiscantero:
> Can you also make the value configurable ? 50 MBs is not a particularly big number, and I can see people wanting to change that according to hardware they have.

Regarding this comment, there are indeed two scenarios that we need to cover:
1) Too much work for the Publisher: From what I've seen, if the Publisher is healthy, the queue may increase by up to 10 or so and then decrease again. If the queue increases steadily, there is no point in having a larger queue as it will never be enough.
2) When the Publisher is healthy and connectivity to the hub is lost temporarily, the queue may increase to a very large size, but when connection is restored, the queue will be emptied correctly.

To cover both scenarios, we indeed need a larger queue size. Instead of making it configurable, how about setting it to half of the available RAM? (Process.GetCurrentProcess().PrivateMemorySize64)

CC @karok2m @koepalex

karok2m:
@luiscantero, half of available memory does sound better than 50 MB. Though, I would still prefer it to be configurable both from publisher cli arguments as well as configuration for publisher service, similar to batch size and interval.

Having said that, I will also be fine with it being hard-coded, as long as we all agree on that.

koepalex:
> @luiscantero, half of available memory does sound better than 50 MB. Though, I would still prefer it to be configurable both from publisher cli arguments as well as configuration for publisher service, similar to batch size and interval.
> 
> Having said that, I will also be fine with it being hard-coded, as long as we all agree on that.

Provided capability to configure maximum outgress messages via command line or environment variable.

api/src/Microsoft.Azure.IIoT.Api/src/Publisher/Models/EngineConfigurationApiModel.cs
```diff
@@ -42,5 +42,11 @@ public class EngineConfigurationApiModel {
             EmitDefaultValue = false)]
         public int? MaxMessageSize { get; set; }
 
+        /// <summary>
+        /// Max IoT D2C messages
```
karok2m, line 46: Maybe `Max IoT D2C outgress message buffer size` ? I would like to see here that this is buffer size, otherwise as a newcomer this summary and name of the parameter would be confusing for me.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
```diff
@@ -24,7 +24,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine {
     /// Dataflow engine
     /// </summary>
     public class DataFlowProcessingEngine : IProcessingEngine, IDisposable {
-
+
```
karok2m, line 27: Please remove extra spaces.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/IEngineConfiguration.cs
```diff
@@ -31,5 +31,12 @@ public interface IEngineConfiguration {
         /// Diagnostics interval
         /// </summary>
         TimeSpan? DiagnosticsInterval { get; }
+
+        /// <summary>
+        /// Define the maximum number of messages,
```
karok2m, line 36: Please mention that this is the number of messages in outgress **buffer**. Without it the summary is not clear.

BTW, 200 * 256 KB is exactly 50 MB.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Services/PublisherJobService.cs
```diff
@@ -48,6 +48,18 @@ public sealed class PublisherJobService : IPublishServices<string> {
             return 50; // default
         });
 
+        /// <summary>
+        /// Read default max outgress messages from environment
```
karok2m, line 52: Maybe `Read default max outgress message buffer size from environment` ?

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliConfigKeys.cs
```diff
@@ -97,6 +97,11 @@ public static class LegacyCliConfigKeys {
         /// </summary>
         public const string MaxMessageSize = "MaxMessageSize";
 
+        /// <summary>
+        /// Key for the max (IoT Hub D2C) messages
+        /// </summary>
+        public const string MaxOutgressMessages = "maxoutgressmessages";
```
karok2m, line 103: All other keys have camel-case, please change `MaxOutgressMessages` similarly.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliOptions.cs
```diff
@@ -230,6 +232,11 @@ public class LegacyCliOptions : Dictionary<string, string>, IAgentConfigProvider
         /// </summary>
         public int? MaxMessageSize => LegacyCliModel.MaxMessageSize;
 
+        /// <summary>
+        /// The Maximum (IoT D2C) messages
```
karok2m, line 236: Please add in the summary that this is message buffer size.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/EngineConfigurationModel.cs
```diff
@@ -30,5 +30,12 @@ public class EngineConfigurationModel {
         /// Diagnostics setting
         /// </summary>
         public TimeSpan? DiagnosticsInterval { get; set; }
+
+        /// <summary>
+        /// Define the maximum number of messages,
```
karok2m, line 35: Same here, please mention it being maximum number of messages in a buffer.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliOptions.cs
```diff
@@ -144,6 +144,8 @@ public class LegacyCliOptions : Dictionary<string, string>, IAgentConfigProvider
                         (int k) => this[LegacyCliConfigKeys.BatchTriggerInterval] = TimeSpan.FromSeconds(k).ToString() },
                     { "ms|iothubmessagesize=", "The maximum size of the (IoT D2C) message.",
                         (int i) => this[LegacyCliConfigKeys.MaxMessageSize] = i.ToString() },
+                    { "mom|maxoutgressmessages=", "The maximum amount of the (IoT D2C) messages",
```
karok2m, line 147: Maybe just `om` to keep argument naming consistent ?

Also, please mention that this is ougress buffer size. With just `The maximum amount of the (IoT D2C) messages` I am left wondering whether publisher will stop sending messages after this number or what.

api/src/Microsoft.Azure.IIoT.Api/src/Publisher/Models/EngineConfigurationApiModel.cs
koepalex: updated XML comment

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/IEngineConfiguration.cs
koepalex: updated XML comment and fixed the memory site

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Services/PublisherJobService.cs
koepalex: updated XML comment

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliOptions.cs
koepalex: updated XML comment

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/EngineConfigurationModel.cs
koepalex: updated XML comment

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliOptions.cs
koepalex: changed to om and updated description

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliConfigKeys.cs
koepalex: updated to use camel-case

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
koepalex: spaces removed

PR #656, opened 2020-08-05
fix avg notification per message for Pubsub

koepalex:
Updated the calculation of average Notifications per Message for Pub/Sub encoded messages correctly. 
One Pub/Sub message contains array of messages and each of those messages have array of value changes (notifications) as Payload.


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
```diff
@@ -91,6 +91,7 @@ public class NetworkMessageEncoder : IMessageEncoder {
             var messageSize = 2; // array brackets
             maxMessageSize -= 2048; // reserve 2k for header
             var chunk = new Collection<NetworkMessage>();
+            int notificationsPerMessage = 0;
```
luiscantero, line 94: Pls make ulong

PR #660, opened 2020-08-06
Support aad-register from the Cloud Shell

luiscantero:
* aad-register.ps1 can be called directly from the Cloud Shell after git cloning the repo
* add-register.ps1 supports the parameter -tenantId
* deploy.ps1 also now works in the Cloud Shell, even though deployment takes about 30 min and the shell gets disconnected after 20 min
* Added new info to the doc
* Some cosmetic changes to the doc and other files
* Tested the following scenarios:
1. Deploy from local PC using deploy.ps1 with and without authTenantId
2. Deploy from the Cloud Shell using deploy.ps1 with and without authTenantId
3. Generate aad.json using aad-register.ps1 with and without tenantId and deploy both versions from a PC using -aadConfig param

### DIFF
--- a/deploy/templates/azuredeploy.minimum.json
+++ b/deploy/templates/azuredeploy.minimum.json
@@ -20,7 +20,7 @@
             "type": "string",
             "defaultValue": "master",
             "metadata": {
-                "description": "The branch from which to deploy deploy services and application. Default to master."
+                "description": "The branch from which to deploy services and application. Default to master."
             }
         },
         "storageName": {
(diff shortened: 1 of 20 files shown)


deploy/templates/azuredeploy.minimum.json
```diff
@@ -6,21 +6,21 @@
...
-                "description": "The artifacts url from which to pull all linked templates.  Default is official repository."
+                "description": "The artifacts url from which to pull all linked templates. Default is official repository."
             }
         },
         "branchName": {
             "type": "string",
             "defaultValue": "master",
             "metadata": {
-                "description": "The branch from which to deploy deploy services and application.  Default to master."
+                "description": "The branch from which to deploy deploy services and application. Default to master."
```
monikavar, line 23: minor: deploy two times.

luiscantero: Fixed

PR #661, opened 2020-08-07
Remove source of confusion

bamajeed:
<application-name> should be instead of <name> otherwise if a they are different then the return url will not be set correctly


luiscantero:
@bamajeed The application name doesn't have to be the same as app registration name. e.g. some companies may require the owner's name or dept. to be embedded in the app registration name, whereas the app service name can be something completely different. But this indeed seems to be a source of confusion and this change probably covers most case-scenarios. Thank you!

PR #668, opened 2020-08-10
Replace Collection with List

luiscantero:
Replace Collection<> with List<> in encoder classes as List<> supports initial capacity and improves performance


luiscantero:
List<> is faster in debug mode, Collection<> in release. Closing.

PR #666, opened 2020-08-10
Improvements to diag info

luiscantero:
* Reformat diag info for easier readability and consistency
* Added units to diag info values
* Corrected term "outgress" to "egress" including option em|maxegressmessagequeue
* Changed default egress queue from 20 (max 50 MB) to 8192 (max 2 GB), as it behaved better during local debugging in VS, otherwise too many messages are dropped
* Some cosmetic improvements

image

PR #672, opened 2020-08-12
Revert diag info changes from #666/670

luiscantero:


luiscantero:
As discussed with Erich

PR #674, opened 2020-08-13
Replaced references of Pub/Sub with PubSub as per official naming.

karok2m:
Replaced references of Pub/Sub with PubSub as per official naming.


karok2m:
@barnstee, @mregen, @koepalex, I will wait for one of you to approve this as well before merging as you are the domain experts on this.

PR #689, opened 2020-09-03
Bugfix/publisher

cristipogacean:
* fixed creation of duplicate dataSetWriterId in job definition bug
 * changed NodePublishBulkAsync behavior to generate a new dataSetWriterId respective a new dataset for every new api call

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Services/PublisherJobService.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Services/PublisherJobService.cs
@@ -312,7 +312,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Publisher.Clients {
                     DiagnosticsInterval = TimeSpan.FromSeconds(60),
                     MaxMessageSize = 0,
                     MaxOutgressMessages = DefaultMaxOutgressMessages.Value
-        },
+                },
                 ConnectionString = null
             };
         }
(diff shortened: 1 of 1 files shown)


monikavar (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Services/PublisherJobService.cs
```diff
@@ -335,8 +342,7 @@ public sealed class PublisherJobService : IPublishServices<string> {
             List<PublishedDataSetVariableModel> variables = null;
             foreach (var writer in publishJob.WriterGroup.DataSetWriters) {
                 if (writer.DataSet.DataSetSource.Connection.IsSameAs(connection) &&
-                    writer.DataSet.DataSetSource.SubscriptionSettings?.PublishingInterval ==
```
barnstee, line 338: Why remove this?

barnstee (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Services/PublisherJobService.cs
cristipogacean: the dataSetWriterId is now EndpointId+samplingInterval, so the sampling interval comparison was replaced with dataSetWriterId comparison. It would had been redundant to keep it.

PR #691, opened 2020-09-03
bind mounts for layered Iot Edge deployments

cristipogacean:
* added host bind mounts for layered IoT Edge deployments
* the PkiRootPath is set to the same mount storage for all edge modules so that the modules can share the same pki store


barnstee (approved):

karok2m:
@cristipogacean, what will happen if `/mount/` directory does not exist on the host ?

cristipogacean:
@karok2m, on Linux if the /mount/ does not exist on the host it will be created. On Windows host if the folder does not exist the container fails to start, that's the reason I defaulted it to the ProgramData/iotedge folder, that must exist as long as the IotEdge runtime is installed.

karok2m (approved):

PR #693, opened 2020-09-04
fix Invalid values are set when heartbeat and job check intervals are…

dacolgit:
… removed.


cristipogacean (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Services/PublisherRegistry.cs
```diff
@@ -133,31 +133,26 @@ public sealed class PublisherRegistry : IPublisherRegistry, IPublisherEndpointQu
...
-                                request.Configuration.MaxWorkers <= 0 ?
-                                    null : request.Configuration.MaxWorkers;
-                        }
-                        if (request.Configuration.Capabilities != null) {
-                            patched.Configuration.Capabilities =
-                                request.Configuration.Capabilities.Count == 0 ?
-                                    null : request.Configuration.Capabilities;
-                        }
+                        
+                        patched.Configuration.JobOrchestratorUrl =
```
karok2m, line 137: @dacolgit, will this differentiate between the cases where `JobCheckInterval` is explicitly set to `null` in the request body and where it is omitted ? Maybe this is related to serialization/deserialization of request body.

So, in the case of this body, I am assuming that `JobCheckInterval` will be set to `null`, right ?

```json
{
  "configuration": {
    "JobCheckInterval": null
  }
}
```

But in the case of this body, where `JobCheckInterval` is not present, `JobCheckInterval` will not be modified, right ?

```json
{
  "configuration": {
    "HeartbeatInterval": "00:02:00"
  }
}
```

dacolgit: yes

cristipogacean (dismissed):

PR #695, opened 2020-09-04
Added configuration of default messaging mode and message encoding.

karok2m:
Added the following environment variables:
* `PCS_DEFAULT_PUBLISH_MESSAGING_MODE`: for specifying default messaging mode for jobs definitions for OPC Publisher module
* `PCS_DEFAULT_PUBLISH_MESSAGE_ENCODING`: for specifying default message encoding for jobs definitions for OPC Publisher module


cristipogacean (approved):
The code is good for me. The solution though is a workaround for the fact that the API is incomplete. We must fix the API anyhow and when the API is fixed, this workaround is superfluous.

karok2m:
> The code is good for me. The solution though is a workaround for the fact that the API is incomplete. We must fix the API anyhow and when the API is fixed, this workaround is superfluous.

There is a separate work item for exposing functionality of setting messaging mode and message encoding through APIs.

PR #700, opened 2020-09-07
adding date to log entries

monikavar:
Adding timestamp in UTC in log template.

### DIFF
--- a/common/src/Microsoft.Azure.IIoT.Core/src/Diagnostics/Extensions/SerilogEx.cs
+++ b/common/src/Microsoft.Azure.IIoT.Core/src/Diagnostics/Extensions/SerilogEx.cs
@@ -66,6 +66,6 @@ namespace Serilog {
         }
 
         private const string kDefaultTemplate =
-            "[{Timestamp:yyyy-MM-dd HH:mm:ss} {Level:u3} {SourceContext:lj}] {Message:lj} {NewLine}{Exception}";
+            "[{Timestamp:u} {Level:u3} {SourceContext:lj}] {Message:lj} {NewLine}{Exception}";
     }
 }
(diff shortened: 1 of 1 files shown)


barnstee (approved):

common/src/Microsoft.Azure.IIoT.Core/src/Diagnostics/Extensions/SerilogEx.cs
```diff
@@ -66,6 +66,6 @@ public static class SerilogEx {
         }
 
         private const string kDefaultTemplate =
-            "[{Timestamp:HH:mm:ss} {Level:u3} {SourceContext:lj}] {Message:lj} {NewLine}{Exception}";
+            "[{Timestamp:yyyy-MM-dd HH:mm:ss} {Level:u3} {SourceContext:lj}] {Message:lj} {NewLine}{Exception}";
```
karok2m, line 69: @monikavar, can the time zone also be added ? UTC would be preferable.

monikavar: Good point!
What would be more preferable? UTC time or local time with timezone?
e.g.,
1. `2020-09-07 14:22:29Z`
2. `2020-09-07 16:22:29 +02:00`

karok2m: I would prefer UTC one.

monikavar: done.

cristipogacean (approved):

karok2m (approved):

PR #705, opened 2020-09-10
Revert param and env name for compat with 2.7.183

luiscantero:
* Revert egress to outgress in param name and environment variable to avoid breaking change with 2.7.180/183
* Egress wording stays otherwise

### DIFF
--- a/common/src/Microsoft.Azure.IIoT.Abstractions/src/PcsVariable.cs
+++ b/common/src/Microsoft.Azure.IIoT.Abstractions/src/PcsVariable.cs
@@ -294,6 +294,6 @@ namespace Microsoft.Azure.IIoT {
             "PCS_LOGS_PATH";
         /// <summary> The maximum size of the (IoT D2C) message egress queue </summary>
         public const string PCS_MAX_EGRESS_MESSAGE_QUEUE =
-            "PCS_DEFAULT_PUBLISH_MAX_EGRESS_MESSAGE_QUEUE";
+            "PCS_DEFAULT_PUBLISH_MAX_OUTGRESS_MESSAGES";
     }
 }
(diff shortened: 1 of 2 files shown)


common/src/Microsoft.Azure.IIoT.Abstractions/src/PcsVariable.cs
```diff
@@ -294,6 +294,6 @@ public static class PcsVariable {
             "PCS_LOGS_PATH";
         /// <summary> The maximum size of the (IoT D2C) message egress queue </summary>
         public const string PCS_MAX_EGRESS_MESSAGE_QUEUE =
```
karok2m, line 296: Please rename PcsVariable as well. Having PcsVariable with a value different from its name is very confusing.

karok2m (changes requested):

common/src/Microsoft.Azure.IIoT.Abstractions/src/PcsVariable.cs
luiscantero: The idea is here to just rename the external API surface to avoid breaking changes. Everything else will remain as egress as there is outgress.

karok2m: I understand that, I see that internal models still have `MaxEgressMessageQueue`. What I am asking for is that the PCS variable has the same name as its value. PcsVariable is for having a centralized location for storing environment variables that we use for configuration. So please only rename `PcsVariable.PCS_MAX_EGRESS_MESSAGE_QUEUE` to `PcsVariable.PCS_DEFAULT_PUBLISH_MAX_OUTGRESS_MESSAGES` to reflect the value that we will expect from environment, which is `PCS_DEFAULT_PUBLISH_MAX_OUTGRESS_MESSAGES`.

luiscantero:
Will take another approach

PR #711, opened 2020-09-14
Hotfix/edge update

cristipogacean:
* edge layered deployment update to edge runtime 1.0.9.4
* setting publisher job's iotdevice as child of the iotedge device executing the job

### DIFF
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Jobs/Default/DefaultJobOrchestrator.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Jobs/Default/DefaultJobOrchestrator.cs
@@ -66,7 +66,7 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Jobs {
                                 // Try again on the current value in the database
                                 jobProcessInstruction = CalculateInstructions(existingJob, workerId);
                                 if (jobProcessInstruction != null) {
-                                    _jobEventHandler.OnJobScopingAsync(_jobService, jobProcessInstruction.Job, workerId).ConfigureAwait(false);
+                                    _jobEventHandler.OnJobAssignmentAsync(_jobService, jobProcessInstruction.Job, workerId).ConfigureAwait(false);
                                 }
                                 return Task.FromResult(jobProcessInstruction != null);
                             }, ct);
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Jobs/Default/DefaultJobOrchestrator.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Jobs/Default/DefaultJobOrchestrator.cs
@@ -24,13 +24,19 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Jobs {
         /// <param name="workerRepository"></param>
         /// <param name="demandMatcher"></param>
         /// <param name="jobOrchestratorConfig"></param>
+        /// <param name="jobEventHandler"></param>
+        /// <param name="jobService"></param>
         public DefaultJobOrchestrator(IJobRepository jobRepository,
             IWorkerRepository workerRepository, IDemandMatcher demandMatcher,
-            IJobOrchestratorConfig jobOrchestratorConfig) {
+            IJobOrchestratorConfig jobOrchestratorConfig,
+            IJobEventHandler jobEventHandler,
+            IJobService jobService) {
             _jobRepository = jobRepository;
             _demandMatcher = demandMatcher;
             _workerRepository = workerRepository;
             _jobOrchestratorConfig = jobOrchestratorConfig;
+            _jobEventHandler = jobEventHandler;
+            _jobService = jobService;
         }
 
         /// <inheritdoc/>
@@ -59,6 +65,9 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Jobs {
                             await _jobRepository.UpdateAsync(job.Id, existingJob => {
                                 // Try again on the current value in the database
                                 jobProcessInstruction = CalculateInstructions(existingJob, workerId);
+                                if (jobProcessInstruction != null) {
+                                    _jobEventHandler.OnJobScopingAsync(_jobService, jobProcessInstruction.Job, workerId).ConfigureAwait(false);
+                                }
                                 return Task.FromResult(jobProcessInstruction != null);
                             }, ct);
                             return jobProcessInstruction;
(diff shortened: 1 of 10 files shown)


common/src/Microsoft.Azure.IIoT.Core/src/Hub/Client/IoTHubServiceHttpClient.cs
```diff
@@ -69,15 +69,42 @@ public sealed class IoTHubServiceHttpClient : IoTHubHttpClientBase,
...
+                            capabilities = twin.Capabilities
+                        });
+                    }
                     var response = await _httpClient.PutAsync(device, ct);
                     response.Validate();
                 }
-                catch (ConflictingResourceException)
-                    when (!string.IsNullOrEmpty(twin.ModuleId) || force) {
+                catch (ConflictingResourceException) {
+                    // try update the device scope if provided and not iotedge device and not a module
```
barnstee, line 89: I don't understand this comment

common/src/Microsoft.Azure.IIoT.Core/src/Hub/Client/IoTHubServiceHttpClient.cs
```diff
@@ -69,15 +69,42 @@ public sealed class IoTHubServiceHttpClient : IoTHubHttpClientBase,
...
+                    if ((twin.Capabilities?.IotEdge).GetValueOrDefault(false) &&
```
barnstee, line 90: if twin.Capabilities?.IotEdge is null, this will generate an exception

services/src/Microsoft.Azure.IIoT.Services.OpcUa.Publisher/src/Startup.cs
```diff
@@ -223,6 +223,7 @@ public class Startup {
                 .AsImplementedInterfaces().SingleInstance();
             builder.RegisterType<CosmosDbServiceClient>()
                 .AsImplementedInterfaces();
+            // TODO check why duplicate?
```
barnstee, line 226: Please verify why we have this before check in and remove the TODO

barnstee (changes requested):

common/src/Microsoft.Azure.IIoT.Core/src/Hub/Client/IoTHubServiceHttpClient.cs
cristipogacean: re-phased

common/src/Microsoft.Azure.IIoT.Core/src/Hub/Client/IoTHubServiceHttpClient.cs
cristipogacean: If twin itself is null, indeed it generates an exception, but this cannot happen since twin argument is validated in the beginning of the method.  GetValueOrDefault(false) will make sure the end result is false in case either Capabilities or IoTEdge are null.

services/src/Microsoft.Azure.IIoT.Services.OpcUa.Publisher/src/Startup.cs
cristipogacean: ok, I removed the duplicated code

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Jobs/Hub/IoTHubJobConfigurationHandler.cs
```diff
@@ -46,6 +46,39 @@ public class IoTHubJobConfigurationHandler : IJobEventHandler {
...
+            }
+        }
+
+        /// <inheritdoc/>
+        public async Task OnJobScopingAsync(IJobService manager, JobInfoModel job, string workerId) {
+            if (job.JobConfiguration?.IsObject != true) {
+                return;
+            }
+            try {
+                var edgeDeviceTwin = await _ioTHubTwinServices.FindAsync(workerId.Split("_publisher")[0]);
```
koepalex, line 61: >workerId.Split("_publisher")[0] [](start = 73, length = 31)

Avoid potential crash, do not access index zero

koepalex (approved):
:shipit:

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Jobs/Hub/IoTHubJobConfigurationHandler.cs
cristipogacean: right, I added an argument validation to make sure the workerId is not null. The split itself will always return an array with size at lease 1, therefore index zero always exist.

barnstee (approved):

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Jobs/Default/DefaultJobOrchestrator.cs
```diff
@@ -59,6 +65,9 @@ public class DefaultJobOrchestrator : IJobOrchestrator {
                             await _jobRepository.UpdateAsync(job.Id, existingJob => {
                                 // Try again on the current value in the database
                                 jobProcessInstruction = CalculateInstructions(existingJob, workerId);
+                                if (jobProcessInstruction != null) {
+                                    _jobEventHandler.OnJobScopingAsync(_jobService, jobProcessInstruction.Job, workerId).ConfigureAwait(false);
```
marcschier, line 69: Better name might be "OnJobAssignmentAsync"

common/src/Microsoft.Azure.IIoT.Core/src/Hub/Client/IoTHubServiceHttpClient.cs
marcschier: This changes current behavior for everything using patch.  My preference would be to leave the original code and specify "force = true".

services/src/Microsoft.Azure.IIoT.Services.OpcUa.Publisher/src/Startup.cs
```diff
@@ -223,8 +223,6 @@ public class Startup {
                 .AsImplementedInterfaces().SingleInstance();
             builder.RegisterType<CosmosDbServiceClient>()
                 .AsImplementedInterfaces();
-            builder.RegisterType<DefaultJobService>()
```
marcschier, line 226: Sure?

marcschier (approved):

cristipogacean:
_Note_: with this change, connection string of the IoT device associated to the job is no longer persisted in the jobs config json stored in Cosmos DB, but is resolved every time a job is pushed to a publisher for being executed.

marcschier (approved):

services/src/Microsoft.Azure.IIoT.Services.OpcUa.Publisher/src/Startup.cs
karok2m: It is a duplicate registration. Another registration can be found on line `218`.

PR #717, opened 2020-09-16
Use latest OPC stack

mregen:
- use latest 363 bits, now supporting netstandard2.1
- include preview feed of OPC Foundation for development
- fix also a few build errors due to warnings


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Transport/Http/WebSocketChannelListener.cs
```diff
@@ -213,6 +214,18 @@ public class WebSocketChannelListener : ITcpChannelListener,
                 }));
         }
 
+        bool ITcpChannelListener.ReconnectToExistingChannel(IMessageSocket socket, uint requestId, uint sequenceNumber, uint channelId, X509Certificate2 clientCertificate, ChannelToken token, OpenSecureChannelRequest request) {
```
mregen, line 217: add inheritdoc

NuGet.Config
```diff
@@ -3,5 +3,6 @@
   <packageSources>
     <add key="nugetv3" value="https://api.nuget.org/v3/index.json" />
     <add key="myget" value="https://www.myget.org/F/azureiiot/api/v3/index.json" />
+    <add key="opcua-preview" value="https://opcfoundation.pkgs.visualstudio.com/opcua-netstandard/_packaging/opcua-preview/nuget/v3/index.json" />
```
barnstee, line 6: How do we make sure we don't accidentally pull in preview packages? Are they marked "-preview"?

barnstee (approved):

NuGet.Config
mregen: yep, they are all marked -preview

koepalex (approved):
:shipit:

PR #718, opened 2020-09-17
Show warning for deprecated Publisher params

luiscantero:
* Show warning if param or environment variable has been deprecated, warn that it will be removed in a future version
* If specified, suggest using new version
* If both are present, newer version overrides deprecated one
* Added to deprecation list: om|maxoutgressmessages -> em|maxegressmessages 
* Added to deprecation list: PCS_DEFAULT_PUBLISH_MAX_OUTGRESS_MESSAGES -> PCS_DEFAULT_PUBLISH_MAX_EGRESS_MESSAGE_QUEUE

image


modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/ModuleProcess.cs
```diff
@@ -134,6 +137,40 @@ public class ModuleProcess : IProcessControl {
...
+        private void CheckDeprecatedParams(ILogger logger) {
+            // List with pairs of deprecated and new/replacement options.
+            // If newOption is null, warning will not suggest using it instead.
+            var deprecatedOptions = new List<(string deprecatedOption, string newOption)> {
+                // Deprecated on 2020-09-17.
+                (LegacyCliConfigKeys.MaxOutgressMessages,
+                 LegacyCliConfigKeys.MaxEgressMessageQueue),
+
+                // Deprecated on 2020-09-17.
+                (PcsVariable.PCS_DEFAULT_PUBLISH_MAX_OUTGRESS_MESSAGES,
```
karok2m, line 154: Both `PCS_DEFAULT_PUBLISH_MAX_OUTGRESS_MESSAGES` and `PCS_DEFAULT_PUBLISH_MAX_EGRESS_MESSAGE_QUEUE` are part of backend OPC Publisher **Microservice** configuration. As far as I understand, none of them should be used for configuring OPC Publisher **Module**. I don't see them being used in module either. Can you please validate this and remove this case from the check here if they are not used ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Runtime/PublishServicesConfig.cs
```diff
@@ -70,8 +70,16 @@ public class PublishServicesConfig : ConfigBase, IPublishServicesConfig {
...
-                    () => GetIntOrDefault(PcsVariable.PCS_MAX_EGRESS_MESSAGE_QUEUE,
-                    () => _DefaultMaxEgressMessageQueue));
+                    () => GetIntOrDefault(PcsVariable.PCS_DEFAULT_PUBLISH_MAX_EGRESS_MESSAGE_QUEUE,
+                    () => -1));
+
+                if (queueSize == -1) {
+                    // Fallback to deprecated option,
+                    // use PCS_DEFAULT_PUBLISH_MAX_EGRESS_MESSAGE_QUEUE instead.
+                    queueSize = GetIntOrDefault(kDefaultMaxEgressMessageQueue,
+                        () => GetIntOrDefault(PcsVariable.PCS_DEFAULT_PUBLISH_MAX_OUTGRESS_MESSAGES,
```
karok2m, line 80: Would you mind also writing a test that validates that it is applied ? Something similar to `PublishServicesConfigTest2Async()` in `components\opc-ua\src\Microsoft.Azure.IIoT.OpcUa.Publisher\tests\Clients\PublisherJobClientTests.cs`.

karok2m (changes requested):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Runtime/PublishServicesConfig.cs
luiscantero: I cannot test it like PublishServicesConfigTest2Async because the deprecated keys are mapped to the new ones before they reach the "configuration" object.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/ModuleProcess.cs
luiscantero: I have removed DEFAULT_PUBLISH_MAX_OUTGRESS_MESSAGES. Will look for a better place (as part of another PR) or create a PBI to handle all deprecated PCS, as discussed.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Runtime/PublishServicesConfig.cs
```diff
@@ -70,8 +70,16 @@ public class PublishServicesConfig : ConfigBase, IPublishServicesConfig {
...
                 var queueSize = GetIntOrDefault(kDefaultMaxEgressMessageQueue,
-                    () => GetIntOrDefault(PcsVariable.PCS_MAX_EGRESS_MESSAGE_QUEUE,
-                    () => _DefaultMaxEgressMessageQueue));
+                    () => GetIntOrDefault(PcsVariable.PCS_DEFAULT_PUBLISH_MAX_EGRESS_MESSAGE_QUEUE,
+                    () => -1));
+
+                if (queueSize == -1) {
+                    // Fallback to deprecated option,
+                    // use PCS_DEFAULT_PUBLISH_MAX_EGRESS_MESSAGE_QUEUE instead.
+                    queueSize = GetIntOrDefault(kDefaultMaxEgressMessageQueue,
```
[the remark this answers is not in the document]
luiscantero: Done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Runtime/PublishServicesConfig.cs
karok2m: Yes, you are correct about the mapping. What I mean is that you can test that when `PCS_DEFAULT_PUBLISH_MAX_OUTGRESS_MESSAGES` is set to some value `x` that value is actually applied. For example, you can check that `Confi.DefaultMaxEgressMessageQueue` has that value `x` or that job definition has that value `x` when job is created. If you want to, you can mark this conversation as resolved as well, since we know that general mechanism works.

luiscantero: What I meant was that I cannot test the params maxoutgressmessages/maxoutgressmessages in the same fashion, but I can test the PCS variables. Added two tests for that.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Runtime/PublishServicesConfig.cs
```diff
@@ -42,7 +50,7 @@ public class PublishServicesConfig : ConfigBase, IPublishServicesConfig {
                     () => _defaultBatchTriggerInterval));
 
                 if (batchInterval.TotalMilliseconds >= 100 && batchInterval.TotalMilliseconds <= 3600000) {
-                    //ToDo: This behavior should be logged.
+                    _logger.Verbose($"DefaultBatchTriggerInterval set to {batchInterval}");
```
luiscantero, line 53: @karok2m I added logging to this file. Is this what you had in mind? Verbose, Debug or Information?

services/src/Microsoft.Azure.IIoT.Services.OpcUa.Publisher/src/Startup.cs
```diff
@@ -66,7 +66,8 @@ public class Startup {
         /// </summary>
         /// <param name="env"></param>
         /// <param name="configuration"></param>
-        public Startup(IWebHostEnvironment env, IConfiguration configuration) :
+        /// <param name="logger"></param>
+        public Startup(IWebHostEnvironment env, IConfiguration configuration, ILogger logger) :
```
karok2m, line 70: Can you please validate that `Microsoft.Azure.IIoT.Services.OpcUa.Publisher` project is able to start and run ? DI container builder is not configured yet at this point, so I am not sure that it will be able to resolve `ILogger` dependency.

I got this error, when I only added `ILogger` parameter for this constructor of `Startup`:

```
Exception thrown: 'System.InvalidOperationException' in Microsoft.Extensions.DependencyInjection.Abstractions.dll
An unhandled exception of type 'System.InvalidOperationException' occurred in Microsoft.Extensions.DependencyInjection.Abstractions.dll
Unable to resolve service for type 'Serilog.ILogger' while attempting to activate 'Microsoft.Azure.IIoT.Services.OpcUa.Publisher.Startup'.
```

karok2m (changes requested):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Runtime/PublishServicesConfig.cs
luiscantero: Ok, so the warning should be shown inside of "else" and not in the "if". I simplified the msg as follows as it makes managing the ranges easier: "DefaultBatchTriggerInterval: Provided value {batchInterval} should be >= 100 and <= 3600000. Defaulting to {_defaultBatchTriggerInterval.TotalMilliseconds}.".
Let me know what you think :)

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Runtime/PublishServicesConfig.cs
```diff
@@ -10,11 +10,13 @@ namespace Microsoft.Azure.IIoT.OpcUa.Publisher.Runtime {
     using Microsoft.Azure.IIoT.OpcUa.Publisher.Models;
     using Microsoft.Extensions.Configuration;
     using System;
+    using Serilog;
 
     /// <summary>
     /// Configuration of defaults for job definition generation logic for OPC Publisher module.
     /// </summary>
     public class PublishServicesConfig : ConfigBase, IPublishServicesConfig {
+        private readonly ILogger _logger;
```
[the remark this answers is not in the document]
luiscantero: Yes, fixed

services/src/Microsoft.Azure.IIoT.Services.OpcUa.Publisher/src/Startup.cs
luiscantero: Hmm, the standalone publisher works, but you are right, the service doesn't and it doesn't look good: [Generic Host restricts Startup constructor injection](https://github.com/dotnet/aspnetcore/issues/9337#issuecomment-539859667). I will look into it.

karok2m (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Runtime/PublishServicesConfig.cs
karok2m: Message looks good.

services/src/Microsoft.Azure.IIoT.Services.OpcUa.Publisher/src/Startup.cs
karok2m: With current changes OPC Publisher microservice works fine.

PR #723, opened 2020-09-21
Remove OPC Vault and some other stuff

marcschier:


mregen (approved):
:shipit:

mregen:
>     using System.Globalization;

these tests might be useful, but the class is gone?

---
Refers to: components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/tests/Stack/Encoders/TypeSerializerTests.cs:10 in 5ed079a. [](commit_id = 5ed079ad6c407e569c9bd8936d85ddf741482f24, deletion_comment = True)

barnstee (changes requested):
Can you split this up into smaller chunks and specify what you mean with "other stuff"? Ideally, 1 PR per removed feature.

PR #730, opened 2020-09-23
fix Publisher fail to start if max IoT Messages quota is reached

dacolgit:


barnstee (approved):

common/src/Microsoft.Azure.IIoT.Core/src/Utils/Retry.cs
```diff
@@ -112,7 +112,12 @@ public static class Retry {
                     return await work();
                 }
                 catch (Exception ex) {
-                    await DelayOrThrow(logger, cont, policy, maxRetry, k, ex, ct);
+                    if (ex is IotHubQuotaExceededException) {
+                        logger.Error(ex.Message);
+                    } 
+                    else {
+                        await DelayOrThrow(logger, cont, policy, maxRetry, k, ex, ct);
+                    }
```
karok2m, line 120: @dacolgit, please revert back changes to `Retry.cs`. Default logic for continuation in `Retry` class is such that only exceptions of type `ITransientException` should be ignored. If you need a different retry condition, you can pass it from the caller as `cont` lambda.

Also, exceptions of type `IotHubQuotaExceededException` do not seem to be transient in nature, so immediate retries should not result in a successful call. So I would not even suggest passing that as a condition for retry continuation.

karok2m (changes requested):

karok2m (approved):

mregen (approved):

PR #734, opened 2020-09-24
Only create the sarif files in a cloud build

mregen:


mregen:
note: TSA upload seems to have intermittent issues, not because of this change

.gitignore
```diff
@@ -17,6 +17,7 @@
 /**/*.vc.db
 /.vscode/**
 /build/**
+*.sarif
```
karok2m, line 20: I am afraid that this will result in people not noticing that `*.sarif` files are created when they shouldn't be.

karok2m (approved):

.gitignore
mregen: I mean its just for git, and with my change the Errorlog property is just enabled in cloud builds...

PR #735, opened 2020-09-24
Prevent non-secrets from being masked in logs

dan-silver:
Edit:
Instead of the below approach we're no longer parsing the output from IAI and saving those as secret variables. This is a much cleaner way to solve the problem and we're getting rid of the script that parses the IAI .env file.

--------

When a variable is marked as a secret in an Azure pipeline it makes sure that secret is never leaked in logs by replacing it with ```*```. Right now, over 60 values are parsed from the .env file (created by IAI) and marked as secrets but only 4 are used.

The issue is that if a secret is ```a``` then all occurrences of that character will be replaced by ```*```. Logs looked very strange due to some secrets being short and this made developing new features difficult when logs aren't readable. Examples:
```
2020-09-24T18:56:39.9516731Z VERBOSE: Exporting function 'Out-***'.
2020-09-24T18:56:40.4989587Z ##[debug]Loaded ***6 strings.
2020-09-24T18:56:40.9062876Z VERBOSE: Importing function 'Out-Vsts***'.
2020-09-24T18:56:41.2701041Z VERBOSE: AGENT_VERSION: '2.***7***.***'
```
This change adds a parameter to ```parseenvfile.ps1``` for specifying which values to save as secrets instead of extracting all values from the .env file.


tools/e2etesting/pipeline.yml
```diff
@@ -40,7 +40,7 @@ jobs:
     inputs:
       azureSubscription: '$(AzureSubscription)'
       KeyVaultName: '$(ClientCredentialsKeyVaultName)'
-      SecretsFilter: '*'
+      SecretsFilter: 'IAIClientId,IAIClientSecret'
```
dan-silver, line 43: Note the change to this line is unrelated but still a good idea to only download the secrets we need from keyvault.

tools/e2etesting/ParseEnvFile.ps1
```diff
@@ -1,5 +1,6 @@
 param(
```
markusstuhler, line 1: I think we don't need this step anymore. I only used it to read the name of the KeyVault. However, the DetermineKeyVaultName.ps1 script uses another approach (by selecting the KeyVault name based on the 'application'-Tag.
The secrets will later be retrieved directly from the KeyVault. We should add a SecretsFilter in the step "Retrieve KeyVault secrets from existing deployment" below.

tools/e2etesting/pipeline.yml
markusstuhler: I think we should investige whether IAI can use the authentication context that Azure DevOps provides somehow so that we don't need this step at all.

tools/e2etesting/ParseEnvFile.ps1
dan-silver: Great catch. I've removed this step and added a secrets filter to that key vault retrieval step.

mregen (approved):

karok2m (approved):

PR #738, opened 2020-09-27
do not change OPC UA namespace URI after parsing it.

dovjar:
Current implementation of OPC UA nodeId parser uses Uri class from net core. This class adds trailing slash after authority e.g. http://authority will become http://authority/

From RFC 3986:
When authority is present, the path must
   either be empty or begin with a slash ("/") character.

As i understand uri http://authority  is correct.

Some siemens PLC OPc UA servers reports Node id in format: http://serviceInterface#i=1 , 
NodeId parser transforms this nodeId to: http://serviceInterface/#i=1 and when Twin module queries by this new transformed nodeId - get the error "BadNodeIdUnknown".

Proposed change:
ParseNodeIdUri function do not changes NodeId namespace part if it is valid Uri.


mregen:
Hi @dovjar, thanks for the PR and for debugging the issue. We need to discuss this topic, it might be better to always use the normalized Uri and to fix the twin module. Please stay tuned.

mregen (changes requested):
Hi @dovjar, I debugged your fix but I don't see your described behaviour in this function:

`nsUri = uri.NoQueryAndFragment().AbsoluteUri;`

It always produces the proper nsuri, with or without the post '/' as in the passed in uri.

Please add some test cases in NodeIdExTests.cs which show the improper behaviour, otherwise we can't take your fix.

dovjar:
Hi,
added to new tests:
DecodeNodeIdFromIntUrl - pass (because of my proposed change in ParseNodeId function, can comment my changes and it will fail)
and
ParseNodeIdUsingAbsoluteUri

the last one fails.
providing "http://contoso.com#i=1", expecting ns would be "http://contoso.com".

If i'll provide "http://contoso.co,/ua" - it will parse correctly "http://contoso.com/ua", because there is the path fragment "/ua".

Unexpected behavior occurs then providing uri without path fragment, like "http://contoso.com",

mregen:
Thanks @dovjar , now it makes sense...

PR #740, opened 2020-09-28
Build IAI in test pipeline

dan-silver:
Instead of pulling latest IAI from blob storage, we should build IAI from the branch being tested. The CI pipeline already had tasks for building IAI so these have been extracted into ```build_iiot_deployment.yml``` and reused in both pipelines.

Also note that ```./tools/scripts/set-version.ps1``` uses https://github.com/dotnet/Nerdbank.GitVersioning and it's configured to modify the pipeline run name. The pipeline name was being used as the resource group name, which will no longer work since it has special characters. So the resource group name variable has been modified to: ```$[format('deployment{0:yyyyMMdd}{1}', pipeline.startTime, counter(variables.AgentPool, 1))]```

Another task has been created for deploying IAI services from a private container registry.


karok2m (approved):

tools/templates/build_iiot_deployment.yml
```diff
@@ -0,0 +1,25 @@
+parameters:
```
marcschier, line 1: Can/should we use this template in our regular IAI build template?

marcschier (approved):

PR #751, opened 2020-10-06
WorkerSupervisor enable the reduction of Worker 

koepalex:
**Issue:**
The number of Worker within publisher can't be reduced. 
Calling *PATCH /registry/v2/publishers/{publisherId}* with smaller number of MaxWorker will not result in terminating worker. Example payload:

```json
{
  "configuration": {
    "maxWorkers": 4
  }
}
```

**Solution:**
WorkerSupervisor have to also reduce number of Worker within EnsureWorker method.

**Others:**
* Updated IWorkerSupervisor interface
  * Removed `CreateWorker` because it wasn't used externally
  * Removed `StopWorker(WorkerId)` because it wasn't used externally
  * Added getter `int NumberOfWorker` to be used from tests
* Added tests for WorkerSupervisor

### DIFF
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
@@ -56,7 +56,7 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
         }
 
         /// <inheritdoc/>
-        public int NumberOfWorker => _instances.Count;
+        public int NumberOfWorkers => _instances.Count;
 
         /// <summary>
         /// Create worker
@@ -77,11 +77,11 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
         }
 
         /// <summary>
-        /// Stop worker
+        /// Stop a single worker
         /// </summary>
         /// <returns>awaitable task</returns>
         private async Task StopWorker() {
-            // sort worker, so that worker in state stopped, stoping or WaitingForJob will terminate first 
+            // sort workers, so that a worker in state Stopped, Stopping or WaitingForJob will terminated first 
             var worker = _instances.OrderBy(kvp => kvp.Key.Status).First();
             var workerId = worker.Key.WorkerId;
             _logger.Information("Stopping worker with id {WorkerId}", workerId);
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
@@ -83,7 +83,9 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
         private async Task StopWorker() {
             // sort worker, so that worker in state stopped, stoping or WaitingForJob will terminate first 
             var worker = _instances.OrderBy(kvp => kvp.Key.Status).First();
-
+            var workerId = worker.Key.WorkerId;
+            _logger.Information("Stopping worker with id {WorkerId}", workerId);
+            
             await worker.Key.StopAsync();
             worker.Value.Dispose();
             _instances.Remove(worker.Key);
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
@@ -26,12 +26,13 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
         /// <param name="lifetimeScope"></param>
         /// <param name="agentConfigProvider"></param>
         /// <param name="logger"></param>
+        /// <param name="timerDelay">The time the supervisor is waiting before ensure worker</param>
         public WorkerSupervisor(ILifetimeScope lifetimeScope,
-            IAgentConfigProvider agentConfigProvider, ILogger logger) {
+            IAgentConfigProvider agentConfigProvider, ILogger logger, int timerDelay = 10) {
             _lifetimeScope = lifetimeScope;
             _agentConfigProvider = agentConfigProvider;
             _logger = logger;
-            _ensureWorkerRunningTimer = new Timer(TimeSpan.FromSeconds(10).TotalMilliseconds);
+            _ensureWorkerRunningTimer = new Timer(TimeSpan.FromSeconds(timerDelay).TotalMilliseconds);
             _ensureWorkerRunningTimer.Elapsed += EnsureWorkerRunningTimer_ElapsedAsync;
         }
 
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
@@ -81,7 +81,7 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
         /// </summary>
         /// <returns>awaitable task</returns>
         private async Task StopWorker() {
-            // sort workers, so that a worker in state Stopped, Stopping or WaitingForJob will terminated first 
+            // sort workers, so that a worker in state Stopped, Stopping or WaitingForJob will terminate first 
             var worker = _instances.OrderBy(kvp => kvp.Key.Status).First();
             var workerId = worker.Key.WorkerId;
             _logger.Information("Stopping worker with id {WorkerId}", workerId);
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
@@ -62,7 +62,7 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
         /// Create worker
         /// </summary>
         /// <returns></returns>
-        public Task<IWorker> CreateWorker() {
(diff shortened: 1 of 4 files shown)


common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
```diff
@@ -107,26 +117,34 @@ public class WorkerSupervisor : IWorkerSupervisor, IDisposable {
...
-                    _logger.Information("Starting worker '{workerId}'...", stoppedWorker.WorkerId);
-                    workerStartTasks.Add(stoppedWorker.StartAsync());
-                }
-                await Task.WhenAll(workerStartTasks);
-                // the configuration might have been changed by workers execution
-                var newWorkers = _agentConfigProvider.Config?.MaxWorkers ?? kDefaultWorkers;
-                if (workers >= newWorkers) {
-                    break;
-                }
+            if (_agentConfigProvider.Config?.MaxWorkers <= 0) {
```
luiscantero, line 121: MaxWorkers can be null

luiscantero (dismissed):

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
```diff
@@ -56,26 +56,36 @@ public class WorkerSupervisor : IWorkerSupervisor, IDisposable {
         }
 
         /// <inheritdoc/>
-        public Task<IWorker> CreateWorker() {
+        public int NumberOfWorker => _instances.Count;
```
barnstee, line 59: Workers

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
```diff
@@ -56,26 +56,36 @@ public class WorkerSupervisor : IWorkerSupervisor, IDisposable {
...
-            }
-
-            var worker = _instances.Where(w => w.Key.WorkerId == workerId).Single();
-
+        /// <summary>
+        /// Stop worker
+        /// </summary>
+        /// <returns>awaitable task</returns>
+        private async Task StopWorker() {
+            // sort worker, so that worker in state stopped, stoping or WaitingForJob will terminate first
```
barnstee, line 84: workers

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/IWorkerSupervisor.cs
```diff
@@ -12,16 +12,8 @@ namespace Microsoft.Azure.IIoT.Agent.Framework {
     public interface IWorkerSupervisor : IHostProcess {
 
         /// <summary>
-        /// Create worker
+        /// The amount of worker currently running
```
barnstee, line 15: workers

barnstee (changes requested):

karok2m:
@koepalex, is it communicated back to the Job Orchestrator or Publisher Services which specific worker is stopped ? I want to make sure that if a stopped Worker was executing a job (let's say all of them were doing something) then this job is not picked up only after 15 minutes when it is marked as stale.

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
koepalex: changed

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/IWorkerSupervisor.cs
koepalex: changed

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
koepalex: changed comment

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
koepalex: Nullable int can be safely checked against numbers, check will return false if property is null.

```c#
void Main()
{
	var cfg = new Config();
	cfg.MaxWorker = null;
	if (cfg?.MaxWorker <= 0) {
		Console.WriteLine("Should not hit");
	}
	else if (cfg?.MaxWorker > 0)
	{
		Console.WriteLine("Should not hit");
	}
	else {
		Console.WriteLine("As expected");
	}
}

class Config 
{
	public int? MaxWorker {get;set;}
}
```

Output:
```pwd
As expected
```

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
```diff
@@ -77,11 +77,11 @@ public class WorkerSupervisor : IWorkerSupervisor, IDisposable {
         }
 
         /// <summary>
-        /// Stop worker
+        /// Stop a single worker
         /// </summary>
         /// <returns>awaitable task</returns>
         private async Task StopWorker() {
-            // sort worker, so that worker in state stopped, stoping or WaitingForJob will terminate first 
+            // sort workers, so that a worker in state Stopped, Stopping or WaitingForJob will terminated first
```
barnstee, line 84: terminate

barnstee (changes requested):

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
luiscantero: Yes, in this case its clear what happens, but otherwise the implicit default behavior may lead to unexpected results, e.g. whatever happens when MaxWorkers > 0

koepalex: When MaxWorkers is bigger than 0, it was explicitly set by the customer (via API) and we have to create or terminate worker to reach the desired number

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
koepalex: In German, we call it "verschlimbessern"  :) changed the text

koepalex:
> @koepalex Alexander Koepke FTE, is it communicated back to the Job Orchestrator or Publisher Services which specific worker is stopped ? I want to make sure that if a stopped Worker was executing a job (let's say all of them were doing something) then this job is not picked up only after 15 minutes when it is marked as stale.
> 
> **UPDATE**: I am asking this because this was the behavior that we saw previously: We reduced the number of workers and restarted the publisher to apply the change. And if a job was executed by (let's say) 6th worker and now publisher has only 5 workers, then the job was assigned to a new Worker only after 15 minutes.

Good catch 🐱‍💻👏, with latest commit I've added that the worker once it is stopped it will send a final heartbeat method with `Status = Stopped`. Will this be enough that the orchestrator reassign the job in your scenario?

common/src/Microsoft.Azure.IIoT.Agent.Framework/tests/Worker/WorkerSupervisorTests.cs
```diff
@@ -0,0 +1,203 @@
...
+
+        private IContainer GetAutofacTestConfiguration() {
+            var containerBuilder = new ContainerBuilder();
+            containerBuilder.RegisterType<TestWorker>().As<IWorker>();
+            var container = containerBuilder.Build();
+
+            return container;
+        }
+
+        private const int kDefaultDelay = 15000; //Internal timer which creates the Worker run every 10 seconds
```
karok2m, line 200: Is it possible to make delay of internal timer configurable and use values smaller than 10 seconds for tests ? Now your tests will wait for 75 seconds only because of the timer.

koepalex: Implemented via optional parameter, were the default value is the original value

common/src/Microsoft.Azure.IIoT.Agent.Framework/tests/Worker/WorkerSupervisorTests.cs
```diff
@@ -0,0 +1,203 @@
+﻿namespace Microsoft.Azure.IIoT.Agent.Framework.Tests.Worker {
```
karok2m, line 1: Copyright notice is missing.

barnstee (approved):

common/src/Microsoft.Azure.IIoT.Agent.Framework/tests/Worker/WorkerSupervisorTests.cs
```diff
@@ -0,0 +1,203 @@
...
+            await Task.Delay(kDefaultDelay); 
+            Assert.Equal(kDefaultMaxWorker, sut.NumberOfWorkers);
+            loggerMock.Verify();
+
+            // clean up
+            await sut.StopAsync();
+        }
+
+        [Fact]
+        public async Task Test_005_IncreaseAndDecreaseMaxWorker_Expect_Exception() {
```
karok2m, line 167: I do not see checks for exceptions in the test so `_Expect_Exception` name seems misleading.

koepalex: changed method name

karok2m (dismissed):

barnstee (approved):

karok2m (approved):

PR #754, opened 2020-10-07
Remove legacy publisher option -to

luiscantero:


barnstee (commented):
Is this option supported by the Publisher?

karok2m:
> Is this option supported by the Publisher?

OPC Publisher does not have `to` option any more.

barnstee (approved):

karok2m (approved):

PR #755, opened 2020-10-08
Collected IoT Edge environment variables in IoTEdgeVariables class.

karok2m:
Changes:
* Collected IoT Edge environment variables in `IoTEdgeVariables` class.
* Fixing a typo.

### DIFF
--- a/common/src/Microsoft.Azure.IIoT.Abstractions/src/IoTEdgeVariables.cs
+++ b/common/src/Microsoft.Azure.IIoT.Abstractions/src/IoTEdgeVariables.cs
@@ -1,0 +1,31 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+namespace Microsoft.Azure.IIoT.Abstractions {
+    using System;
+    using System.Collections.Generic;
+    using System.Text;
+
+    /// <summary>
+    /// Common IoT Edge runtime environment variables.
+    /// </summary>
+    public static class IoTEdgeVariables {
+
+        /// <summary> IoT Edge device id </summary>
+        public const string IOTEDGE_DEVICEID = "IOTEDGE_DEVICEID";
+        /// <summary> IoT Edge module id </summary>
+        public const string IOTEDGE_MODULEID = "IOTEDGE_MODULEID";
+        /// <summary> IoT Edge gateway hostname </summary>
+        public const string IOTEDGE_GATEWAYHOSTNAME = "IOTEDGE_GATEWAYHOSTNAME";
+        /// <summary> IoT Edge workload URI </summary>
+        public const string IOTEDGE_WORKLOADURI = "IOTEDGE_WORKLOADURI";
+        /// <summary> IoT Edge module generation id </summary>
+        public const string IOTEDGE_MODULEGENERATIONID = "IOTEDGE_MODULEGENERATIONID";
+        /// <summary> IoT Edge API version </summary>
+        public const string IOTEDGE_APIVERSION = "IOTEDGE_APIVERSION";
+        /// <summary> IoT Hub hostname </summary>
+        public const string IOTEDGE_IOTHUBHOSTNAME = "IOTEDGE_IOTHUBHOSTNAME";
+    }
+}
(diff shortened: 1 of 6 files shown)


common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Clients/EdgeletClient.cs
```diff
@@ -29,10 +30,10 @@ public sealed class EdgeletClient : ISecureElement {
         /// <param name="logger"></param>
         public EdgeletClient(IHttpClient client, IJsonSerializer serializer,
             ILogger logger) : this(client, serializer,
-            Environment.GetEnvironmentVariable("IOTEDGE_WORKLOADURI")?.TrimEnd('/'),
-            Environment.GetEnvironmentVariable("IOTEDGE_MODULEGENERATIONID"),
-            Environment.GetEnvironmentVariable("IOTEDGE_MODULEID"),
-            Environment.GetEnvironmentVariable("IOTEDGE_APIVERSION"),
+            Environment.GetEnvironmentVariable(IoTEdgeVariables.IOTEDGE_WORKLOADURI)?.TrimEnd('/'),
```
barnstee, line 33: This trim end should be in the class instead of here.

common/src/Microsoft.Azure.IIoT.Hub.Module.Client/src/Default/IoTSdkFactory.cs
```diff
@@ -53,9 +54,9 @@ public sealed class IoTSdkFactory : IClientFactory, IDisposable {
             }
 
             // The runtime injects this as an environment variable
-            var deviceId = Environment.GetEnvironmentVariable("IOTEDGE_DEVICEID");
-            var moduleId = Environment.GetEnvironmentVariable("IOTEDGE_MODULEID");
-            var ehubHost = Environment.GetEnvironmentVariable("IOTEDGE_GATEWAYHOSTNAME");
+            var deviceId = Environment.GetEnvironmentVariable(IoTEdgeVariables.IOTEDGE_DEVICEID);
```
barnstee, line 57: The call to get environment variable should be in the class instead of here.

common/src/Microsoft.Azure.IIoT.Abstractions/src/IoTEdgeVariables.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Microsoft.Azure.IIoT.Abstractions {
+
+    /// <summary>
+    /// Common IoT Edge runtime environment variables.
+    /// </summary>
+    public static class IoTEdgeVariables {
+
+        /// <summary> IoT Edge device id </summary>
+        public const string IOTEDGE_DEVICEID = "IOTEDGE_DEVICEID";
```
barnstee, line 14: These should be properties

barnstee (changes requested):

common/src/Microsoft.Azure.IIoT.Abstractions/src/IoTEdgeVariables.cs
karok2m: `IoTEdgeVariables` is not intended as a property or configuration class. It is only a collection of IoT Edge runtime environment variable names so that we do not use duplicate string literals in the code. This class also gives us the possibility to see where in the code IoT Edge environment variables are referenced. It serves the same purpose as `PcsVariable` collection for out `PCS_` environment variables.

common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Clients/EdgeletClient.cs
karok2m: Fixed.

barnstee (approved):

common/src/Microsoft.Azure.IIoT.Hub.Module.Client/src/Default/IoTSdkFactory.cs
karok2m: @barnstee, as stated in the other comment, `IoTEdgeVariables` is intended as a collection of string literals of IoT Edge runtime environment variable names. I agree that we need a separate configuration class for those values that are coming from IoT Edge runtime as environment variables. But let's do that as a part of a separate PR.

PR #758, opened 2020-10-12
Fix session/subscription cleanup not working when session disconnected

dacolgit:


koepalex (approved):
:shipit:

barnstee (commented):
LGTM

cristipogacean (commented):
what happens when the session owns more than one subscription?

dacolgit:
> what happens when the session owns more than one subscription?
just checked it works

cristipogacean:
I don't think the subscription manager should trigger directly removal of the owning session, but the session manager should exclusively remove sessions.

I have just run in an null reference exception that I have never seen before and I suspect this change is the reason:
image

PR #798, opened 2020-10-15
Removing AppsAndCertRequests CosmosDB container after migration.

karok2m:
Removing `AppsAndCertRequests` CosmosDB container after migration of Application entities from there.


koepalex (approved):
:shipit:

barnstee (approved):

PR #819, opened 2020-10-22
Avoid NullReferenceException in case of job continuation

koepalex:
**Issue:** there are multiple places were the JobProcessor is set to null. Which can lead to NullReferenceException in Cancel method e.g. when job continuation is triggered.

**Solution:** Check for null in Cancel method and update worker status on every place were it set to null to have the information in cosmosdb in sync.


marcschier (approved):

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
```diff
@@ -235,6 +237,8 @@ public class Worker : IWorker, IDisposable {
 
                     // Check if the job is to be continued with new configuration settings
                     if (_jobProcess.JobContinuation == null) {
+                        _jobProcess.Status = WorkerStatus.Stopped;
```
barnstee, line 240: also check for null here?

PR #835, opened 2020-10-28
Publisher 2.5: Use LTS base image for MCR

mregen:
Flagged linux base images because 2.2 is not LTS.


koepalex (approved):
agree, .NET Core 2.2 is end of life

monikavar (approved):

PR #838, opened 2020-10-29
upgrading iotedge version in layered deployment

monikavar:
- Upgrading iotedge version to 1.0.10 
    - As of release 1.0.10, metrics are automatically exposed by default on http port 9600 of the Edge Hub and Edge Agent.
    - To disable metrics exposed by Edge Hub and Edge Agent, please set _MetricsEnabled_ environment variable to false.


karok2m:
@monikavar, please validate that OPC Twin will work with `1.0.10` version of IoT Edge. OPC Twin uses credentials of non-child leaf IoT device to start `DeviceClient`s for endpoints, which will most probably not work with `1.0.10` version. So please make sure that you can activate an endpoint and browse nodes in OPC UA server.

PR #850, opened 2020-11-04
fix bug: Fails when more 468 values are added to publisher

dacolgit:
the root cause of this bug is the explicit setting of the parameter --ms=0.
This parameter should configure the message size for IoTHuB, but it wrongly set also the message size in the publisher/client connection.  
This parameter was wrongly used to set the MaxMessageSize in TransportQuotaConfig in EndpointConfiguration


cristipogacean (approved):

koepalex (approved):
:shipit:

monikavar (approved):

PR #859, opened 2020-11-08
Get the semantic version for docker image tags

mregen:


barnstee (approved):

mregen:
as discussed with @karok2m last week...

karok2m (approved):

karok2m:
We have deployed Docker images for this branch to AKS and have validated that correct image tag is set in layered deployments, with `-alpha` suffix.

Issue #831, opened 2020-10-27
Error getting device scope result from IoTHub for OPC-UA endpoint IoT Device (Unauthorized)


deanwyns:
**Describe the bug**
The edge hub is unable to authorize the device against IoT Hub (connected OPC-UA endpoint).

**To Reproduce**
I am not sure how this is reproduced. IIoT gets into a state where the device is unauthorized.
We are using version **2.7.170**.

1. Connect an OPC-UA endpoint (ours has no security policy, so None)
2. Endpoint is Ready and ActivatedAndConnected
3. Publish some nodes
4. No telemetry is received. Check the edgeHub logs and see error.

**Expected behavior**
OPC-UA endpoint device should connect successfully and send telemetry.

**Screenshots**
N/A

**Desktop (please complete the following information):**
Not really applicable but:
 - OS: Mac OS
 - Browser: Firefox
 - Version 81

**Edge Hub logs**
```
<6> 2020-10-26 15:14:24.187 +00:00 [INF] - Unable to authenticate client uat2debc6682f391903f31826bcbaefd2571e234e49 with cached service identity uat2debc6682f391903f31826bcbaefd2571e234e49. Resyncing service identity...
<4> 2020-10-26 15:14:24.685 +00:00 [WRN] - Error while refreshing the service identity for uat2debc6682f391903f31826bcbaefd2571e234e49
Microsoft.Azure.Devices.Edge.Hub.CloudProxy.DeviceScopeApiException: Message: Error getting device scope result from IoTHub, HttpStatusCode: Unauthorized, Content: {"Message":"ErrorCode:IotHubUnauthorizedAccess;Unauthorized","ExceptionMessage":"Tracking ID:9d72d5d2308048c08b7ae7fd31df403b-G:4-TimeStamp:10/26/2020 15:14:24"}
   at Microsoft.Azure.Devices.Edge.Hub.CloudProxy.DeviceScopeApiClient.GetIdentitiesInScope(Uri uri) in /home/vsts/work/1/s/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs:line 131
[...]

PR #863, opened 2020-11-09
Ensure device scope is set on endpoint device twin for endpoint id to…

marcschier:
… be used as leaf device.


marcschier:
#831

cristipogacean (approved):

karok2m (approved):

PR #917, opened 2020-11-19
Add support to reuse test resources

markusstuhler:
Added support to automatically reuse testing resources


koepalex (approved):
required for unblocking e2e testing

PR #921, opened 2020-11-20
Feature/automated testing mstuhler

markusstuhler:


koepalex (approved):

e2e-tests/IIoTPlatform-E2E-Tests/TestConstants.cs
```diff
@@ -139,6 +139,31 @@ internal static class EnvironmentVariablesNames {
             /// IoT Hub connection string
             /// </summary>
             public const string PCS_IOTHUB_CONNSTRING = "PCS_IOTHUB_CONNSTRING";
+
+            /// <summary>
+            /// The connection string of the event-hub compatible endpoint of IoT Hub.
+            /// </summary>
+            public const string IOTHUB_EVENTHUB_CONNECTIONSTRING = "IOTHUB_EVENTHUB_CONNECTIONSTRING";
```
karok2m, line 146: We have `PCS_IOTHUB_EVENTHUBENDPOINT` for the same value. Maybe we can reuse that.

e2e-tests/IIoTPlatform-E2E-Tests/TestConstants.cs
```diff
@@ -139,6 +139,31 @@ internal static class EnvironmentVariablesNames {
...
+            
+            /// <summary>
+            /// The connection string of the storage account that will be used for checkpointing (when monitoring IoT Hub)
+            /// </summary>
+            public const string STORAGE_CONNECTIONSTRING = "STORAGE_CONNECTIONSTRING";
```
karok2m, line 151: Would this be similar to `PCS_STORAGE_CONNSTRING` ? If so, then maybe we can reuse that environment variable.

PR #924, opened 2020-11-23
Added new env variable, removed trailing spaces

dacolgit:


koepalex:
@dacolgit please make this changes together with first standalone e2e test. We need to check that the functionality is working all together.

PR #928, opened 2020-11-24
IAI: Upgraded NuGet packages.

karok2m:


koepalex (dismissed):
General question: should we use the `*-preview` NuGet packages for production?

barnstee:
> General question: should we use the `*-preview` NuGet packages for production?

No we should not, unless there is nothing else?

koepalex (approved):
:shipit:

karok2m:
> > General question: should we use the `*-preview` NuGet packages for production?
> 
> No we should not, unless there is nothing else?

Ideally we shouldn't if there are non-preview versions. But most of `Microsoft.Azure.Management.*` NuGet packages are still in preview and there are no non-preview versions of them. In my last commit I migrated to `Microsoft.Graph` from `Microsoft.Graph.Beta`, which is only possible since recently.

PR #931, opened 2020-11-25
IAI: Added principalType to role assignment.

karok2m:
To fix a known issue with delay of Service Principal creation, we have added  `principalType` field in role assignment properties.

Here are a few relevant links that point to this solution:
* Stack Overflow: [Sometimes ARM template will throw PrincipalNotFound Error when Working with User-assigned Managed Identity](https://stackoverflow.com/questions/60516853/sometimes-arm-template-will-throw-principalnotfound-error-when-working-with-user)
* Microsoft Docs: [Add Azure role assignments using Azure Resource Manager templates](https://docs.microsoft.com/en-us/azure/role-based-access-control/role-assignments-template#new-service-principal)


koepalex (approved):

koepalex:
Do we also have other ARM templates were we need to apply this change?

PR #940, opened 2020-12-01
Feature/automated testing updated app service deployment

markusstuhler:
Prevent AppService from being stopped on idle.


koepalex (approved):
AlwaysOn will really help to avoid errors in long running tests

PR #957, opened 2020-12-07
Added cancellation tokens to async methods.

karok2m:


e2e-tests/IIoTPlatform-E2E-Tests/TestHelper.cs
```diff
@@ -132,23 +148,33 @@ internal static class TestHelper {
...
         /// Switch to publisher orchestrated mode
         /// </summary>
         /// <param name="destinationFilePath">Path of the PublishedNodesFile.json file to be deleted</param>
         /// <param name="context">Shared Context for E2E testing Industrial IoT Platform</param>
-        public static void SwitchToOrchestratedMode(string destinationFilePath, IIoTPlatformTestContext context) {
+        /// <param name="ct">Cancellation token</param>
+        public static async Task SwitchToOrchestratedModeAsync(
+            string destinationFilePath,
+            IIoTPlatformTestContext context,
+            CancellationToken ct = default
```
koepalex, line 176: I would remove all the `= default` if we are going to support CancellationToken than the caller should be forced to decide to use it (or not), meaning the caller has to pass default or a concrete instance. This comment is valid for almost every modified method.

koepalex (approved):
:shipit:

e2e-tests/IIoTPlatform-E2E-Tests/TestHelper.cs
karok2m: I don't agree with that. I will argue that providing the default is the right way to give a choice of using `CancellationToken` or not. If the user of method choses not to use a `CancellationToken`, then they should not be forced into writing additional code if there is a possibility of not doing that.

PR #960, opened 2020-12-08
Remove Vault experimental code and dependencies from 2.7

marcschier:


api/src/Microsoft.Azure.IIoT.Api.Testing/cli/Program.cs
```diff
@@ -122,8 +118,6 @@ public class Program : IDisposable {
             _twin = _scope.Resolve<ITwinServiceApi>();
             _registry = _scope.Resolve<IRegistryServiceApi>();
             _publisher = _scope.Resolve<IPublisherServiceApi>();
-            _vault = _scope.Resolve<IVaultServiceApi>();
-            _jobs = _scope.Resolve<IPublisherJobServiceApi>();
```
barnstee, line 126: Why Publisher job service?

barnstee (approved):

api/src/Microsoft.Azure.IIoT.Api.Testing/cli/Program.cs
marcschier: The field is not used in this program.

PR #970, opened 2020-12-10
Remove vault service and dependencies

marcschier:


barnstee (approved):

deploy/helm/azure-industrial-iot/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: azure-industrial-iot
-version: 0.3.0
+version: 0.3.1
```
karok2m, line 3: We have already released `0.3.1` version of the chart, so let's make this `0.3.2`.

samples/src/Microsoft.Azure.IIoT.App/src/Startup.cs
```diff
@@ -231,8 +230,6 @@ public class Startup {
                 .AsImplementedInterfaces();
             builder.RegisterType<RegistryServiceClient>()
                 .AsImplementedInterfaces();
-            builder.RegisterType<VaultServiceClient>()
```
karok2m, line 234: vault should be removed from comment on line 228/229:

```
 // Register twin, vault, and registry services clients
```

api/src/Microsoft.Azure.IIoT.Api.Testing/cli/Program.cs
```diff
@@ -77,8 +75,6 @@ public class Program : IDisposable {
                 .AsImplementedInterfaces();
             builder.RegisterType<RegistryServiceClient>()
                 .AsImplementedInterfaces();
-            builder.RegisterType<VaultServiceClient>()
```
karok2m, line 80: `vault` should be removed from the comment on line `73`/`75`:

```
// Register twin, vault, and registry services clients
```

deploy/helm/azure-industrial-iot/Chart.yaml
```diff
@@ -15,4 +15,4 @@ maintainers:
 - name: Karapet Kostandyan
   email: kakostan@microsoft.com
 icon: https://raw.githubusercontent.com/Azure/Industrial-IoT/master/docs/media/microsoft-symbol.png
-appVersion: 2.7.105
+appVersion: 2.7.200
```
karok2m, line 18: `2.7.200` should be fine for now, but we will need to update this later on, when we know exact version that we will release.

karok2m: @marcschier, please also change the value of `image:tag` parameter in `deploy/helm/azure-industrial-iot/values.yaml` to `2.7.200`. Those two should have the same values.

karok2m: `deploy/helm/azure-industrial-iot/README.md` should also be updated to mention `2.7.200` instead of `2.7.105`. The value is mentioned there twice.

deploy/helm/azure-industrial-iot/README.md
```diff
@@ -835,7 +833,6 @@ Those are the service ports exposed by components:
 | `deployment.microServices.history.service.port`              | `9043`               |
 | `deployment.microServices.gateway.service.port`              | `9040`               |
 | `deployment.microServices.gateway.service.opcPort`           | `51111`              |
-| `deployment.microServices.vault.service.port`                | `9044`               |
```
karok2m, line 838: Please remove line `797`/`799` which talks about `deployment.microServices.vault.imageRepository` parameter.

deploy/helm/azure-industrial-iot/values.yaml
```diff
@@ -74,7 +74,6 @@ azure:
   # The following CosmosDB containers will be created (<db_name>.<container_name>):
   #   - iiot_opc.iiot_opc-indices
   #   - iiot_opc.iiot_opc-requests
-  #   - OpcVault.AppsAndCertRequests
```
karok2m, line 77: @marcschier, this value is still used in `components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Migration/VaultApplicationMigration.cs`, which is not removed by this PR. The migration job will still create this CosmosDB container. So this value is still relevant.

karok2m (changes requested):
Please see comments above.

deploy/helm/azure-industrial-iot/Chart.yaml
marcschier: Thanks for the thorough review - wasn't sure why it said 105, when we are 199, so assumed some post processing or so :-).

api/src/Microsoft.Azure.IIoT.Api.Testing/cli/Program.cs
marcschier: Ah, the joy of copy paste :-)

barnstee (approved):

karok2m (approved):

PR #976, opened 2020-12-11
Remove Historic access service

marcschier:
* Remove preview history service and business logic related.
* Move the raw historic access API, controller and related models to Twin service to align with twin module API


services/src/Microsoft.Azure.IIoT.Services.Processor.Onboarding/src/Properties/launchSettings.json
```diff
@@ -1,27 +1,7 @@
 {
-  "iisSettings": {
```
karok2m, line 2: @marcschier, is the change to `launchSettings.json` of onboarding service intentional ? Just asking because I do not see other changes to onboarding service.

marcschier: Good question.  This was an opportunistic change. VS generates one when there is none checked in and for quite some time it appears it wasnt.   This meant at some point in the past a generated one was accidently checked in.  I now updated the one to be the same as the rest of the launchsettings.json in the tree, particularly for non-web services.

mregen (commented):
I read in the title 'remove historic access service' but I see some portions of historic access API sneaking into the twin service, or is it even new code? Hard to tell, so this is really moving a feature and an API change for the 2.8 release, right? All fine for me if this API move is intentional and approved.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.csproj
```diff
@@ -3,6 +3,11 @@
     <TargetFramework>netstandard2.1</TargetFramework>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
+  <ItemGroup>
+    <Compile Remove="Utils\**" />
```
cristipogacean, line 7: is this needed?

cristipogacean (dismissed):
:shipit:

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Testing/src/Tests/HistoricalAccess/Extensions/VariantEncoderEx.cs
```diff
@@ -0,0 +1,105 @@
...
+    using Microsoft.Azure.IIoT.OpcUa.Twin.Models;
+    using Microsoft.Azure.IIoT.Serializers;
+    using Opc.Ua;
+    using System;
+    using System.Linq;
+
+    /// <summary>
+    /// Variant encoder extensions to encode and decode details and results
+    /// </summary>
+    internal static class VariantEncoderEx {
```
barnstee, line 16: Please use variant encoder from OPC Foundation stack.

barnstee (changes requested):
Moving historic read back into OPC Twin is OK for me. We know folks like Equinor are looking for this functionality. Please use the OPC UA stack encoder, though...

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Testing/src/Tests/HistoricalAccess/Extensions/VariantEncoderEx.cs
marcschier: No, not in this change, too risky and large change.  Please handle through a separate work item.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.csproj
marcschier: Yep

barnstee (approved):

PR #1008, opened 2021-01-14
Update howto-deploy-modules-portal.md

jehona-m:
Replaced ":latest" with current version number.


koepalex (commented):
I don't like the idea... latest is referencing release, which is currently 2.7.206 why setting it explicitly?

barnstee (approved):

PR #1015, opened 2021-01-18
Add file system watcher for file delete operations.

barnstee:


koepalex (dismissed):

koepalex (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -128,7 +129,21 @@ public class LegacyJobOrchestrator : IJobOrchestrator {
         }
 
         private void _fileSystemWatcher_Changed(object sender, FileSystemEventArgs e) {
-            RefreshJobFromFile();
+            if (e.ChangeType == WatcherChangeTypes.Deleted) {
+                _logger.Information("Published nodes file deleted, cancelling all publishing jobs");
+
+                _lock.Wait();
```
karok2m, line 135: @barnstee, we should either have a try-finally block here for releasing the lock or we can use a [`lock` statement](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/lock-statement). As it is, this code seems to be dangerous and may result in a deadlock.

```
_lock.Wait();
try {
    ...
}
finally {
    _lock.Release();
}
```

karok2m (changes requested):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
barnstee: Fixed.

karok2m (approved):

PR #1016, opened 2021-01-18
Extend matching of subscription notification to published nodes 

koepalex:
**Problem:**
Using DataSetFieldId as identifier for PubSub payload is not working if published node was defined by Expanded Node ID from published nodes json.

**Expected:**
node id and expected node id should work

**Implementation:**
Match subscription notification to "meta description" (DataSetSource) not only be direct compare of NodeId also by compare of expanded Node Id.

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
@@ -436,7 +436,8 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine {
             else { //do the long running lookup as less as possible
                 foreach (var dataSetWriter in message.WriterGroup.DataSetWriters) {
                     foreach (var publishedVariableData in dataSetWriter.DataSet.DataSetSource.PublishedVariables.PublishedData) {
-                        if (publishedVariableData.PublishedVariableNodeId == notification.NodeId &&
+                        if ((publishedVariableData.PublishedVariableNodeId == notification.NodeId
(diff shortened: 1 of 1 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
```diff
@@ -436,7 +436,8 @@ public class NetworkMessageEncoder : IMessageEncoder {
             else { //do the long running lookup as less as possible
                 foreach (var dataSetWriter in message.WriterGroup.DataSetWriters) {
                     foreach (var publishedVariableData in dataSetWriter.DataSet.DataSetSource.PublishedVariables.PublishedData) {
-                        if (publishedVariableData.PublishedVariableNodeId == notification.NodeId &&
+                        if ((publishedVariableData.PublishedVariableNodeId == notification.NodeId
+                              || publishedVariableData.PublishedVariableNodeId.ToExpandedNodeId(context).AsString(context) == notification.NodeId.ToExpandedNodeId(context.NamespaceUris).AsString(context)) &&
```
barnstee, line 440: Please double-check why the second call to ToExpandedNodeId takes the namespaceUris

barnstee (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
koepalex: Parameter ToExpandedNodeId is different if the main object is string or NodeId, it is correct

karok2m (approved):

PR #1017, opened 2021-01-19
published nodes with different "groups/assets" for same endpoint causing closing and reopening of sessions

koepalex:
**Problems:**
* Jobs generated out of published nodes file, all have same id
* GetAvailableJobAsync didn't lock when called from multiple worker
* SendHeartbeatAsync logic with "dirty flag" (_updated) error prone

**Solutions:**
* Use DataGroupId as job id (if available) otherwise include GUID
* Add lock and exception handling to GetAwailableJobAsync
* SendHeartbeatAsync just working on _assignedJobs dictionary

**Tests:**
* added test to verify job ids are unequal

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
@@ -79,8 +79,8 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine {
         /// <param name="ct"></param>
         /// <returns></returns>
         public Task<JobProcessingInstructionModel> GetAvailableJobAsync(string workerId, JobRequestModel request, CancellationToken ct = default) {
+            _lock.Wait(ct);
             try {
-                _lock.Wait(ct);
                 if (_assignedJobs.TryGetValue(workerId, out var job)) {
                     return Task.FromResult(job);
                 }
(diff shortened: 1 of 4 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -79,53 +79,79 @@ public class LegacyJobOrchestrator : IJobOrchestrator {
         /// <param name="ct"></param>
         /// <returns></returns>
         public Task<JobProcessingInstructionModel> GetAvailableJobAsync(string workerId, JobRequestModel request, CancellationToken ct = default) {
-            if (_assignedJobs.TryGetValue(workerId, out var job)) {
+            try {
+                _lock.Wait(ct);
```
karok2m, line 83: Lock should be acquired before the try block. That way, if we fail to acquire it, we will not get to try-catch-finally block and will not release it.

koepalex: good point, updated for GetAvailableJobAsync and SendHearthbeatAsync

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -79,53 +79,79 @@ public class LegacyJobOrchestrator : IJobOrchestrator {
...
                 return Task.FromResult(job);
             }
-            if (_availableJobs.Count > 0 && _availableJobs.TryDequeue(out job)) {
-                _assignedJobs.AddOrUpdate(workerId, job);
-                if (_availableJobs.Count == 0) {
-                    _updated = false;
-                }
+            catch(OperationCanceledException) {
+                JobProcessingInstructionModel job = null;
+                return Task.FromResult(job);
```
karok2m, line 95: Why do we return `null` when there was a cancellation and don't propagate the error up ?

koepalex: propagated OperationCanceledExceptions to caller for both SendHeartbeatAsync and GetAvailableJobAsync

karok2m (approved):

PR #1027, opened 2021-02-03
Enable CA-signed app certs for OPC Publisher standalone mode.

barnstee:


cristipogacean:
Besides the _ApplicationCertificateSubjectName_, the _ApplicationName_ should be provided as well and needs to be the same value as we have in CN field of the _ApplicationCertificateSubjectName_ like in the example below
```
{
    "User": "ContainerAdministrator",
    "Hostname": "publisher",
    "Cmd": [
        "pf=/mount/pn.json",
        "PkiRootPath=/mount/pki",
        "--aa",
        "ApplicationName=TEST-PUBLISHER",
        "ApplicationCertificateSubjectName=CN=TEST-PUBLISHER,OU=Windows2019,OU=\"Test OU\",DC=microsoft,DC=com"
    ],
    "HostConfig": {
        "Mounts": [
            {
                "Type": "bind",
                "Source": "C:\\mount",
                "Target": "C:\\mount"
            }
        ]
    }
}
```

BTW, the above configuration parameters work fine, making this PR a bit redundant.

barnstee:
> Besides the _ApplicationCertificateSubjectName_, the _ApplicationName_ should be provided as well and needs to be the same value as we have in CN field of the _ApplicationCertificateSubjectName_ like in the example below
> 
> ```
> {
>     "User": "ContainerAdministrator",
>     "Hostname": "publisher",
>     "Cmd": [
>         "pf=/mount/pn.json",
>         "PkiRootPath=/mount/pki",
>         "--aa",
>         "ApplicationName=TEST-PUBLISHER",
>         "ApplicationCertificateSubjectName=CN=TEST-PUBLISHER,OU=Windows2019,OU=\"Test OU\",DC=microsoft,DC=com"
>     ],
>     "HostConfig": {
>         "Mounts": [
>             {
>                 "Type": "bind",
>                 "Source": "C:\\mount",
>                 "Target": "C:\\mount"
>             }
>         ]
>     }
> }
> ```
> 
> BTW, the above configuration parameters work fine, making this PR a bit redundant.

Thanks Cristi. Agreed. When you say they work fine, where? in Main?

cristipogacean:
@barnstee, it should work in all 2.7.xxx releases and on master as well. Yes, it works in standalone mode as well. I tested with the above mentioned container start options (standalone) a third party CA-signed cert provided scenario.

barnstee:
> @barnstee, it should work in all 2.7.xxx releases and on master as well. Yes, it works in standalone mode as well. I tested with the above mentioned container start options (standalone) a third party CA-signed cert provided scenario.

Strange. I also tested it and it wasn't working for me (with the subject name from customer).

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliConfigKeys.cs
```diff
@@ -172,6 +172,16 @@ public static class LegacyCliConfigKeys {
         /// </summary>
         public const string OpcOwnCertStorePath = SecurityConfig.ApplicationCertificateStorePathKey;
 
+        /// <summary>
+        /// Key for app cert subject name.
+        /// </summary>
+        public const string OpcApplicationCertificateSubjectName = SecurityConfig.ApplicationCertificateSubjectNameKey;
```
koepalex, line 178: Were is this used?!

barnstee: See below.

cristipogacean:
> > @barnstee, it should work in all 2.7.xxx releases and on master as well. Yes, it works in standalone mode as well. I tested with the above mentioned container start options (standalone) a third party CA-signed cert provided scenario.
> 
> Strange. I also tested it and it wasn't working for me (with the subject name from customer).

Maybe we have a regression, or maybe the ApplicationName was missing or different, or the private key was not provided, or the CA cert missing, etc... The entire cert store is soo error prone, I've been through all of these, the trial/error way. We can sync & share experiences if needed. 

In principle, I think we should try to have one way providing settings (using a single key as argument or env variable) ideally reused for both orchestrated and standalone modes. Furthermore, ideally the same config key will be used in twin & discovery for identical features. The idea with --aa & --pf like arguments was an exception just to ensure backwards compatibility with 2.5 for standalone mode, therefore I'm a bit reluctant on extending these parameters for config keys that already available.

barnstee:
> > > @barnstee, it should work in all 2.7.xxx releases and on master as well. Yes, it works in standalone mode as well. I tested with the above mentioned container start options (standalone) a third party CA-signed cert provided scenario.
> > 
> > 
> > Strange. I also tested it and it wasn't working for me (with the subject name from customer).
> 
> Maybe we have a regression, or maybe the ApplicationName was missing or different, or the private key was not provided, or the CA cert missing, etc... The entire cert store is soo error prone, I've been through all of these, the trial/error way. We can sync & share experiences if needed.
> 
> In principle, I think we should try to have one way providing settings (using a single key as argument or env variable) ideally reused for both orchestrated and standalone modes. Furthermore, ideally the same config key will be used in twin & discovery for identical features. The idea with --aa & --pf like arguments was an exception just to ensure backwards compatibility with 2.5 for standalone mode, therefore I'm a bit reluctant on extending these parameters for config keys that already available.

No regression. I got it working by setting these tow settings as environment variables on the OPC Publisher container module. They just don't work as command line arguments, but I think environment variables are a good enough workaround for now. *thank you* Cristi for your help!

karok2m:
@barnstee, is this fix applied on the `master` branch ?

barnstee:
> @barnstee, is this fix applied on the `master` branch ?

No, I haven't gotten an approval yet.

koepalex (approved):
:shipit:

PR #1029, opened 2021-02-10
Changing repository name for new ACR

jyotig-ms:


tools/scripts/acr-copy-release.ps1
```diff
@@ -122,9 +122,10 @@ Get-ChildItem $BuildRoot -Recurse -Include "container.json" | ForEach-Object {
         $targetImageArgs = @()
         foreach ($targetTag in $imageTagsMap.Item($sourceTag)) {
             $targetImageArgs += "--image"
-            $targetImageArgs += "$($imageName):$($targetTag)"
+            $targetImageArgs += "public/$($imageName):$($targetTag)"
```
karok2m, line 125: Should `public` namespace in source and destination container registries be an optional parameter ?

karok2m (approved):

tools/scripts/acr-copy-release.ps1
jyotig-ms: As of now, all our MCR images are public, so it should be fine here. This is a requirement from MCR to prefix public images like this. If we want to add private images on MCR in future, yes this will need to change.

Issue #1045, opened 2021-03-01
Empty catch blocks in aad-register.ps1 


WilliamBerryiii:
**Describe the bug**
There are several empty powershell `catch` blocks in aad-register.ps1 which causes excess troubleshooting friction when registering Azure Industrial IoT applications. 

**To Reproduce**
See lines 227 and 249 ... when control flow hits these blocks you are not notified of the error condition, script failure, and must begin a manual search for where issues are occurring. 

**Expected behavior**
Errors that crop up in AAD application registration are traceable without code modification.

PR #1046, opened 2021-03-01
Add error message to empty catch blocks in aad-register.ps1. 

WilliamBerryiii:
There are empty catch blocks that result in script failure, without reporting where the failure occurred, when falling through user credential collection in app registration. 
This closes issue #1045.


deploy/scripts/aad-register.ps1
```diff
@@ -224,6 +224,7 @@ Function New-ADApplications() {
                     -TenantId $tenantId `
                 }
             catch {
+                Write-Host "An exception was encountered collecting user credentials when registering $($applicationDisplayName): $($_.Exception.Message)"
```
luiscantero, line 227: Let's avoid passive voice when possible and simplify, e.g.: "Failed collecting user credentials while registering $($applicationDisplayName): $($_.Exception.Message)"

deploy/scripts/aad-register.ps1
```diff
@@ -246,6 +247,7 @@ Function New-ADApplications() {
                     $user = (Get-AzureADUser -SearchString $creds.Account.Id)[0]
                 }
                 catch {
+                    Write-Host "Getting user principal for $($creds.Account.Id) failed when searching by the acount ID."
```
luiscantero, line 250: Fix typo "acount" and avoid passive voice, e.g. "Failed getting user principal for $($creds.Account.Id) while searching by account ID"

luiscantero (changes requested):

deploy/scripts/aad-register.ps1
WilliamBerryiii: Thanks and great callout!  Will update.

deploy/scripts/aad-register.ps1
WilliamBerryiii: Thanks and great callout!  Will update.

WilliamBerryiii:
Addressed code review comments, updated branch with latest change sets and pushed. Thanks for the review!

luiscantero (approved):

PR #1047, opened 2021-03-02
Update GetSecrets.ps script to write key vault secrets to launchSettings.json

MaidaB:


tamaskleizer (approved):

tools/e2etesting/GetSecrets.ps1
```diff
@@ -1,8 +1,20 @@
-param($KeyVaultName)
+param(
+	[Parameter(Mandatory=$true)]
```
karok2m, line 2: @MaidaB, can you please use 4 spaces for indentation instead of tab character ? The indentation of this file is inconsistent at the moment. I can see tabs being used on lines `47`, `48` and `51` as well.

tools/e2etesting/GetSecrets.ps1
```diff
@@ -16,4 +28,25 @@ foreach ($secret in $secrets) {
 	$values[$secret.Name.ToUpperInvariant().Replace("-", "_")] = $secretValueText
 }
 
-$values | ConvertTo-Json -Depth 10
\ No newline at end of file
+# Find the full path of the launchSettings.json file
+$folderName = "e2e-tests"
+$currentPath = (Get-Location).Path
+$path = Get-ChildItem -Path $currentPath $folderName -Recurse
```
karok2m, line 34: Question: why do you need `-Recurse` here while on line `37` it is not present when performing the same search ?

tools/e2etesting/GetSecrets.ps1
```diff
@@ -16,4 +28,25 @@ foreach ($secret in $secrets) {
...
\ No newline at end of file
+# Find the full path of the launchSettings.json file
+$folderName = "e2e-tests"
+$currentPath = (Get-Location).Path
+$path = Get-ChildItem -Path $currentPath $folderName -Recurse 
+while (!$path) {	
+    $currentPath = $currentPath + '\..'
+    $path = Get-ChildItem -Path $currentPath $folderName 
+}
+$settingsFile = $path.FullName + "\IIoTPlatform-E2E-Tests\Properties\launchSettings.json"
```
karok2m, line 39: @MaidaB, would you please consider making the script cross-platform so that it can run on Linux/WSL as well ? For that you would only need to use PowerShell functions such as `Join-Path` instead of using Windows path separator. Line `36` would also need to change slightly. Please feel free to ignore this comment if this is not a priority.

tools/e2etesting/GetSecrets.ps1
MaidaB: fixed

karok2m (approved):

tools/e2etesting/GetSecrets.ps1
MaidaB: First we perform the forward search. For example if you started the script from C:\ it needs to go to the repository and check what's inside, so it would be C:\Industrial-IoT\e2e-tests.
If that doesn't work we can assume that we are already inside of the repository and we have to go back. For example if I start the script from C:\Industrial-IoT\tools\e2etesting it will go two folders back and find e2e-tests there.

tools/e2etesting/GetSecrets.ps1
MaidaB: I tested it on a Linux machine and it works fine when PowerShell is installed

Issue #1051, opened 2021-03-02
Feature: Add early check to see if a website with the given application name already exists. 


WilliamBerryiii:
The Azure Industrial IoT deployment is a lengthy process and it would be helpful to have a global unique name check earlier in the deployment script execution.  This check would shorten a failed deployment (due to a non-unique website name) by 5-10 minutes and allow for faster iteration and feedback. Even with a check in place, it would still be possible for a given application name to be registered between the check and the website deployment; but that scenario is far less likely than the name having already been taken. 

I'm making a PR with a minimal change to `deploy.ps1` to add this check.

WilliamBerryiii:
Per comment in associated PR. I'm closing this issue for further evaluation.

PR #1052, opened 2021-03-03
Add early check to deploy.ps1 for global website name conflict. 

WilliamBerryiii:
This PR is an offer to address feature request/issue #1051 by doing Azure Website global name check on the application name and `throws` if there is a conflict.

### DIFF
--- a/deploy/scripts/deploy.ps1
+++ b/deploy/scripts/deploy.ps1
@@ -698,6 +698,9 @@ Function New-Deployment() {
                 $script:applicationName = $script:resourceGroupName
             }
         }
+        if (Test-AzureName -Website $script:applicationName -eq $true) {
+            throw "Conflict: Website with given name $script:applicationName already exists."
+        }
         if (($script:type -eq "all") -or ($script:type -eq "app")) {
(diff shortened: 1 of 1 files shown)


marcschier (approved):

marcschier (changes requested):
Thinking about this, what happens if the deployment is an update of an existing deployment?

WilliamBerryiii:
@marcschier you raise a very valid point which I had thought could be worked around with an additional resource group tag check.  I am however realizing now that I had a local powershell profile that was providing the credentials needed to run `Test-AzureName`, which wouldn't necessarily be configured for folks running the deployment script. Further, I didn't realize that the PS cmdlets used such a wide array of profile definitions: https://stackoverflow.com/a/38676505 - yikes!

I'm going to keep working through this scenario as it's raised a "update" deployment issue where the Azure Workbook name already exists and the deployment fails; but I want to get a better handle on what's happening there before I file a follow-up issue.

I'm going to Close this PR for now, and if I can make progress on a better solution, I'll re-submit.  Thanks for you time.

PR #1066, opened 2021-03-17
refactor test preparation and cleanup

tamaskleizer:
* introduce twin collection and context
* allow TestOutputHelper to be null in TestHelper
* move preparation and cleanup steps to context


e2e-tests/IIoTPlatform-E2E-Tests/Twin/TwinTestContext.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TwinTestContext"/> class.
+        /// Used for preparation executed once before any tests of the collection are started.
+        /// </summary>
+        public TwinTestContext() : base() {
+            PrepareTestEnvironment();
+        }
+
+        /// <summary>
+        /// Disposes of resources.
```
MaidaB, line 24: Disposes resources

e2e-tests/IIoTPlatform-E2E-Tests/Twin/TwinTestContext.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// Used for cleanup executed once after all tests of the collection were executed.
+        /// </summary>
+        protected override void Dispose(bool disposing) {
+            if (!disposing) {
+                return;
+            }
+
+            OutputHelper = null;
```
MaidaB, line 32: Maybe add a comment why this is needed

MaidaB (dismissed):

e2e-tests/IIoTPlatform-E2E-Tests/Twin/TwinTestContext.cs
```diff
@@ -0,0 +1,101 @@
...
+    using Xunit;
+
+    public class TwinTestContext : IIoTPlatformTestContext {
+        
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TwinTestContext"/> class.
+        /// Used for preparation executed once before any tests of the collection are started.
+        /// </summary>
+        public TwinTestContext() : base() {
+            PrepareTestEnvironment();
```
karok2m, line 20: I would suggest to move call to `PrepareTestEnvironment();` to the constructor of test suite, where `TwinTestContext` will be injected. As it is at the moment, the constructor of `TwinTestContext` does too much work which will also take a considerable amount of time to finish.

tamaskleizer: Having it here ensures that this is executed only once before the test collection is started. This refactoring was needed because there will be more classes included in this test collection, so it will decrease overall test execution time (by doing only one preparation and cleanup).

karok2m: Wouldn't it also be executed once if you call it from the constructor of `TwinBrowseTestTheory` ?

karok2m: One other aspect of the current setup that bothers me is that during debugging session you would have to wait for all this code in constructor of `TwinTestContext` to execute before the execution reaches constructor of `TwinBrowseTestTheory`. That can result is a delay of several minutes before code execution reaches your test suite, which can be very surprising for somebody that is not familiar with this code.

MaidaB: Moving the setup to the TwinBrowseTestTheory class will increase the execution time because we will have at least two more classes for the twin tests (TwinCallTestTheory, TwinReadNodeTestTheory) so in my opinion this is better. Also having one setup for the whole collection is a common practice in xunit and shouldn't be a big surprise for a developer.

tamaskleizer: The constructor of the test class is not executed once before it's test methods, but before each test method call. In that case you would have the ~1 min preparation before each test. You can solve this with a class fixture, or a context of a testcollection. You won't be able to follow these in the test explorer during debugging, which is a bit annoying, but overall test execution time should decrease dramatically.

karok2m (approved):

e2e-tests/IIoTPlatform-E2E-Tests/Twin/TwinTestContext.cs
karok2m: OK, got it. I kind of forgot about the suite being instantiated once per each test case. And I didn't know that the same fixture object will be shared among the test suites as well, good to know. Thanks for clarification.

PR #1079, opened 2021-03-23
Refactoring to base monitored item model

peterbryntesson:
fix: refactored MonitoredItemModel into BaseMonitoredItemModel and Data/EventMonitoredItemModel


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Models/EventMonitoredItemModel.cs
```diff
@@ -0,0 +1,59 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+namespace Microsoft.Azure.IIoT.OpcUa.Protocol.Models {
+    using Microsoft.Azure.IIoT.OpcUa.Core.Models;
+
+    /// <summary>
+    /// Data monitored item
```
simonjaeger, line 10: Event

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Extensions/PublishedDataSetEventModelEx.cs
```diff
@@ -34,15 +34,13 @@ public static class PublishedDataSetEventsModelEx {
...
                 DiscardNew = publishedEvent.DiscardNew,
                 QueueSize = publishedEvent.QueueSize,
                 TriggerId = publishedEvent.TriggerId,
                 MonitoringMode = publishedEvent.MonitoringMode,
                 StartNodeId = publishedEvent.EventNotifier,
                 RelativePath = publishedEvent.BrowsePath,
                 AttributeId = null,
-                DataChangeFilter = null,
                 IndexRange = null,
                 SamplingInterval = null
```
simonjaeger, line 45: Looks like if this is set to null, the GetValueOrDefault methods used on this property will default to 1 second. Whereas for events a client should always set the sampling interval to 0. To enforce this, maybe this is an appropriate place?
[](https://reference.opcfoundation.org/v104/Core/docs/Part4/5.12.1/)

Similar consideration for any other properties? E.g., queue size of 0?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -570,8 +570,9 @@ internal sealed class SubscriptionWrapper : ISubscription {
                 var revisedKeepAliveCount = _subscription.Configuration.KeepAliveCount
                     .GetValueOrDefault(session.DefaultSubscription.KeepAliveCount);
                 _subscription.MonitoredItems?.ForEach(m => {
-                    if (m.HeartbeatInterval != null && m.HeartbeatInterval != TimeSpan.Zero) {
-                        var itemKeepAliveCount = (uint)m.HeartbeatInterval.Value.TotalMilliseconds /
+                    var dataModel = m as DataMonitoredItemModel;
+                    if (dataModel != null && dataModel.HeartbeatInterval != null && dataModel.HeartbeatInterval != TimeSpan.Zero) {
```
simonjaeger, line 574: Just a detail, but you can combine the cast and check in the if statement:

```cs
if (m is DataMonitoredItemModel dataModel && ...
```

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -942,10 +956,10 @@ private class MonitoredItemWrapper {
             /// </summary>
             /// <param name="template"></param>
             /// <param name="logger"></param>
-            public MonitoredItemWrapper(MonitoredItemModel template, ILogger logger) {
+            public MonitoredItemWrapper(BaseMonitoredItemModel template, ILogger logger) {
```
simonjaeger, line 959: Can we add (or track) tests for this method? So we can be confident that the filters are right.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Models/EventMonitoredItemModel.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <summary>
+        /// Event filter
+        /// </summary>
+        public EventFilterModel EventFilter { get; set; }
+
+        /// <summary>
+        /// Clone
+        /// </summary>
+        /// <returns></returns>
+        public override BaseMonitoredItemModel Clone() {
```
simonjaeger, line 22: Would be good to write tests now that we have changed the Clone() a bit. Can we add (or track)?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesJobConverter.cs
```diff
@@ -74,8 +74,8 @@ public class PublishedNodesJobConverter {
         /// <param name="items"></param>
         /// <param name="legacyCliModel">The legacy command line arguments</param>
         /// <returns></returns>
-        private IEnumerable<WriterGroupJobModel> ToWriterGroupJobs(
-            IEnumerable<PublishedNodesEntryModel> items, LegacyCliModel legacyCliModel) {
+        internal IEnumerable<WriterGroupJobModel> ToWriterGroupJobs(
```
simonjaeger, line 77: Why?

simonjaeger (changes requested):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Extensions/PublishedDataSetEventModelEx.cs
peterbryntesson: Fixed for SamplingInterval. Regarding QueueSize, if that should always be 0 - we shouldn't have it as a setting. Let's find out.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
peterbryntesson: Fixed

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
peterbryntesson: I think we can cover this with unit tests for MonitoredItemModel. Will add in this PR

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Models/EventMonitoredItemModel.cs
peterbryntesson: I will write unit tests for MonitoredItemModel in this PR

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesJobConverter.cs
peterbryntesson: Beats me. fixed.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Models/EventMonitoredItemModel.cs
peterbryntesson: Fixed

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Model/MonitoredItemModelTestHelper.cs
```diff
@@ -0,0 +1,313 @@
+﻿namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Tests.Model {
+    using Microsoft.Azure.IIoT.OpcUa.Core.Models;
+    using Microsoft.Azure.IIoT.OpcUa.Protocol.Models;
+    using System;
+    using System.Collections.Generic;
+
+    internal static class MonitoredItemModelTestHelper {
```
simonjaeger, line 7: Great that you added the test! Thanks! 

I'm not sure why we need this class however. It does the same checks that IsSameAs implemented on the model classes does? Those are sufficient. Shouldn't have to duplicate that code here? It's not used outside the test project.


This class we shouldn't have to have however. Instead, can we call the IsSameAs on the actual objects instead? So that our tests actually verify those and not this helper.

simonjaeger (changes requested):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Models/BaseMonitoredItemModel.cs
```diff
@@ -69,23 +70,51 @@ public class MonitoredItemModel {
...
-        /// </summary>
-        public AggregateFilterModel AggregateFilter { get; set; }
-
-        /// <summary>
-        /// heartbeat interval not present if zero
-        /// </summary>
-        public TimeSpan? HeartbeatInterval { get; set; }
+        /// <param name="other"></param>
+        /// <returns></returns>
+        public virtual bool IsSameAs(BaseMonitoredItemModel other) {
```
koepalex, line 83: Any specific reason not to use .NET Standard IComparable or to overload == and != operator?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Models/BaseMonitoredItemModel.cs
```diff
@@ -69,23 +70,51 @@ public class MonitoredItemModel {
...
-        /// Aggregate filter
-        /// </summary>
-        public AggregateFilterModel AggregateFilter { get; set; }
-
-        /// <summary>
-        /// heartbeat interval not present if zero
-        /// </summary>
-        public TimeSpan? HeartbeatInterval { get; set; }
+        /// <param name="other"></param>
+        /// <returns></returns>
```
koepalex, line 82: Please remove or will out

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesJobConverter.cs
```diff
@@ -14,14 +14,14 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Models {
...
     using System.Collections.Generic;
-    using System.Diagnostics;
     using System.IO;
     using System.Linq;
     using System.Text;
     using System.Threading.Tasks;
     using Microsoft.Azure.IIoT.Exceptions;
     using Microsoft.Azure.IIoT.OpcUa.Publisher.Config.Models.Data;
     using Microsoft.Azure.IIoT.OpcUa.Publisher.Config.Models.Events;
+    using System.Diagnostics;
```
koepalex, line 24: Is this new namespace really used?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Extensions/PublishedDataSetSourceModelEx.cs
```diff
@@ -38,20 +39,26 @@ public static class PublishedDataSetSourceModelEx {
...
-                model.SubscriptionSettings?.PublishingInterval.ToString() +
-                publishedVariableData?.Id +
-                publishedVariableData?.PublishedVariableNodeId +
-                publishedVariableData?.PublishedVariableDisplayName +
-                publishedVariableData?.SamplingInterval +
-                publishedVariableData?.HeartbeatInterval +
-                publishedEventData?.Id +
-                publishedEventData?.EventNotifier;
-            return id.ToSha1Hash();
+            var sb = new StringBuilder();
```
koepalex, line 42: Nice change, I like it :blush:

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Models/BaseMonitoredItemModel.cs
peterbryntesson: The IsSameAs pattern is used *everywhere* else in the code - I moved this code from from the MonitoredItemModelEx extension into the base class. However, I like your suggestion - I was thinking of implementing IComparable but didn't do. I think overriing Equals() is a better alternative. However, all the sub objects (like DataChangeFilterModel etc) does not follow this pattern right now. Do we think that I should change this for all applicable classes? There will be classes that still does IsSameAs - will it be weird to have two different patterns? That's what I'm thinking about anyway

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Model/MonitoredItemModelTestHelper.cs
peterbryntesson: Well, I was thinking about having two parallel ways of comparing the objects. But I see your point, maybe they are overkill.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesJobConverter.cs
peterbryntesson: Yes, StopWatch() is used.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Models/BaseMonitoredItemModel.cs
koepalex: Got it, sounds good I will ask the same questions someone else ;)

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Models/BaseMonitoredItemModel.cs
peterbryntesson: I don't understand what you mean here

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Model/MonitoredItemModelTestHelper.cs
peterbryntesson: Removed the separate IsSameAs in test project

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Models/BaseMonitoredItemModel.cs
peterbryntesson: I just committed a version where we implement Equals in our new classes, but keep IsSameAs everywhere else. Take a look at it. If you want I can revert to IsSameAs in these new classes as well.

peterbryntesson: Looking at the code right now there are 31 implementations of IsSameAs for different classes. So it wouldn't be impossible to refactor them all if we would like to. It does change a lot of code though - so it's a high-risk change.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Models/BaseMonitoredItemModel.cs
peterbryntesson: Now I understand and I have fixed it. Did a quick search and there are 2430 other occurrences of this in the code :-)

simonjaeger (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Models/BaseMonitoredItemModel.cs
koepalex: If you see it as high-risk based on the amount of code that needs to be changed. We should leave like it is, for now. Thanks

koepalex (approved):

PR #1081, opened 2021-03-26
Tool to extract assessments per registry

marcschier:


tools/clean.sh
```diff
@@ -0,0 +1,65 @@
...
+        az group delete -g $group -y
+    fi
+done
+
+# purge deleted keyvault
+#for vault in $(az keyvault list-deleted -o tsv | tr -d '\r'); do
+#    echo "deleting keyvault $vault ..."
+#    az keyvault purge --name $vault
+#done
+
```
hansgschossmann, line 65: i would remove or make it optional

tools/clean.sh
```diff
@@ -0,0 +1,65 @@
...
+    groups=$(az group list --query "[?starts_with(name, '$prefix')].name" -o tsv | tr -d '\r')
+else
+    # remove groups not marked for keeping
+    groups=$(az group list --query "[?tags.DoNotDelete!='true'].name" -o tsv | tr -d '\r')
+fi
+
+# remove groups 
+for group in $groups; do
+    if [[ $group = MC_* ]] ; then
+        echo "skipping $group ..."
```
hansgschossmann, line 51: interesting

hansgschossmann (dismissed):
:shipit:

tools/scripts/acr-scan.ps1
```diff
@@ -0,0 +1,97 @@
...
+    The subscription to use - otherwise uses default.
+
+ .PARAMETER All
+    Include also vulnerabilities that are not patchable.
+#>
+
+Param(
+    [string] $Registry = "industrialiotprod",
+    [string] $Subscription = "IOT_GERMANY",
+    [switch] $All
```
hansgschossmann, line 25: not sure if we should expose this information

tools/scripts/acr-scan.ps1
```diff
@@ -0,0 +1,97 @@
...
+$argumentList = @("acr", "show", "--name", $script:Registry, "-ojson")
+$result = (& "az" $argumentList)
+if ($LastExitCode -ne 0) {
+    throw "az $($argumentList) failed with $($LastExitCode)."
+}
+$acr = $result | ConvertFrom-Json
+
+# Get all vulnerability assessments
+$argumentList = @("security", "sub-assessment", "list", "-ojson",
+    "--assessed-resource-id", $acr.id,
```
hansgschossmann, line 48: "-o json"? json is default anyhow.

hansgschossmann (approved):
:shipit:

tools/scripts/acr-scan.ps1
marcschier: It is in several other scripts already.  We should move all these scripts to another repo eventually.

tools/scripts/acr-scan.ps1
marcschier: You can set another default as azure cli configuration, then json is not anymore.

PR #1089, opened 2021-04-06
Fix FetchOpcNodeDisplayName flag for events

simonjaeger:
When the flag _FetchOpcNodeDisplayName_ (--fd) is set, it has an undesirable impact on the published events. Fixed by filtering the monitored items when the subscription wrapper is asked to resolve display names.

The flag is intended for OpcNodes and not OpcEvents: "... flag whether to fetch the display names of the nodes." (from LegacyCliConfigKeys).


peterbryntesson:
LGTM. Nice catch!

koepalex (approved):
:shipit:

PR #1092, opened 2021-04-08
Encode EncodeableDictionary as dictionary

simonjaeger:
The EncodeableDictionary is only used for encoding event values. It was encoded as an array on a property called EventFields. There are already classes to encode arrays, but it also is not reflective of how the encoder should encode a dictionary. 

The following change makes the encoder produce key-value pairs. One caveat is that decoding is challenging when the keys are not known. However, this is not required for events as far as I understand.


koepalex (approved):
:shipit:

PR #1097, opened 2021-04-13
Add event timestamp fields

simonjaeger:
Added "Time" and "ReceiveTime" as default select clauses (if not already added) so that SourceTimestamp and ServerTimestamp can be set. Corrected a bug in GetFieldDisplayName when trying to access a display name outside the template array.


peterbryntesson:
LGTM

peterbryntesson (approved):
LGTM

mregen (approved):

PR #1103, opened 2021-04-15
Add monitored item wrapper tests

simonjaeger:
Adding tests to ensure that the properties and filters are set correctly from the respective templates; `DataMonitoredItemModel `and `EventMonitoredItemModel`. 

To do this, the access modifier of the `MonitoredItemWrapper` class and its `Create` method had to be changed. The `Session `parameter was removed only to bring the two parameters needed; `MessageContext `and `TypeTree`.

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
@@ -422,7 +422,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Protocol.Services {
                             toAdd.Template.MonitoringMode = Publisher.Models.MonitoringMode.Disabled;
                         }
                         try {
-                            toAdd.Create(rawSubscription.Session, codec, activate);
+                            toAdd.Create(rawSubscription.Session.MessageContext, rawSubscription.Session.NodeCache.TypeTree, codec, activate);
                             if (toAdd.EventTemplate != null) {
                                 toAdd.Item.AttributeId = Attributes.EventNotifier;
                             }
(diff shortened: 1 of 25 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -1022,24 +1022,25 @@ private class MonitoredItemWrapper {
...
             /// Create new stack monitored item
             /// </summary>
-            /// <param name="session"></param>
+            /// <param name="messageContext"></param>
+            /// <param name="typeTree"></param>
             /// <param name="codec"></param>
             /// <param name="activate"></param>
             /// <returns></returns>
-            internal void Create(Session session, IVariantEncoder codec, bool activate) {
+            public void Create(ServiceMessageContext messageContext, ITypeTable typeTree, IVariantEncoder codec, bool activate) {
```
peterbryntesson, line 1030: In the simplified event configuration I need access to the NodeCache, so use NodeCache as the parameter instead

simonjaeger: Good catch! I will fix.

buzzfrog, line 1030: Is it better to name the parameter typeTable to reflect interface name

simonjaeger: Agreed.

simonjaeger: Now resolved.

mregen (approved):

PR #1109, opened 2021-04-17
Use IP address of the test servers instead of fqdn

MaidaB:
Switch to using IP addresses of the test servers so that we can use the same tests for servers that have only private IP (for example for the nested edge validation)


cristipogacean (approved):
:shipit:

PR #1113, opened 2021-04-19
Caching pending alarms

peterbryntesson:
feat: Added code and data structures to cache the latest copy of all condition type derived messages for events that have PendingAlarms.IsEnabled flag set.

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
@@ -741,7 +741,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Protocol.Services {
                         Notifications = new List<MonitoredItemNotificationModel>()
                     };
 
-                    if (notification != null && notification.Events != null) {
+                    if (notification?.Events != null) {
                         for (var i = 0; i < notification.Events.Count; i++) {
                             var monitoredItem = subscription.MonitoredItems.SingleOrDefault(
                                     m => m.ClientHandle == notification.Events[i].ClientHandle);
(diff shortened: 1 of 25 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -735,10 +738,36 @@ internal sealed class SubscriptionWrapper : ISubscription {
                         EndpointUrl = subscription.Session?.Endpoint?.EndpointUrl,
                         SubscriptionId = Id,
                         Timestamp = publishTime,
-                        Notifications = notification.ToMonitoredItemNotifications(
-                                subscription.MonitoredItems)?.ToList()
+                        Notifications = new List<MonitoredItemNotificationModel>()
                     };
 
+                    if (notification != null && notification.Events != null) {
```
simonjaeger, line 744: Could be `notification?.Events != null`

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/Events/PendingAlarmsOptionsModel.cs
```diff
@@ -49,6 +49,18 @@ public class PendingAlarmsOptionsModel {
                 TimeSpan.FromMilliseconds(SnapshotInterval.Value) : (TimeSpan?)null;
         }
 
+        /// <summary>
+        /// ConditionIdIndex
+        /// </summary>
+        [IgnoreDataMember]
+        public int ConditionIdIndex { get; set; } = -1;
```
simonjaeger, line 56: These could be nullable uints. `uint?`. Would be cleaner to check for null than -1.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/Events/PendingAlarmsOptionsModel.cs
```diff
@@ -49,6 +49,18 @@ public class PendingAlarmsOptionsModel {
                 TimeSpan.FromMilliseconds(SnapshotInterval.Value) : (TimeSpan?)null;
         }
 
+        /// <summary>
+        /// ConditionIdIndex
```
simonjaeger, line 53: I know the code looks like this... but can we make these summaries better? A sentence that describes the property.

simonjaeger (changes requested):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/Events/PendingAlarmsOptionsModel.cs
peterbryntesson: Had to make then int?, sinces values[xx] required int. Didn't want to have a cast there

simonjaeger (approved):

simonjaeger:
LGTM

mregen (approved):

PR #1114, opened 2021-04-19
Set endpoint from the settings instead of pn.json

MaidaB:


e2e-tests/IIoTPlatform-E2E-Tests/TestHelper.cs
```diff
@@ -131,12 +131,15 @@ internal static partial class TestHelper {
                             Assert.NotNull(entryModels[0].OpcNodes);
                             Assert.NotEmpty(entryModels[0].OpcNodes);
 
-                            result.Add(url, entryModels[0]);
+                            // Set endpoint url correctly when it's not specified in pn.json
+                            entryModels[0].EndpointUrl = $"opc.tcp://{ipAddress}:50000";
```
cristipogacean, line 135: >entryModels[0].EndpointUrl = $"opc.tcp://{ipAddress}:50000"; [](http://example.com/codeflow?start=28&length=60)

something looks strange here. if the OpcPlcConfig.Urls only contains ip addresses, why is it called "Urls"?

cristipogacean (approved):

PR #1115, opened 2021-04-19
fix NotAuthorized exception on endpoint activate

cristipogacean:
fix for Activate endpoint fails the first time on IoT Edge LTS (1.1.0), succeeds on second. (CONNECT failed: RefusedNotAuthorized)

### DIFF
--- a/e2e-tests/IIoTPlatform-E2E-Tests/TestHelper.Registry.cs
+++ b/e2e-tests/IIoTPlatform-E2E-Tests/TestHelper.Registry.cs
@@ -239,9 +239,18 @@ namespace IIoTPlatform_E2E_Tests {
                 request.AddHeader(TestConstants.HttpHeaderNames.Authorization, accessToken);
                 request.Resource = string.Format(TestConstants.APIRoutes.RegistryActivateEndpointsFormat, endpointId);
 
-                var response = client.ExecuteAsync(request, ct).GetAwaiter().GetResult();
+                // TODO remove workaround
+                // This request used to fail when called for the first time. The bug was fixed in the master,
+                // but we will keep this workaround until we can consume the fixed platform in the E2E tests.
+                bool IsSuccessful = false;
+                int numberOfRetries = 3;
+                while (!IsSuccessful && numberOfRetries > 0) {
+                    var response = await client.ExecuteAsync(request, ct);
+                    IsSuccessful = response.IsSuccessful;
+                    numberOfRetries--;
+                }
 
-                Assert.True(response.IsSuccessful, "POST /registry/v2/endpoints/{endpointId}/activate failed!");
+                Assert.True(IsSuccessful, "POST /registry/v2/endpoints/{endpointId}/activate failed!");
 
                 while (true) {
                     Assert.False(ct.IsCancellationRequested, "Endpoint was not activated within the expected timeout");


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Services/EndpointRegistry.cs
```diff
@@ -726,12 +729,13 @@ public sealed class EndpointRegistry : IEndpointRegistry, IApplicationEndpointRe
         private async Task ActivateAsync(EndpointRegistration registration,
             RegistryOperationContextModel context, CancellationToken ct = default) {
 
+            // Ensure device scope for the registration before getting the secret.
+            // Changing device's scope regenrerates the secret.
```
mregen, line 733: typo :-)

e2e-tests/IIoTPlatform-E2E-Tests/TestHelper.Registry.cs
```diff
@@ -234,17 +234,10 @@ public static class Registry {
                 request.AddHeader(TestConstants.HttpHeaderNames.Authorization, accessToken);
                 request.Resource = string.Format(TestConstants.APIRoutes.RegistryActivateEndpointsFormat, endpointId);
 
-                // This request fails when called for the first time, retry the call while the bug is being investigated
```
mregen, line 237: your fix is for this workaround?

mregen (approved):

MaidaB (dismissed):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Services/EndpointRegistry.cs
cristipogacean: fixed, thanks!

e2e-tests/IIoTPlatform-E2E-Tests/TestHelper.Registry.cs
cristipogacean: exactly. the issue was that the first activate endpoint always fails

MaidaB (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Discovery/src/Services/DiscoveryServices.cs
```diff
@@ -281,7 +281,7 @@ public class DiscoveryServices : IDiscoveryServices, IScannerServices, IDisposab
             _logger.Information("Start {mode} discovery run...", request.Mode);
             var watch = Stopwatch.StartNew();
 
-            if (request.Mode == DiscoveryMode.Url) {
+            if (request.Mode == DiscoveryMode.Url || request.Mode == DiscoveryMode.Off) {
```
marcschier, line 284: Is this why in master discovery did not work anymore?

e2e-tests/IIoTPlatform-E2E-Tests/TestHelper.Registry.cs
```diff
@@ -196,8 +196,7 @@ public static class Registry {
                     string discoveryUrl,
                     CancellationToken ct = default) {
 
-                var applicationId = await GetApplicationIdAsync(context, discoveryUrl, ct);
-
+                var applicationId = context.ApplicationId == null ? await GetApplicationIdAsync(context, discoveryUrl, ct) : context.ApplicationId;
```
marcschier, line 199: Nit: For readability I prefer ternary to not include method calls.  Better a if else here, or first assign applicationId from context, then test for null and retrieve it "async" as "fall back".  A comment would also be helpful.

marcschier (dismissed):
See my comment.

MaidaB (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Discovery/src/Services/DiscoveryServices.cs
cristipogacean: The real fix is the change in the OnboardingClient.cs, and this change is a safeguard.

e2e-tests/IIoTPlatform-E2E-Tests/TestHelper.Registry.cs
cristipogacean: Right, will change that. Ideally the entire cleanup logic needs to be reviewed because is inconsistent.

MaidaB (dismissed):

MaidaB (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Discovery/src/Services/DiscoveryServices.cs
hansgschossmann: Why do we discover when DiscoveryMode is Off?

cristipogacean: It's the case when discovery request has DiscoveryMode.Off set and contains discoveryUrl(s). This was supposed to go targeted on the provided url(s) and get all endpoints, but not do any network scanning. The DiscoveryMode.Url was introduced later and, with this, the semantic of DiscoveryMode.Off was changed (in a pretty dramatic way - Off was actually triggering a full blown network scan that). I just restored the old semantic for backwards compatibility. We can discuss if we want to change the semantic here, but for now I just fixed the bug.

PR #1116, opened 2021-04-20
Moved the call to ConditionRefresh

peterbryntesson:
Call to ConditionRefresh was in a slightly wrong place and not wrapped in a try/catch block, causing intermittent null reference exceptions


simonjaeger (approved):
LGTM

mregen (approved):
:shipit:

PR #1118, opened 2021-04-21
Add reversible encoding to EncodeableDictionary

simonjaeger:
Added reversible encoding to EncodeableDictionary by writing the set of keys. Added tests to the EncodeableDictionary.

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Stack/Encoders/Models/EncodeableDictionary.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Stack/Encoders/Models/EncodeableDictionary.cs
@@ -70,12 +70,16 @@ namespace Opc.Ua.Encoders {
 
                 // Write keys for decoding.
                 encoder.WriteUInt32(kKeysCountIdentifier, (uint)Count);
-                encoder.WriteStringArray(kKeysIdentifier, this.Select(x => x.Key).ToArray());
+                if (Count > 0) {
+                    encoder.WriteStringArray(kKeysIdentifier, this.Select(x => x.Key).ToArray());
+                }
             }
 
-            foreach (var entry in this) {
-                if (!string.IsNullOrEmpty(entry.Key)) {
-                    encoder.WriteDataValue(entry.Key, entry.Value);
+            if (Count > 0) {
+                foreach (var entry in this) {
+                    if (!string.IsNullOrEmpty(entry.Key)) {
+                        encoder.WriteDataValue(entry.Key, entry.Value);
+                    }
                 }
             }
         }
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Stack/Encoders/Models/EncodeableDictionary.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Stack/Encoders/Models/EncodeableDictionary.cs
@@ -13,6 +13,11 @@ namespace Opc.Ua.Encoders {
     /// </summary>
     public class EncodeableDictionary : List<KeyDataValuePair>, IEncodeable {
 
+        /// <summary>
+        /// Identifier for the keys in the reversible format.
+        /// </summary>
+        private const string kKeysIdentifier = "_Keys";
+
         /// <inheritdoc/>
         public ExpandedNodeId TypeId =>
             nameof(EncodeableDictionary);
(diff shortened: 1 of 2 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Stack/Encoders/Models/EncodeableDictionary.cs
```diff
@@ -6,12 +6,18 @@
...
 
     /// <summary>
     /// Encodeable dictionary carrying field names and values
     /// </summary>
     public class EncodeableDictionary : List<KeyDataValuePair>, IEncodeable {
 
+        /// <summary>
+        /// Identifier for the keys in the reversible format.
+        /// </summary>
+        private const string kKeysIdentifier = "_Keys";
```
peterbryntesson, line 19: kKeysIdentifier? Is that the correct format?

simonjaeger: Good question. Actually yes (from .editorconfig):
image

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Stack/Encoders/Models/EncodeableDictionary.cs
```diff
@@ -54,7 +70,23 @@ public class EncodeableDictionary : List<KeyDataValuePair>, IEncodeable {
...
         /// <inheritdoc/>
         public virtual void Decode(IDecoder decoder) {
-            // No operation. Cannot decode without known keys.
+            // Read keys for decoding. 
+            StringCollection keys = null;
+            try {
+                // May throw in some decoders if an empty array was encoded.
+                keys = decoder.ReadStringArray(kKeysIdentifier);
+            }
+            catch {
```
buzzfrog, line 79: Is this a good catch principle?

simonjaeger: No, it is not. But as I tried to put in the comments, because each decoder may throw a custom exception, we either have to enumerate all such exceptions or catch this broadly. For instance, the JsonDecoderEx will throw JsonReaderException, which the others obviously will not. If future decoders are added, these exceptions must either be standardized or put enumerated here. Thoughts @buzzfrog?

buzzfrog: you can't have a catch on a exception that are higher in the hierarchy?
What happens if we get a memory corrupt exception? (unlikely)

buzzfrog: can it even be a security issue? with buffer overrun?

simonjaeger: Good question. I don't think we can guarantee that all the different exceptions will derive from one in a higher chain. I think we will end up at the base. But do you think we should attempt to enumerate all the different exceptions instead?

buzzfrog: Can you not verify that it is not empty before you do decode?

simonjaeger: That is right. I cannot verify that. Unless we write a property for that, which we read first. Such as KeyLength. Thoughts on that approach?

simonjaeger: I have changed my approach to this. If we write the number of keys we don't have to deal with this. What do you think @buzzfrog?

peterbryntesson (approved):
LGTM

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Stack/Encoders/Models/EncodeableDictionary.cs
buzzfrog: LGTM

mregen (dismissed):
:shipit:

mregen (approved):
:shipit:

PR #1121, opened 2021-04-22
Nested edge integration with E2E pipelines

MaidaB:
- nested edge scripts added
- an optional nested edge job added to the pipelines
- DeployPLCs.ps1 needs to use azure cli methods because one of them is not supported by PowerShell cmdlets ( az container create  with a subnet parameter)

Orchestrated pipeline:
- New pipeline variable: NestedEdgeFlag (optional)

Standalone pipeline:
- New pipeline variables: EdgeVersion, NestedEdgeFlag (optional), NestedEdgeSshConnections (optional)
- IIoTEdgeConfig.cs is used to set the edge version, the nested edge flag and ssh connections array
- In the tests the two steps approach is used to copy pn.json file (first to the jumpbox VM and then to the machine in the lower nested edge layer)


cristipogacean (approved):

hansgschossmann (dismissed):
:shipit:

e2e-tests/IIoTPlatform-E2E-Tests/Standalone/A_PublishSingleNodeStandaloneTestTheory.cs
```diff
@@ -47,11 +48,26 @@ public class A_PublishSingleNodeStandaloneTestTheory {
 
         [Fact, PriorityOrder(3)]
         public void Test_StartPublishingSingleNode_Expect_Success() {
-            var cts = new CancellationTokenSource(TestConstants.MaxTestTimeoutMilliseconds);
-
+            var cts = new CancellationTokenSource(TimeSpan.FromMinutes(1));
```
mregen, line 51: >TimeSpan.FromMinutes(1) [](start = 50, length = 23)

why a magic number here when there used to be a configurable time?

e2e-tests/IIoTPlatform-E2E-Tests/Standalone/A_PublishSingleNodeStandaloneTestTheory.cs
```diff
@@ -66,9 +82,10 @@ public class A_PublishSingleNodeStandaloneTestTheory {
                 })
                 .ToArray();
 
+            cts = new CancellationTokenSource(TimeSpan.FromMinutes(5));
```
mregen, line 85: >TimeSpan.FromMinutes(5) [](start = 46, length = 23)

same as above

e2e-tests/IIoTPlatform-E2E-Tests/Standalone/A_PublishSingleNodeStandaloneTestTheory.cs
MaidaB: When running in the nested edge, this call is expected to time out, and waiting 30 minutes is too much, probably I should write it as a comment

e2e-tests/IIoTPlatform-E2E-Tests/Standalone/A_PublishSingleNodeStandaloneTestTheory.cs
MaidaB: this one can be the default one, I will change it

cristipogacean (approved):

Issue #743, opened 2020-09-30
"Failed to add local host address" exception in Discovery service


terrymandin:
I iinstalled the Industrial IIoT Platform v2.5.58.  The Windows gateway Discovery module is not finding any OPC Servers or endpoints - none are visible in the UI.  When I looked at the IoT Edge "Discovery" module I found the following Exception:

```
[2020-09-30 18:51:09Z INF Microsoft.Azure.IIoT.OpcUa.Edge.Discovery.Services.ProgressPublisher] : Starting port scanning (0 probes active)...
[2020-09-30 18:51:10Z INF Microsoft.Azure.IIoT.OpcUa.Edge.Discovery.Services.ProgressPublisher] : Found 0 ports on servers (207 scanned)...
[2020-09-30 18:51:10Z INF Microsoft.Azure.IIoT.OpcUa.Edge.Discovery.Services.DiscoveryServices] Uploading 0 results...
[2020-09-30 18:51:10Z INF Microsoft.Azure.IIoT.OpcUa.Edge.Discovery.Services.DiscoveryServices] 0 results uploaded.
[2020-09-30 18:51:10Z INF Microsoft.Azure.IIoT.OpcUa.Edge.Discovery.Services.ProgressPublisher] : Discovery operation completed.
[2020-09-30 18:55:57Z INF Microsoft.Azure.IIoT.Module.Framework.Hosting.ModuleHost] New settings processed.
[2020-09-30 18:56:16Z INF Microsoft.Azure.IIoT.OpcUa.Edge.Discovery.Services.ProgressPublisher] : Discovery operation started.
[2020-09-30 18:56:16Z INF Microsoft.Azure.IIoT.OpcUa.Edge.Discovery.Services.DiscoveryServices] Start Fast discovery run...
[2020-09-30 18:56:16Z INF Microsoft.Azure.IIoT.OpcUa.Edge.Discovery.Services.ProgressPublisher] : Starting network scan (0 probes active)...
[2020-09-30 18:56:16Z INF Microsoft.Azure.IIoT.OpcUa.Edge.Discovery.Services.DiscoveryServices] Start scanning ["172.31.160.1/32", "172.31.175.0/24 [vEthernet (Ethernet) 3]"]...
[2020-09-30 18:56:16Z INF Microsoft.Azure.IIoT.Module.Framework.Hosting.ModuleHost] New settings processed.
[...]

PR #1125, opened 2021-04-27
Fix discovery exception when resolving the gateway's hostname

cristipogacean:
fix for https://github.com/Azure/Industrial-IoT/issues/743
* discovery uses how gateway hostname environment variable
* log reports refined

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Discovery/src/Services/DiscoveryServices.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Discovery/src/Services/DiscoveryServices.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Discovery.Services {
     using Microsoft.Azure.IIoT.Exceptions;
     using Microsoft.Azure.IIoT.Utils;
     using Microsoft.Azure.IIoT.Serializers;
+    using Microsoft.Azure.IIoT.Abstractions;
     using System;
     using System.Collections.Generic;
     using System.Collections.Concurrent;
@@ -491,7 +492,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Discovery.Services {
                         // Check local host
                         if (host.EqualsIgnoreCase("localhost") && Host.IsContainer) {
                             // Also resolve docker internal since we are in a container
-                            host = kDockerHostName;
+                            host = Environment.GetEnvironmentVariable(IoTEdgeVariables.IOTEDGE_GATEWAYHOSTNAME);
                             continue;
                         }
                         break;
@@ -515,7 +516,8 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Discovery.Services {
             try {
                 if (Host.IsContainer) {
                     // Resolve docker host since we are running in a container
-                    var entry = await Dns.GetHostEntryAsync(kDockerHostName);
+                    string HostName = Environment.GetEnvironmentVariable(IoTEdgeVariables.IOTEDGE_GATEWAYHOSTNAME);
+                    var entry = await Dns.GetHostEntryAsync(HostName);
                     foreach (var address in entry.AddressList
                                 .Where(a => a.AddressFamily == AddressFamily.InterNetwork)
                                 .Where(a => !addresses.Any(b => a.Equals(b)))) {
@@ -647,7 +649,6 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Discovery.Services {
 
         /// <summary> Progress reporting every 3 seconds </summary>
         private static readonly TimeSpan kProgressInterval = TimeSpan.FromSeconds(3);
-        private const string kDockerHostName = "host.docker.internal";
 
         private readonly ILogger _logger;
         private readonly IJsonSerializer _serializer;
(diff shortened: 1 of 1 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Discovery/src/Services/DiscoveryServices.cs
```diff
@@ -512,14 +518,20 @@ public class DiscoveryServices : IDiscoveryServices, IScannerServices, IDisposab
...
+                        _logger.Information("Gateway Host Name not set");
+                        return;
+                    }
+                    _logger.Debug("Resolve IP for Gateway Host Name: {address}", hostName);
+                    var entry = await Dns.GetHostEntryAsync(hostName);
                     foreach (var address in entry.AddressList
                                 .Where(a => a.AddressFamily == AddressFamily.InterNetwork)
                                 .Where(a => !addresses.Any(b => a.Equals(b)))) {
-                        _logger.Information("Including host address {address}", address);
+                        _logger.Information("Include gateway host address {address}", address);
```
luiscantero, line 534: I'm wondering if "Including" would be a better word. With "Include" I'm not sure what is happening.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Discovery/src/Services/DiscoveryServices.cs
```diff
@@ -512,14 +518,20 @@ public class DiscoveryServices : IDiscoveryServices, IScannerServices, IDisposab
         /// <returns></returns>
         private async Task AddLoopbackAddressesAsync(List<IPAddress> addresses) {
             // Check local host
+            string hostName = Environment.GetEnvironmentVariable(IoTEdgeVariables.IOTEDGE_GATEWAYHOSTNAME);
             try {
                 if (Host.IsContainer) {
                     // Resolve docker host since we are running in a container
-                    var entry = await Dns.GetHostEntryAsync(kDockerHostName);
+                    if (string.IsNullOrEmpty(hostName)) {
+                        _logger.Information("Gateway Host Name not set");
```
luiscantero, line 526: We could change to "host name" to have consistent casing

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Discovery/src/Services/DiscoveryServices.cs
```diff
@@ -512,14 +518,20 @@ public class DiscoveryServices : IDiscoveryServices, IScannerServices, IDisposab
...
+                        return;
+                    }
+                    _logger.Debug("Resolve IP for Gateway Host Name: {address}", hostName);
```
luiscantero, line 529: "gateway host name" for consistent casing

luiscantero (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Discovery/src/Services/DiscoveryServices.cs
cristipogacean: thanks! done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Discovery/src/Services/DiscoveryServices.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Discovery/src/Services/DiscoveryServices.cs
cristipogacean: done

luiscantero (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Discovery/src/Services/DiscoveryServices.cs
```diff
@@ -491,13 +492,18 @@ public class DiscoveryServices : IDiscoveryServices, IScannerServices, IDisposab
...
                             continue;
                         }
                         break;
                     }
-                    catch (Exception e) {
-                        _logger.Warning(e, "Failed to resolve the host for {discoveryUrl}", discoveryUrl);
+                    catch (SocketException se) {
+                        _logger.Warning("Failed to resolve the host for {discoveryUrl} due to {message}",
+                            discoveryUrl, se.Message);
+                        return list;
```
jschweda, line 503: shouldn't this catch be in the loop and we should continue to resolve other adresses?

cristipogacean: I find this construction rather weird... Actually we only have a single "host" string as argument, so the while is superflous. We want though to handle the special case when this argument is "localhost". "Localhost" in the module context makes no sense but we want to target the discovery call to the iotedge host system itself. I would immediately get rid of the "while" and handle "localhost" redirection before resolving the host. But, I am not a fan of on-the-fly off-context refactorings so I'll leave it be,  for now :)

MaidaB (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Discovery/src/Services/DiscoveryServices.cs
jschweda: thanks for the clarification

PR #1126, opened 2021-04-27
Hides hidden fields from json

peterbryntesson:
fix; hidden fields now doesn't show up in pubsub json for events
fix: added check so that namespace index isn't out-of-bounds

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
@@ -1111,12 +1111,18 @@ namespace Microsoft.Azure.IIoT.OpcUa.Protocol.Services {
                         }
                     }
 
+                    List<SimpleAttributeOperand> addedSelectClauses = new List<SimpleAttributeOperand>();
+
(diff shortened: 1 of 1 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -1111,12 +1111,18 @@ public class MonitoredItemWrapper {
                         }
                     }
 
+                    List<SimpleAttributeOperand> addedSelectClauses = new List<SimpleAttributeOperand>();
```
simonjaeger, line 1114: Could you add a comment about this so it's clear what we are doing? And perhaps call it internalSelectClauses.

simonjaeger (approved):
LGTM

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
peterbryntesson: Good point. Fixed.

mregen (approved):
:shipit:

PR #1142, opened 2021-05-04
Prevent empty key-value pairs in EncodeableDictionary

simonjaeger:
Do not write key-value pairs with an empty key. Added test to check as well. Generalized the `KeyCount` parameter as well to `Count`. More interpretable in the output.

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Stack/Encoders/Models/EncodeableDictionary.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Stack/Encoders/Models/EncodeableDictionary.cs
@@ -4,7 +4,6 @@
 // ------------------------------------------------------------
 
 namespace Opc.Ua.Encoders {
-    using Newtonsoft.Json;
     using Opc.Ua;
     using System.Collections.Generic;
     using System.Linq;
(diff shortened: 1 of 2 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Stack/Encoders/Models/EncodeableDictionary.cs
```diff
@@ -20,9 +19,9 @@ public class EncodeableDictionary : List<KeyDataValuePair>, IEncodeable {
         private const string kKeysIdentifier = "_Keys";
 
         /// <summary>
-        /// Identifier for the number of keys in the reversible format.
+        /// Identifier for the number of key-value pairs in the reversible format.
         /// </summary>
-        private const string kKeysCountIdentifier = "_KeysCount";
+        private const string kCountIdentifier = "_Count";
```
peterbryntesson, line 24: Why the rename from kKeyCountIdentifier to kCountIdentifier?

simonjaeger: To generalize it more. It can be used by a decoder to know how many key-value pairs there are. It's the same number as for keys. But seemed a bit more general than having for instance a KeysCount and ValuesCount, when that number is always identical. Does that make sense @peterbryntesson ?

peterbryntesson: Yes,that makes sense. Current code is still counting keys though.

simonjaeger: Yes, internally we only use it for that. I made it from an "Output" perspective. So it is more general in the output. I should have clarified that.

peterbryntesson (approved):
LGTM

mregen (approved):

Issue #1137, opened 2021-05-01
Autofac.Core.DependencyResolutionException at /history/v2/read/<endpointId>/values


pablolaurin:
**Describe the bug**
_Autofac.Core.DependencyResolutionException_ is thrown when posting to _/history/v2/read/{endpointId}/values_.

**To Reproduce**
1. Deploy the latest version:

    `\deploy -type local -version latest`

2. Run microservices locally as described at [Debug Azure Industrial IoT locally](https://github.com/Azure/Industrial-IoT/blob/master/docs/deploy/howto-run-microservices-locally.md)

3. Activate an endpoint as described at [Use the REST API using Postman](https://azure.github.io/Industrial-IoT/tutorials/tut-use-postman.html)

4. Try to read history using Postman (post to _/history/v2/read/{endpointId}/values_ for an active endpoint)

5. See the following error returned as the response body:

`{
  "ClassName": "Autofac.Core.DependencyResolutionException",
  "Message": "An exception was thrown while activating Microsoft.Azure.IIoT.OpcUa.History.Clients.HistoricAccessAdapter`1[[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].",
  "Data": {
    "ActivatorChain": "Microsoft.Azure.IIoT.OpcUa.History.Clients.HistoricAccessAdapter`1[[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]"
  },
  "InnerException": {
    "ClassName": "Autofac.Core.DependencyResolutionException",
    "Message": "None of the constructors found with 'Autofac.Core.Activators.Reflection.DefaultConstructorFinder' on type 'Microsoft.Azure.IIoT.OpcUa.History.Clients.HistoricAccessAdapter`1[System.String]' can be invoked with the available services and parameters:\r\nCannot resolve parameter 'Microsoft.Azure.IIoT.OpcUa.Protocol.IVariantEncoderFactory codec' of constructor 'Void .ctor(Microsoft.Azure.IIoT.OpcUa.History.IHistoricAccessServices`1[System.String], Microsoft.Azure.IIoT.OpcUa.Protocol.IVariantEncoderFactory)'.",
    "Data": {},
    "InnerException": null,
    "HelpURL": null,
[...]

PR #1152, opened 2021-05-06
Adding missing variant encoder factory dependency.

marcschier:
The unit tests were passing because the dependency was added as a test dependency.  It should be the dependency of the overall service.  Addresses #1137 as per already provided suggestion in the bug.


karok2m (approved):

karok2m:
@marcschier, do we need this fix on `master` as well ?

mregen (approved):

marcschier:
> @marcschier, do we need this fix on `master` as well ?

No, in master this service was deleted.

Issue #1005, opened 2021-01-13
Reversible encoding support: OPC UA datatypes cannot be decoded with the pubsub json format


Haeavar:
**Describe the bug**
If we subscribe e.g. the _ServerStatus_ node (i=2256) of an OPC UA server we can't decode its value using the _JsonDecoderEx_ implementation.

`                    var decoder = new JsonDecoderEx(input.Body.AsStream());
                    while (decoder.ReadEncodeable(null, typeof(NetworkMessage)) is NetworkMessage message)
...
`

The value of the corresponding payload entry for this node is _NULL_.

**To Reproduce**
1. Subscribe the node _"Id": "nsu=http://opcfoundation.org/UA/;i=2256"_ of an OPC UA server via publishednodes.json
2. Create an azure function with an EventHubTrigger for the incoming events
3. Deode the event data using the _JsonEncoderEx_
4. See the value of the "nsu=http://opcfoundation.org/UA/;i=2256" payload is NULL

**Expected behavior**
I would expect, any other value than NULL. 

Please advice, how to work with the OPC UA PubSub json formatted event data of the publisher in an appropriate way.

barnstee:
I can confirm that this is fixed in the latest version of OPC Publisher currently under development and produces the following JSON pyload for node ID 2256 (ServerStatus):
"ServerStatus": {
  "Value": {
    "Type": 22,
    "Body": {
      "TypeId": {
        "Id": 862
      },
      "Body": {
        "StartTime": "2021-01-15T11:30:18.4093264Z",
        "CurrentTime": "2021-01-15T11:30:24.1766754Z",
        "State": 0,
        "BuildInfo": {
          "ProductUri": "https://github.com/Azure/iot-edge-opc-plc.git",
          "ManufacturerName": "Microsoft",
          "ProductName": "IoTEdge OPC UA PLC",
          "SoftwareVersion": "1.4.363.107+a837774969718713af74ec66d8890e6fc131ddf1",
          "BuildNumber": "1.4.363.107",
          "BuildDate": "2020-10-09T13:23:30Z"
        },
        "SecondsTillShutdown": 0
      }
    }
  },
  "SourceTimestamp": "2021-01-15T11:30:24.1766754Z",
  "ServerTimestamp": "2021-01-15T11:30:24.1767558Z"
}

Until this version is released (sometime this spring), as a workaround, please subscribe to leaf nodes, i.e. the nodes with ID 2257, 2258, etc.

[...]

PR #1154, opened 2021-05-06
Add reversible encoding flag

simonjaeger:
Allowing for `UseReversibleEncoding` on encoders to be set with a flag `--re|--reversibleencoding`.

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
@@ -45,7 +45,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine {
         public Task<IEnumerable<NetworkMessageModel>> EncodeAsync(
             IEnumerable<DataSetMessageModel> messages, int maxMessageSize, bool useReversibleEncoding) {
             try {
-                var resultJson = EncodeAsJson(messages, maxMessageSize);
+                var resultJson = EncodeAsJson(messages, maxMessageSize, useReversibleEncoding);
                 var resultUadp = EncodeAsUadp(messages, maxMessageSize);
                 var result = resultJson.Concat(resultUadp);
                 return Task.FromResult(result);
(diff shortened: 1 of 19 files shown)


peterbryntesson (approved):
Mostly parameter passing LGTM

marcschier (approved):

mregen:
Whats the use case?

simonjaeger:
@mregen 

1. The ability to get the Locale of LocalizedText objects. A customer needs to know the language of error messages of the machines (English or German). For example, the Message property of an Event is only encoded with the Locale when UseReversibleEncoding is set.
1. To be able to use the JsonDecoderEx to process the output of the OPC Publisher, related to https://github.com/Azure/Industrial-IoT/issues/1005

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/IMessageEncoder.cs
```diff
@@ -43,17 +43,19 @@ public interface IMessageEncoder {
         /// </summary>
         /// <param name="message"></param>
         /// <param name="maxMessageSize"></param>
+        /// <param name="useReversibleEncoding"></param>
```
koepalex, line 46: Please don't follow bad practices, rather remove the <param /> comment elments than adding a new empty one

simonjaeger: Agreed! I have fixed it. Removing comments causes build error. Filled out all params instead.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
```diff
@@ -100,7 +101,7 @@ public class NetworkMessageEncoder : IMessageEncoder {
                     var helperEncoder = new JsonEncoderEx(helperWriter, encodingContext) {
                         UseAdvancedEncoding = true,
                         UseUriEncoding = true,
-                        UseReversibleEncoding = false
+                        UseReversibleEncoding = useReversibleEncoding
```
mregen, line 104: should always be true

mregen:
According to the standard, NetworkMessage is always reversible, only DataValue per configuration.. need to double check how this is implemented in main

mregen:
we need to have a quick discussion on our side how to address this across all releases

koepalex:
As discussed with @mregen 

* Current implementation is slightly wrong
  * NetworkMessage and DataSetMessage need to be reversible encoded but this will only affect `StatusCode` and StatusCode is omitted  if it is *OK*
  * `DataSetFieldContentMask` is set to _DataValue_ which enforces an *non-reversible*  encoding

Ideal solution would be to configure `DataSetfieldContentMask` as _Variant_  and the encoder should distinguish based on it but encode NetworkMessage and DataSetMessage always as reversible (independent from payload).

Suggested solution: to set behavior as global setting, seems to be ok, because DataSetFieldContentMask isn't configurable at the moment.

koepalex (approved):

mregen (approved):
🥇

PR #1155, opened 2021-05-06
merge cert store fix (#780)

monikavar:
- https://github.com/Azure/Industrial-IoT/pull/780 Cert Store fix


marcschier (changes requested):
Could you split these in 2 PRs?

marcschier (approved):

PR #1156, opened 2021-05-07
merge NullReferenceException fix (#819) 

monikavar:
- #819 Avoid NullReferenceException in case of job continuation


marcschier (approved):

marcschier (changes requested):
We seem to be missing a large change that looks as if it was fixing several issues.   We need to wait for Alex and discuss with him what we should port of it:  https://github.com/Azure/Industrial-IoT/pull/751

marcschier (approved):

PR #1159, opened 2021-05-07
Fix discovery exception when resolving the gateway's hostname (#1125)

marcschier:
* discovery using gateway hostname
* add logs for discovery gateway host


cristipogacean (approved):
:shipit:

PR #1168, opened 2021-05-10
Fix startup crashes in Application Insights

marcschier:
* Undo update to bad ApplicationInsights packages (https://github.com/microsoft/ApplicationInsights-dotnet/issues/2187)
* Fix null reference exception
* Add IoT SDK diagnostics message.

### DIFF
--- a/common/src/Microsoft.Azure.IIoT.Diagnostics.AppInsights/src/Default/ApplicationInsightsTelemetryInitializer.cs
+++ b/common/src/Microsoft.Azure.IIoT.Diagnostics.AppInsights/src/Default/ApplicationInsightsTelemetryInitializer.cs
@@ -18,7 +18,7 @@ namespace Microsoft.Azure.IIoT.Diagnostics.AppInsights.Default {
         /// Constructor
         /// </summary>
         /// <param name="processIdentity"></param>
-        public ApplicationInsightsTelemetryInitializer(IProcessIdentity processIdentity) {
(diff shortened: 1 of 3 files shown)


common/src/Microsoft.Azure.IIoT.Diagnostics.AppInsights/src/Default/ApplicationInsightsTelemetryInitializer.cs
```diff
@@ -18,15 +18,15 @@ public class ApplicationInsightsTelemetryInitializer : ITelemetryInitializer {
         /// Constructor
         /// </summary>
         /// <param name="processIdentity"></param>
-        public ApplicationInsightsTelemetryInitializer(IProcessIdentity processIdentity) {
+        public ApplicationInsightsTelemetryInitializer(IProcessIdentity processIdentity = null) {
```
karok2m, line 21: @marcschier, do we need a default constructor with `null` as `processIdentity` ? I can see only one initialization of `ApplicationInsightsTelemetryInitializer` (in `AppInsightsContainerBuilderEx` class) and that one even checks that `processIdentity` is not `null`.

marcschier: It is also injected and then process identity might be missing.

karok2m (approved):

PR #1170, opened 2021-05-11
Handle RefreshRequiredEventType

peterbryntesson:
fix: handling refresh required event by calling ConditionRefresh() on the subscription

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
@@ -1478,7 +1478,6 @@ namespace Microsoft.Azure.IIoT.OpcUa.Protocol.Services {
                             _logger.Information("ConditionRefresh for subscription " +
                                 "{subscription} failed with an exception '{message}'", Item.Subscription.DisplayName, e.Message);
                             noErrorFound = false;
-
                         }
                         if (noErrorFound) {
                             _logger.Information("ConditionRefresh for subscription " +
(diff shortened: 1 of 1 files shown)


peterbryntesson:
> The ConditionRefresh logic you added is replicated, right? Can we extract it to a separate method and reuse that?

Maybe. We could put the code in SubscriptionWrapper - but here in the code we only have access to the raw subscription. I decided to just duplicate the code would be better - but I can change it.

simonjaeger (approved):
LGTM

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -1460,6 +1460,31 @@ public class MonitoredItemWrapper {
                         _pendingAlarmsSnapshotTimer.Start();
                         return;
                     }
+                    else if (eventType == ObjectTypeIds.RefreshRequiredEventType) {
+                        var noErrorFound = true;
+
+                        // issue a condition refresh to make sure we are in a correct state
+                        _logger.Information("Now issuing ConditionRefresh for subscription " +
```
mregen, line 1467: has there been an ask to support conditionrefresh2 as well?

mregen (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
peterbryntesson: No, not that I have heard about.

PR #1171, opened 2021-05-11
Fix null reference exception in OnSubscriptionStateChanged

simonjaeger:
When a subscription is registered there is a potential timing issue between when _currentlyMonitored is set and OnSubscriptionStateChanged is called. A null-check fixes this.


buzzfrog:
LGTM

buzzfrog (approved):

PR #1174, opened 2021-05-12
Update documentation to latest from master and update and fix links and versions

marcschier:


deploy/helm/azure-industrial-iot/values.yaml
```diff
@@ -72,9 +70,6 @@ azure:
 
   # Details of Azure CosmosDB that microservices will use as database.
   # The following CosmosDB containers will be created (<db_name>.<container_name>):
```
mregen, line 72: ># The follo [](http://example.com/codeflow?start=2&length=11)

does it mean no cosmosdb containers left and it is now obsolete?

docs/release-announcement.md
```diff
@@ -0,0 +1,16 @@
+# Release announcement
+ 
+## Azure Industrial IoT Platform 2.7.20x
```
mregen, line 3: ># Azure Industrial IoT Platform 2.7.20x  [](http://example.com/codeflow?start=1&length=40)

does the text still make sense?

deploy/helm/azure-industrial-iot/values.yaml
marcschier: No, only job database is remaining, but I think this is not the place to document it, case in point this is why this was not updated/captured.

docs/deploy/howto-add-aks-to-ps1.md
```diff
@@ -487,36 +487,51 @@ deployment:
...
+> above. That will result in `2.8` version of microservices and edge modules to be deployed. If you want
+> to deploy a different version of the platform, please specify it as the value of `image:tag` parameter.
+
+#### Installing `azure-industrial-iot` Helm chart
 
-Here we will install `azure-industrial-iot` Helm chart from the root of the GitHub repo. For that, please
-use `aiiot.yaml` that you created above and run the following command:
+##### Installing the chart from GitHub repository
+
+Here we will install `azure-industrial-iot` Helm chart from the root of the GitHub repo. This deployment
```
mregen, line 498: > we [](http://example.com/codeflow?start=4&length=3)

technical language could be improved, aka use third person

docs/release-announcement.md
marcschier: No, correct, this should change upon release of 2.8 and look forward to 2.9.  However, it would be a separate check in as the LTS text is still being worked on as part of another work item.

koepalex (approved):
:shipit:

PR #1181, opened 2021-05-21
Enable CI/CD for main branch

prmathur-microsoft:
This PR does the following
1. Enables Continuous Integration, Continuous Deployment by enabling
   following
   - PR Build Completion Triggers Orchestrated Pipeline E2E test (Aka CD follows CI)
   - Adds Code Coverage to PR Pipeline as required step (Adds PR Gate)
   - Uses Key vault to retrieve secrets like username, password by default if unspecified for CD
   - Uses default branch "main" for pointing to a 'ImageNameSpace' if unspecified for CD
   - Uses default version i.e. last successfully built version from
     last commit on the main branch to run end to end test if
     unspecified (CI build default)
   - PR builds are generated in 'Industrailiotdev' container registry by default and 'main' is the default ImageNameSpace for 
   building (CI/PR gates)
   - Run scheduled nightly builds on both CI Pipeline (PR Pipeline) and CD Pipeline (Orchestrated E2E Pipeline)
   - Run scheduled nightly on Standalone Pipeline 
2. Adds '-RC' (Release Candidate) to version to indicate a release candidate on main branch
3. Generates unique builds for every different run using Git Height i.e. number of commits post the last version bump.

Post this PR the following practices are recommended to be followed 
1. Main represents the release candidate for future releases (both regular and LTS)
2. Main should only contain code targeted to be released in the order in which it is to be committed to avoid cherry-picking.
3. Even though end to end testing is not gated per PR, upon failure of orchestrated pipeline, team should stop further check-ins to investigate the root cause of instability and only then proceed forward.
4. If there is need to build a feature which is not yet planned to be released a 'preview' or 'alpha' branch should be spawned off until it is ready to be released. In such a case, version should appropriately reflect '-alpha' or '-preview' instead of '-rc'. It is advised to enable CI\CD on preview branches as well.
5. Code development should happen in individual developer branches. Recommendation for such a branch name is 'feature\alias\anybranchname' to clearly indicate the code in development.

Co-authored-by: Karapet Kostandyan <kakostan@microsoft.com>

### DIFF
--- a/tools/e2etesting/ReplaceVariablesInAppSettings.ps1
+++ b/tools/e2etesting/ReplaceVariablesInAppSettings.ps1
@@ -70,7 +70,7 @@ if (!$ImageTag) {
 }
 
 if (!$ImageNamespace) {
-   $ImageNamespace="public"
+   $ImageNamespace="main"
 }
 
 if (!$ContainerRegistryServer) {
--- a/tools/e2etesting/pipeline_orchestrated.yml
+++ b/tools/e2etesting/pipeline_orchestrated.yml
@@ -1,7 +1,26 @@
 name: $(Date:yyyyMMdd)$(Rev:rr)
-
 trigger: none
 pr: none
+
+schedules:
+- cron: "0 10 * * *"
+  displayName: Nightly build of main branch.
+  branches:
+    include:
+    - main
+  always: true
+
+# Trigger Orchestrated E2E Pipeline post completion of build pipeline.
+resources:
+  pipelines:
+  - pipeline: Orchestrated-E2E-CD-Pipeline
+    source: Custom\Azure_IOT\Industrial\Components\Azure.Industrial-IoT
+    trigger:
+      branches:
+        include:
+        - refs/heads/releases/*
+        - refs/heads/master
+        - refs/heads/main
 
 pool:
   name: '$(AgentPool)'
--- a/tools/e2etesting/pipeline_standalone.yml
+++ b/tools/e2etesting/pipeline_standalone.yml
@@ -3,6 +3,25 @@ name: $(Date:yyyyMMdd)$(Rev:rr)
 trigger: none
 pr: none
 
+schedules:
+- cron: "0 10 * * *"
+  displayName: Nightly build of main branch.
+  branches:
+    include:
+    - main
+  always: true
+
+# Trigger Stand Alone E2E Pipeline post completion of build pipeline.
+resources:
+  pipelines:
+  - pipeline: Standalone-E2E-CD-Pipeline
+    source: Custom\Azure_IOT\Industrial\Components\Azure.Industrial-IoT
+    trigger:
+      branches:
+        include:
+        - refs/heads/releases/*
+        - refs/heads/master
+        - refs/heads/main
 pool:
   name: '$(AgentPool)'
 
--- a/tools/e2etesting/steps/deployplatform.yml
+++ b/tools/e2etesting/steps/deployplatform.yml
@@ -17,6 +17,34 @@ steps:
     artifact: 'iai'
     path: $(BasePath)
 
+- task: AzureKeyVault@1
+  displayName: 'Load secrets from Key Vault as variables by default'
+  condition: and(eq(variables['ContainerRegistryServer'], ''), eq(variables['ContainerRegistryUsername'], ''), eq(variables['ContainerRegistryPassword'], ''))
+  inputs:
+    azureSubscription: 'IOT-OPC-WALLS-SP'
+    KeyVaultName: 'developPipelineKV'
+    SecretsFilter: 'ContainerRegistryPassword,ContainerRegistryServer,ContainerRegistryUsername'
+    RunAsPreJob: true
+
+- task: PowerShell@2
+  displayName: Versioning
+  name: setVersionInfo
+  inputs:
+    targetType: filePath
+    filePath: $(BasePath)\..\scripts\set-version.ps1
+
+- task: AzureCLI@2
+  displayName: 'Override platform version if necessary'
+  condition: eq(variables['PlatformVersion'], '')
+  inputs:
+    azureSubscription: '$(AzureSubscription)'
+    azurePowerShellVersion: 'latestVersion'
+    scriptLocation: 'InlineScript'
+    scriptType: 'ps'
+    addSpnToEnvironment: true
+    inlineScript: |
+      Write-Host "##vso[task.setvariable variable=PlatformVersion]$(setVersionInfo.Version_Prefix)$(setVersionInfo.Version_Prerelease)"
+
 - task: AzurePowerShell@5
   displayName: "Replace parameters in appSettings.json (for IAI)"
   inputs:
--- a/tools/scripts/acr-build.ps1
+++ b/tools/scripts/acr-build.ps1
@@ -79,7 +79,7 @@ else {
         # dev feature builds
         $namespace = $namespace.Replace("feature/", "")
     }
-    elseif ($namespace.StartsWith("release/") -or ($namespace -eq "main") -or ($namespace -eq "master")) {
+    elseif ($namespace.StartsWith("release/") -or ($namespace -eq "master")) {
         $namespace = "public"
         if ([string]::IsNullOrEmpty($Registry)) {
             # Release and Preview builds go into staging
@@ -93,7 +93,7 @@ else {
 if ([string]::IsNullOrEmpty($Registry)) {
     $Registry = $env.BUILD_REGISTRY
     if ([string]::IsNullOrEmpty($Registry)) {
-        # Feature builds by default build into dev registry
+        # Feature builds by default into dev registry
         $Registry = "industrialiotdev"
     }
 }
@@ -102,10 +102,12 @@ Write-Warning "Using $($Registry).azurecr.io."
 $latestTag = "latest"
 # get and set build information from gitversion, git or version content
 $sourceTag = $env:Version_Prefix
+$prereleaseTag = $env:Version_Prerelease
 if ([string]::IsNullOrEmpty($sourceTag)) {
     try {
         $version = & (Join-Path $PSScriptRoot "get-version.ps1")
         $sourceTag = $version.Prefix
+        $prereleaseTag = $version.Prerelease
     }
     catch {
         # build as latest if not building from ci/cd pipeline
@@ -116,19 +118,24 @@ if ([string]::IsNullOrEmpty($sourceTag)) {
         $sourceTag = "latest"
     }
 }
-
-# Get build root - this is the top most folder with .dockerignore
-$buildRoot = & $getroot -startDir $Path -fileName ".dockerignore"
-# Get meta data
-$metadata = Get-Content -Raw -Path (join-path $Path "container.json") `
-    | ConvertFrom-Json
-
-# Set image name and namespace in acr based on branch and source tag
-$imageName = $metadata.name
-if ($script:Fast.IsPresent) {
-    if (!$metadata.buildAlways) {
-        Write-Host "Using fast build - Skipping optional image $imageName."
-        return
+if (![string]::IsNullOrEmpty($sourceTag)) {
+    Write-Host "Using version $($sourceTag)$($prereleaseTag) from get-version.ps1"
+}
+else {
+    # Otherwise look at git tag
+    if (![string]::IsNullOrEmpty($env:BUILD_SOURCEVERSION)) {
+        # Try get current tag
+        try {
+            $argumentList = @("tag", "--points-at", $env:BUILD_SOURCEVERSION)
+            $sourceTag = (& "git" $argumentList 2>&1 | ForEach-Object { "$_" });
+            if ($LastExitCode -ne 0) {
+                throw "git $($argumentList) failed with $($LastExitCode)."
+            }
+        }
+        catch {
+            Write-Error "Error reading tag from $($env:BUILD_SOURCEVERSION)"
+            $sourceTag = $null
+        }
     }
 }
 $tagPostfix = ""
@@ -183,8 +190,27 @@ $user = $credentials.username
 $password = $credentials.passwords[0].value
 Write-Debug "Using User name $($user) and passsword ****"
 
-$fullImageName = "$($script:Registry).azurecr.io/$($namespace)$($imageName):$($tagPrefix)$($sourceTag)$($tagPostfix)"
-Write-Host "Building image $($fullImageName)"
+# Get build root - this is the top most folder with .dockerignore
+$buildRoot = & $getroot -startDir $Path -fileName ".dockerignore"
+# Get meta data
+$metadata = Get-Content -Raw -Path (join-path $Path "container.json") `
+| ConvertFrom-Json
+
+
+# Set image name and namespace in acr based on branch and source tag
+$imageName = $metadata.name
+
+$tagPostfix = ""
+$tagPrefix = ""
+if ($Debug.IsPresent) {
+    $tagPostfix = "-debug"
+}
+if (![string]::IsNullOrEmpty($metadata.tag)) {
+    $tagPrefix = "$($metadata.tag)-"
+}
+
+$fullImageName = "$($Registry).azurecr.io/$($namespace)$($imageName):$($tagPrefix)$($sourceTag)$($prereleaseTag)$($tagPostfix)"
+Write-Host "Full image name: $($fullImageName)"
 
 $manifest = @" 
 image: $($fullImageName)
@@ -225,7 +251,7 @@ $definitions | ForEach-Object {
         }
     }
 
-    $image = "$($namespace)$($imageName):$($tagPrefix)$($sourceTag)-$($platformTag)$($tagPostfix)"
+    $image = "$($namespace)$($imageName):$($tagPrefix)$($sourceTag)$($prereleaseTag)-$($platformTag)$($tagPostfix)"
     Write-Host "Start build job for $($image)"
 
     # Create acr command line 
--- a/tools/templates/acrbuild.yml
+++ b/tools/templates/acrbuild.yml
@@ -45,3 +45,4 @@ jobs:
       azureSubscription: azureiiot
       scriptLocation: inlineScript
       inlineScript: powershell ./tools/scripts/acr-build.ps1 $(dockerFolder)
+      
--- a/version.json
+++ b/version.json
@@ -1,7 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "2.8.0",
-  "versionHeightOffset": 0,
+  "version": "2.8.0-rc.{height}",
   "publicReleaseRefSpec": [
     "^refs/heads/main$",
     "^refs/heads/release/\\d+\\.\\d+\\.\\d+"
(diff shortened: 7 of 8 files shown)


tools/templates/sdl.yml
```diff
@@ -14,13 +14,6 @@ jobs:
     DOTNET_CLI_TELEMETRY_OPTOUT: true
     DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
   steps:
-  - task: UseDotNet@2
```
prmathur-microsoft, line 17: Is this needed?

tools/templates/ci.yml
```diff
@@ -32,13 +32,6 @@ jobs:
     DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
     runCodesignValidationInjection: false
   steps:
-  - task: UseDotNet@2
```
prmathur-microsoft, line 35: Is this needed?

tools/templates/acrbuild.yml
```diff
@@ -45,3 +45,10 @@ jobs:
       azureSubscription: azureiiot
       scriptLocation: inlineScript
       inlineScript: powershell ./tools/scripts/acr-build.ps1 $(dockerFolder)
+  - task: AzureCLI@1
```
prmathur-microsoft, line 48: Master had this, Should this be removed?

tools/scripts/acr-build.ps1
```diff
@@ -183,8 +194,27 @@ $user = $credentials.username
 $password = $credentials.passwords[0].value
 Write-Debug "Using User name $($user) and passsword ****"
 
-$fullImageName = "$($script:Registry).azurecr.io/$($namespace)$($imageName):$($tagPrefix)$($sourceTag)$($tagPostfix)"
-Write-Host "Building image $($fullImageName)"
+# Get build root - this is the top most folder with .dockerignore
```
prmathur-microsoft, line 197: main and master diverge here.
$prereleaseTag is a good indicator of what is being built in the latest commit. So I prefer to leave it here. We can discuss how to take this forward, if you prefer to remove it.

tools/e2etesting/steps/variables.yml
```diff
@@ -1,6 +1,6 @@
 variables:
   BasePath: $(System.DefaultWorkingDirectory)\tools\e2etesting
-  ClientCredentialsKeyVaultName: automatedtesting
+  ClientCredentialsKeyVaultName: developPipelineKV
```
prmathur-microsoft, line 3: I'm not 100% confident of the use of this. Let me know and I can remove it.

deploy/src/Microsoft.Azure.IIoT.Deployment/Infrastructure/AksMgmtClient.cs
```diff
@@ -31,7 +31,7 @@ class AksMgmtClient : IDisposable {
         public const string NETWORK_PROFILE_DNS_SERVICE_IP = "10.0.0.10";
         public const string NETWORK_PROFILE_DOCKER_BRIDGE_CIDR = "172.17.0.1/16";
 
-        public const string KUBERNETES_VERSION_FALLBACK = "1.18.14";
```
karok2m, line 34: Let's keep `1.18.14` version here.

tools/docgen.cmd
```diff
@@ -55,7 +55,6 @@ set convert=docker run --rm --mount type=bind,source=%cd%,target=/opt swagger2ma
 call :generate_doc_for_service twin
 call :generate_doc_for_service publisher
 call :generate_doc_for_service registry
-call :generate_doc_for_service history
```
karok2m, line 58: Let's keep this line as `2.8` will still have the history service.

tools/e2etesting/pipeline_orchestrated.yml
```diff
@@ -1,8 +1,27 @@
 name: $(Date:yyyyMMdd)$(Rev:rr)
-
 trigger: none
 pr: none
 
+schedules:
```
karok2m, line 5: Let's add scheduling to `pipeline_standalone.yml` as well so that we have nightly runs of both orchestrated and standalone pipeline.

tools/e2etesting/steps/deployplatform.yml
```diff
@@ -17,6 +17,45 @@ steps:
     artifact: 'iai'
     path: $(BasePath)
 
+- task: AzureKeyVault@1
```
karok2m, line 20: Let's keep in mind that we need to be able to consume variables from pipeline UI as well and this particular step overrides those variable in all cases, even when there were values provided.

tools/e2etesting/steps/variables.yml
karok2m: I agree that we should remove it.

tools/sdkgen.cmd
```diff
@@ -47,7 +47,6 @@ pushd generated
 call :generate_sdk_for_service twin
 call :generate_sdk_for_service publisher
 call :generate_sdk_for_service registry
-call :generate_sdk_for_service history
```
karok2m, line 50: Let's keep history.

tools/swagger.cmd
```diff
@@ -62,7 +62,6 @@ pushd %build_root%\api\swagger
 call :retrieve_spec twin
 call :retrieve_spec publisher
 call :retrieve_spec registry
-call :retrieve_spec history
```
karok2m, line 65: Let's keep history.

tools/e2etesting/steps/deployplatform.yml
```diff
@@ -17,6 +17,45 @@ steps:
...
+  inputs:
+    azureSubscription: '$(AzureSubscription)'
+    azurePowerShellVersion: 'latestVersion'
+    scriptLocation: 'InlineScript'
+    scriptType: 'ps'
+    addSpnToEnvironment: true
+    inlineScript: |
+      Write-Host "##vso[task.setvariable variable=PlatformVersion]$(setVersionInfo.Version_Prefix)$(setVersionInfo.Version_Prerelease)"
+
+- task: AzureCLI@2
```
karok2m, line 47: This would be tied to the image namespace that is set in `acr-build.ps1` which uses `public` namespace when building images from `main`/`master`. So those changes should be synchronized.

tools/scripts/acr-build.ps1
prmathur-microsoft: Conclusion : Since main branch indicates the release candidate, pre-release tag for main should be RC.
In case the need arises where a feature is to be built but cannot be released as yet, and needs to be shared with partner, 'preview' branch should be spun off which cn be used for that purpose with a pre release tag like 'alpha' or 'beta' or 'preview'. 
Hence main should only contains commits which are approved to be released in the cadence like monthly or bi-monthly or LTS.
Nightly runs will determine the stability of main branch i.e. a release candidate.

version.json
```diff
@@ -1,7 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "2.8.0",
-  "versionHeightOffset": 0,
+  "version": "2.8.0-rc.{height}",
```
prmathur-microsoft, line 3: every new commit will build a unique build which will then be tested by orchestrated pipeline.

tools/scripts/acr-build.ps1
```diff
@@ -79,7 +79,12 @@ else {
         # dev feature builds
         $namespace = $namespace.Replace("feature/", "")
     }
-    elseif ($namespace.StartsWith("release/") -or ($namespace -eq "main") -or ($namespace -eq "master")) {
+    elseif (($namespace -eq "main")) {
```
prmathur-microsoft, line 82: Decision was to leave the 'main' namespace as default and build resides in industrialiotdev container registry.

tools/scripts/acr-build.ps1
```diff
@@ -116,19 +128,24 @@ if ([string]::IsNullOrEmpty($sourceTag)) {
         $sourceTag = "latest"
     }
 }
-
-# Get build root - this is the top most folder with .dockerignore
```
prmathur-microsoft, line 120: Is this safe to remove?

tools/templates/acrbuild.yml
karok2m: No, let's keep it. The task makes sure that we have debug images.

tools/templates/sdl.yml
karok2m: I think so. I would keep it.

tools/templates/ci.yml
karok2m: I think so. I would keep it.

tools/templates/ci.yml
```diff
@@ -87,13 +80,6 @@ jobs:
   pool:
     vmImage: $(poolImage)
   steps:
-  - task: UseDotNet@2
```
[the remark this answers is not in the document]
karok2m: I think so. I would keep it.

tools/e2etesting/pipeline_standalone.yml
```diff
@@ -3,6 +3,14 @@ name: $(Date:yyyyMMdd)$(Rev:rr)
 trigger: none
 pr: none
 
+schedules:
```
karok2m, line 6: Do we need to have a dependency on `Azure.Industrial-IoT` pipeline here similar to `pipeline_orchestrated.yml` ? If no, then I would argue that `pipeline_orchestrated.yml` pipeline should not have it either. There should be a symmetry between them in this regard.

tools/scripts/acr-build.ps1
karok2m: I would say so as build and tests are OK.

karok2m (dismissed):

tools/templates/acrbuild.yml
```diff
@@ -45,3 +45,4 @@ jobs:
       azureSubscription: azureiiot
       scriptLocation: inlineScript
       inlineScript: powershell ./tools/scripts/acr-build.ps1 $(dockerFolder)
+
```
prmathur-microsoft, line 48: Did not migrate debug image building due to this change in main -
https://github.com/Azure/Industrial-IoT/commit/9750d55b33f35baf69c9705ef297b12bcce7cb60#diff-7a7e9ac2cd79106fd60115e06f721f070a7b731595080d874deb13a96a93fc37

karok2m (dismissed):

tools/scripts/acr-build.ps1
```diff
@@ -102,10 +108,17 @@ Write-Warning "Using $($Registry).azurecr.io."
 $latestTag = "latest"
 # get and set build information from gitversion, git or version content
 $sourceTag = $env:Version_Prefix
+$prereleaseTag = $env:Version_Prerelease
+if ([string]::IsNullOrEmpty($prereleaseTag))
+{
+    # This can be used on a preview branch *not* targetted for release and maintainence.
+    $prereleaseTag = "-alpha"
```
marcschier, line 115: If never used let's not add it.

tools/scripts/acr-build.ps1
```diff
@@ -79,7 +79,13 @@ else {
...
         $namespace = $namespace.Replace("feature/", "")
     }
-    elseif ($namespace.StartsWith("release/") -or ($namespace -eq "main") -or ($namespace -eq "master")) {
+    elseif (($namespace -eq "main")) {
+        if ([string]::IsNullOrEmpty($Registry)) {
+            # Builds from main should be built nightly and per PR in a container registry to allow testing.
+            $Registry = "industrialiotdev"
+        }
+    }
+    elseif ($namespace.StartsWith("release/") -or ($namespace -eq "master")) {
```
marcschier, line 88: Main is same as master.  You should be able to remove the test for master and main from the if statement and get the same results since the else is the same as the newly added code

tools/e2etesting/ReplaceVariablesInAppSettings.ps1
```diff
@@ -70,7 +70,7 @@ if (!$ImageTag) {
 }
 
 if (!$ImageNamespace) {
-   $ImageNamespace="public"
+   $ImageNamespace="main"
```
marcschier, line 73: Will release still publish to public?  Better remove tests for master and main branch from ifs and only let release branches build to public.  Although not sure why we cannot build release builds from main.

azure-pipelines.yml
```diff
@@ -10,7 +10,7 @@ trigger:
     include:
     - main
     - dev/*
-    - release/*
+    - release/*
```
marcschier, line 13: Space needed?

azure-pipelines.yml
```diff
@@ -46,7 +46,6 @@ stages:
   - template: tools/templates/acrbuild.yml
 - stage: coverage
   displayName: 'Code Coverage'
-  condition: and(succeeded(), ne(variables['Build.Reason'], 'PullRequest'))
```
marcschier, line 49: Why do pull requests need code coverage, we already do tests...  Pull requests are left out to keep the runs as fast as possible, given some flakiness already in the tests.

tools/e2etesting/pipeline_orchestrated.yml
```diff
@@ -1,8 +1,27 @@
...
+  branches:
+    include:
+    - main
+  always: true
+
+# Trigger Orchestrated E2E Pipeline post completion of build pipeline.
+resources:
+  pipelines:
+  - pipeline: E2E Orchestrated run after CI build
+    source: Custom\Azure_IOT\Industrial\Components\Azure.Industrial-IoT
```
marcschier, line 17: Should we have a variable for this and configure it to match our setup?

tools/e2etesting/pipeline_standalone.yml
```diff
@@ -3,6 +3,25 @@ name: $(Date:yyyyMMdd)$(Rev:rr)
...
+  branches:
+    include:
+    - main
+  always: true
+
+# Trigger Orchestrated E2E Pipeline post completion of build pipeline.
+resources:
+  pipelines:
+  - pipeline: E2E Orchestrated run after CI build
+    source: Custom\Azure_IOT\Industrial\Components\Azure.Industrial-IoT
```
marcschier, line 18: Same as above.

tools/scripts/acr-build.ps1
```diff
@@ -116,19 +129,24 @@ if ([string]::IsNullOrEmpty($sourceTag)) {
...
-$metadata = Get-Content -Raw -Path (join-path $Path "container.json") `
-    | ConvertFrom-Json
-
-# Set image name and namespace in acr based on branch and source tag
-$imageName = $metadata.name
-if ($script:Fast.IsPresent) {
-    if (!$metadata.buildAlways) {
-        Write-Host "Using fast build - Skipping optional image $imageName."
-        return
+if (![string]::IsNullOrEmpty($sourceTag)) {
```
marcschier, line 132: I cannot see why this has been changed, can you explain?

marcschier (changes requested):
Would like to learn more re rationale and related proposed build and test flows resulting from these changes.

azure-pipelines.yml
prmathur-microsoft: PR should ideally gate on required code coverage to be checked in. This guarantees minimum testing needed for the code be checked in.

tools/e2etesting/ReplaceVariablesInAppSettings.ps1
prmathur-microsoft: Release pipeline changes are still in process, idea here is that all the default builds should go to Industrailiotdev on its specified branch name. Release builds can go to public.

prmathur-microsoft: This is specially needed for staging nightly runs.

tools/e2etesting/pipeline_orchestrated.yml
prmathur-microsoft: Can you elaborate more? Are you suggesting to use variable to define source ?

tools/scripts/acr-build.ps1
prmathur-microsoft: I wanted to leave 'master' related changes untouched. Happy to do it next PR.

tools/scripts/acr-build.ps1
prmathur-microsoft: I left it as a placeholder for future. It is not needed for for this PR. Like the comment says, its if/when you spawn a preview branch.

tools/scripts/acr-build.ps1
prmathur-microsoft: Idea here is to use $prereleaseTag as -rc for builds running on 'main'. Since main should always indicate the code to be released aka it is next release candidate.

azure-pipelines.yml
prmathur-microsoft: Not needed. I can remove.

tools/e2etesting/ReplaceVariablesInAppSettings.ps1
prmathur-microsoft: Discussed this offline with Marc. I plan to publish a spec detailing this workflow.

tools/scripts/acr-build.ps1
marcschier: Let's remove it then and add it when actually needed, and then properly test it at that point.

tools/scripts/acr-build.ps1
marcschier: It is not related to master.  My point was to remove the test for both and you end up with the same result as what the added code provides (the default path uses the branch name as namespace name and builds to industrialiotdev)

tools/e2etesting/pipeline_orchestrated.yml
prmathur-microsoft: Taking care of these changes. Source will come from "Variables"

azure-pipelines.yml
marcschier: The code coverage step is not a test.  It just runs the tests again.  We should change testing the outcome to not regress from previous runs.  Just running the test again is not extra benefit, quite the opposite.

tools/scripts/acr-build.ps1
prmathur-microsoft: coming up shortly

tools/e2etesting/pipeline_orchestrated.yml
marcschier: Yes, the path should be configurable when the pipeline is set up in our devops instance.  That allows other users to reuse the templates and acripts and run them tests in their devops instances.

tools/scripts/acr-build.ps1
prmathur-microsoft: Good point. I'll refactor this.

azure-pipelines.yml
prmathur-microsoft: Interesting, then Code coverage step needs change. Either it can be merged with Testing to generate report then or if extra step is needed we need to gate on minimum required tests to be checked in any code. I'll make these changes as next PR then. Too big a change for this PR.

tools/scripts/acr-build.ps1
prmathur-microsoft: Explained it offline - in short to generate unique builds to run nightly and CI/CD for every commit. I'll send out the explaining this workflow as well.

marcschier: Cool.  Can we look at using the nuget package version produced by nbgv as tag and then save ourselves all the prerelase tag part?  It would only be a code change in setversion.ps1 and almost none here. Just a suggestion.

marcschier (approved):

tools/e2etesting/pipeline_orchestrated.yml
prmathur-microsoft: As per this DevOps article, Trigger blocks cannot use variables. So I have to revert this change.
https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema%2Cparameter-schema#triggers

karok2m (approved):

PR #671, opened 2020-08-12
Improve diag info from #666 for master

luiscantero:

PR #1185, opened 2021-05-25
Increased outgress queue size from 200 to 4096.

karok2m:
Ported part of changes present in #671 on `main` branch.


koepalex (approved):
:shipit:

luiscantero:
@karok2m  Sorry for being late, I just saw this. Pls. remember, there is no "outgress", it's called egress.

karok2m:
@luiscantero, I've omitted all the renaming parts from the original PR and only left the change of the queue size. So the code is still using outgress, thus the name of the PR itself.

PR #1191, opened 2021-05-28
Fix branch names for E2E pipeline triggers.

karok2m:
Aligning branch names with existing definitions.

### DIFF
--- a/tools/e2etesting/pipeline_orchestrated.yml
+++ b/tools/e2etesting/pipeline_orchestrated.yml
@@ -18,9 +18,9 @@ resources:
     trigger:
       branches:
         include:
-        - refs/heads/releases/*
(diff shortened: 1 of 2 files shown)


karok2m:
Both E2E pipeline triggers seem to work with existing branch names with `refs/heads/` prefixes. So we can also close this PR without merging it. @prmathur-microsoft, what do you think ?

karok2m:
As discussed with @prmathur-microsoft, we need `refs/heads/` prefix for this particular Azure DevOps step. More details can be found in `Note` part of [Branch filters](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/pipeline-triggers?view=azure-devops#branch-filters)

PR #1196, opened 2021-05-31
IAI: Changed logic of latest Kubernetes version selection.

karok2m:
Changes:
* Changed logic of latest Kubernetes version selection.
* Upgraded default AKS version to `1.18.17`.

The main change here is that now we select higher version between latest returned patch version from SDK and `AksMgmtClient.KUBERNETES_VERSION_FALLBACK`. This is required as the call that we perform through SDK no longer returns latest version that is available. At the moment there is a discrepancy between returned versions through SDK and Azure CLI. I've filed a bug for this in the GitHub repository of C# SDK. Please check it for more context on this fix. When the returned list in SDK is fixed, the changed logic should still choose the correct version.

Bug in C# SDK: [[BUG] Discrepancy in available Kubernetes versions (AKS) returned from Azure CLI and C# SDK #1242](https://github.com/Azure/azure-libraries-for-net/issues/1242)


karok2m:
Here is a run of Orchestrated E2E pipeline which deploys the platform with included changes: https://msazure.visualstudio.com/One/_build/results?buildId=43060606&view=results

koepalex (approved):
:shipit:

koepalex:
Pipeline build already passed IAI deployment, can be merged as soon as Github PR Checks have passed

PR #1189, opened 2021-05-27
IAI: Added default values for image namespace, container registry username and password.

karok2m:
Changes:
* Updated NuGet packages.
* Added default values for image namespace, container registry username and password.
* Changed default Kubernetes version to `1.18.14` as `1.18.10` is no longer supported.

Fixes #1182

PR #1197, opened 2021-05-31
IAI: Added default values for helm settings and changed logic of latest Kubernetes version selection. #1189 #1196

karok2m:
Applying the following changes on `main` branch:
* IAI: Added default values for image namespace, container registry username and password. (#1189)
* IAI: Changed logic of latest Kubernetes version selection. (#1196)


karok2m:
Kubernetes version selection change is required as the call that we perform through the SDK to get list of available Kubernetes versions in a specific region no longer returns latest version that is available there. At the moment there is a discrepancy between returned versions through SDK and Azure CLI. I've filed a bug for this in the GitHub repository of C# SDK. Please check it for more context on this fix. When the returned list in SDK is fixed, the changed logic should still choose the correct version.

Bug in C# SDK: [[BUG] Discrepancy in available Kubernetes versions (AKS) returned from Azure CLI and C# SDK #1242](https://github.com/Azure/azure-libraries-for-net/issues/1242)

deploy/src/Microsoft.Azure.IIoT.Deployment/Configuration/HelmSettings.cs
```diff
@@ -13,7 +13,10 @@ class HelmSettings {
         public static string _defaultRepoUrl = "https://microsoft.github.io/charts/repo";
         public static string _defaultChartVersion = "0.3.2";
         public static string _defaultImageTag = "2.7.206";
```
marcschier, line 15: Should we update this to 2.8.0?

marcschier (approved):

koepalex (approved):

deploy/src/Microsoft.Azure.IIoT.Deployment/Configuration/HelmSettings.cs
karok2m: Maybe once it is present in MCR ? At the moment `2.8.*` versions are only present in our development ACRs and for consuming from those I would need to check-in credentials for them.

PR #1204, opened 2021-06-04
Created integration tests framework, first cut.

suneetnangia:
Added initial core framework to provide starting point for integration test for publisher.
Some Features:
1. Using sample opcua servers and mocks in the existing codebase.
2. Minimal change to the existing codebase and no change in non mocking/testing codebase.
3. Sample tests to provide a starting point.

P.S. You may need to adapt it to your test cases e.g. add new node managers with specific use cases.

### DIFF
--- a/common/src/Microsoft.Azure.IIoT.Hub.Mock/src/Services/IoTHubServices.cs
+++ b/common/src/Microsoft.Azure.IIoT.Hub.Mock/src/Services/IoTHubServices.cs
@@ -113,7 +113,8 @@ namespace Microsoft.Azure.IIoT.Hub.Mock {
                 DeviceId = deviceId,
                 ModuleId = moduleId,
                 EnqueuedTimeUtc = DateTime.UtcNow,
-                Message = new Message(payload)
+                Message = new Message(payload),
+                DisposedMessage = payload
             };
             foreach (var item in message.Properties) {
                 ev.Message.Properties.Add(item.Key, item.Value);
@@ -384,6 +385,7 @@ namespace Microsoft.Azure.IIoT.Hub.Mock {
                     DeviceId = Device.Id,
                     ModuleId = Device.ModuleId,
                     Message = message,
+                    DisposedMessage = message.GetBytes(),
                     EnqueuedTimeUtc = DateTime.UtcNow
                 })) {
                     throw new CommunicationException("Failed to upload blob");
@@ -520,6 +522,8 @@ namespace Microsoft.Azure.IIoT.Hub.Mock {
             public DateTime EnqueuedTimeUtc { get; set; }
             /// <summary/>
             public Message Message { get; set; }
+            /// <summary/>
+            public byte[] DisposedMessage { get; set; }
         }
 
         private readonly SqlQuery _query;
--- a/common/src/Microsoft.Azure.IIoT.Hub.Mock/src/Services/IoTHubServices.cs
+++ b/common/src/Microsoft.Azure.IIoT.Hub.Mock/src/Services/IoTHubServices.cs
@@ -114,7 +114,7 @@ namespace Microsoft.Azure.IIoT.Hub.Mock {
                 ModuleId = moduleId,
                 EnqueuedTimeUtc = DateTime.UtcNow,
                 Message = new Message(payload),
-                DisposedMessage = payload
+                MessageData = payload
             };
             foreach (var item in message.Properties) {
                 ev.Message.Properties.Add(item.Key, item.Value);
@@ -385,7 +385,7 @@ namespace Microsoft.Azure.IIoT.Hub.Mock {
                     DeviceId = Device.Id,
                     ModuleId = Device.ModuleId,
                     Message = message,
-                    DisposedMessage = message.GetBytes(),
+                    MessageData = message.GetBytes(),
                     EnqueuedTimeUtc = DateTime.UtcNow
                 })) {
                     throw new CommunicationException("Failed to upload blob");
@@ -523,7 +523,7 @@ namespace Microsoft.Azure.IIoT.Hub.Mock {
             /// <summary/>
             public Message Message { get; set; }
             /// <summary/>
-            public byte[] DisposedMessage { get; set; }
+            public byte[] MessageData { get; set; }
         }
 
         private readonly SqlQuery _query;
--- a/modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublishedNodes/CanSendAlarmToIoTHubTest.json
+++ b/modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublishedNodes/CanSendAlarmToIoTHubTest.json
@@ -11,7 +11,7 @@
           "IsEnabled": true,
           "UpdateInterval": 10,
           "SnapshotInterval": 20,
-          "CompressMessages": false
+          "CompressedPayload": false
         }
       }
     ]
--- a/modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublishedNodes/CanSendAlarmToIoTHubTest.json
+++ b/modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublishedNodes/CanSendAlarmToIoTHubTest.json
@@ -1,0 +1,19 @@
+﻿[
+  {
+    "EndpointUrl": "opc.tcp://localhost:52210/UA/SampleServer",
+    "UseSecurity": false,
+    "OpcEvents": [
+      {
+        "DisplayName": "AlarmConditions",
+        "Id": "i=2253",
+        "TypeDefinitionId": "i=2041",
+        "PendingAlarms": {
+          "IsEnabled": true,
+          "UpdateInterval": 10,
+          "SnapshotInterval": 20,
+          "CompressMessages": false
+        }
+      }
+    ]
+  }
+]
--- a/modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublisherIntegrationTestBase.cs
+++ b/modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublisherIntegrationTestBase.cs
@@ -1,0 +1,196 @@
+﻿namespace Microsoft.Azure.IIoT.Modules.OpcUa.Publisher.Tests {
+    using Autofac;
+    using Microsoft.Azure.IIoT.Agent.Framework;
+    using Microsoft.Azure.IIoT.Diagnostics;
+    using Microsoft.Azure.IIoT.Hub;
+    using Microsoft.Azure.IIoT.Hub.Mock;
+    using Microsoft.Azure.IIoT.Hub.Models;
+    using Microsoft.Azure.IIoT.Module;
+    using Microsoft.Azure.IIoT.Module.Framework;
+    using Microsoft.Azure.IIoT.Module.Framework.Client;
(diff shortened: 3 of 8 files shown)


common/src/Microsoft.Azure.IIoT.Hub.Mock/src/Services/IoTHubServices.cs
```diff
@@ -520,6 +522,8 @@ public class EventMessage {
             public DateTime EnqueuedTimeUtc { get; set; }
             /// <summary/>
             public Message Message { get; set; }
+            /// <summary/>
+            public byte[] DisposedMessage { get; set; }
```
simonjaeger, line 526: Would this be better called MessageData or similar?

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublishedNodes/CanSendAlarmToIoTHubTest.json
```diff
@@ -0,0 +1,19 @@
+﻿[
+  {
+    "EndpointUrl": "opc.tcp://localhost:52210/UA/SampleServer",
```
simonjaeger, line 3: To confirm, you use a fixed port, so this always works?

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublisherIntegrationTestBase.cs
```diff
@@ -0,0 +1,196 @@
...
+        public PublisherIntegrationTestBase() {
+            // This is a fake but correctly formatted connection string.
+            var connectionString = $"HostName=dummy.azure-devices.net; DeviceId={Utils.GetHostName()};SharedAccessKeyName=iothubowner;SharedAccessKey=aXRpc25vdGFuYWNjZXNza2V5";
+            var config = connectionString.ToIoTHubConfig();
+
+            _typedConnectionString = ConnectionString.Parse(config.IoTHubConnString);
+            _exit = new TaskCompletionSource<bool>();            
+        }
+
+        protected async Task<List<JsonDocument>> ProcessMessages(string publishedNodesFile) {
```
simonjaeger, line 54: If I am not mistaken the solution still uses the Async suffix. Can you confirm? If so, let's add it here as well.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublisherIntegrationTestBase.cs
```diff
@@ -0,0 +1,196 @@
...
+            }
+
+            Exit();
+
+            var messages = new List<JsonDocument>();
+            foreach (var evt in Events) {
+                messages.Add(JsonDocument.Parse(Encoding.UTF8.GetString(evt.DisposedMessage)));
+            }
+
+            return messages;
```
simonjaeger, line 69: Would it make sense to return the list or simply return FirstOrDefault here if only expect one message?

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublisherIntegrationTestBase.cs
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        /// <summary>
+        /// Setup publishing from sample server.
+        /// </summary>
+        protected async Task HostPublisherAsync(ILogger logger, string publishedNodesFilePath) {
+            try {
+                var config = _typedConnectionString.ToIoTHubConfig();
+
+                var arguments = new List<string>();
```
simonjaeger, line 79: It would be nice to have a way to set these for each test. For example, set PubSub, reversible encoding, etc. It could be an improvement later if you would prefer.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublisherIntegrationTestBase.cs
```diff
@@ -0,0 +1,196 @@
...
+            _typedConnectionString = ConnectionString.Parse(config.IoTHubConnString);
+            _exit = new TaskCompletionSource<bool>();            
+        }
+
+        protected async Task<List<JsonDocument>> ProcessMessages(string publishedNodesFile) {
+            // publishedNodesFile points to the local server on the same machine and port currently as tests are run one at a time (not parallel) it does not need randomly generated port numbers.
+            _ = Task.Run(() => HostPublisherAsync(Mock.Of<ILogger>(), publishedNodesFile));
+
+            while (Events.Count == 0) {
+                await Task.Delay(500);
```
peterbryntesson, line 59: So it might break after 1 event? Are we satisifed which just one event?

peterbryntesson: No timeouts? It can go on forever?

peterbryntesson (approved):
Started mini-review - not my intention

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublisherIntegrationTestBase.cs
simonjaeger: Actually, maybe we can parameterize it and wait for the N number of events? Thoughts @suneetnangia ?

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublishedNodes/CanSendAlarmToIoTHubTest.json
suneetnangia: Yep, it works as all tests use the same server/port in a class. If we split the tests into multiple classes, it should still be fine as tests are not run in parallel unless tests are configured for parallism (IMHO). If we do run into a situation where we need to have unique ports, we can inject port as param to server fixture and base class.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublisherIntegrationTestBase.cs
suneetnangia: The loop will break when if finds the event for the node as specified in the publishednodes file or if no such node is present erroneous response is returned (bad response). In either cases it _should_ return and not go in loop indefinitely (but if there's such situation we can always add default timeout).
Absolutely, the idea to have this in basic form first before we start to have parameters as we deem right, so if we need to receive more than 1 message or define conditions for loop, we can certainly parameterise it.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublisherIntegrationTestBase.cs
suneetnangia: I guess once we decide if we need one or more than one we can make the appropriate changes, I was not sure at this stage.

simonjaeger (approved):
LGTM

common/src/Microsoft.Azure.IIoT.Hub.Mock/src/Services/IoTHubServices.cs
```diff
@@ -113,7 +113,8 @@ public class IoTHubServices : IIoTHubTwinServices,
                 DeviceId = deviceId,
                 ModuleId = moduleId,
                 EnqueuedTimeUtc = DateTime.UtcNow,
-                Message = new Message(payload)
+                Message = new Message(payload),
+                MessageData = payload
```
koepalex, line 117: Why adding the same payload twice?

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublisherIntegrationTestBase.cs
```diff
@@ -0,0 +1,213 @@
...
+    /// </summary>
+    public class PublisherIntegrationTestBase {
+        /// <summary>
+        /// Whether the module is running.
+        /// </summary>
+        private BlockingCollection<EventMessage> Events { get; set; } = new BlockingCollection<EventMessage>();
+
+        public PublisherIntegrationTestBase() {
+            // This is a fake but correctly formatted connection string.
+            var connectionString = $"HostName=dummy.azure-devices.net; DeviceId={Utils.GetHostName()};SharedAccessKeyName=iothubowner;SharedAccessKey=aXRpc25vdGFuYWNjZXNza2V5";
```
koepalex, line 49: even if this is only an example, please don't check something in that looks like credentials

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublishedNodes/CanSendAlarmToIoTHubTest.json
```diff
@@ -0,0 +1,19 @@
...
+    "OpcEvents": [
+      {
+        "DisplayName": "AlarmConditions",
+        "Id": "i=2253",
+        "TypeDefinitionId": "i=2041",
+        "PendingAlarms": {
+          "IsEnabled": true,
+          "UpdateInterval": 10,
+          "SnapshotInterval": 20,
+          "CompressMessages": false
```
peterbryntesson, line 14: Setting has changed name to 'CompressedPayload'

peterbryntesson (approved):

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublishedNodes/CanSendAlarmToIoTHubTest.json
koepalex: +1

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublisherIntegrationTestBase.cs
suneetnangia: That's a fair point, I tried using plaintext in SharedAccessKey initially but unfortunately we have an existing code in mock class (IoTHubClientFactory) which verifies that the SharedAccessKey is a valid base64 string hence a valid string is used, in this case the above base64 string decodes to "itisnotanaccesskey". I wanted to keep the changes to minimal for existing code.

common/src/Microsoft.Azure.IIoT.Hub.Mock/src/Services/IoTHubServices.cs
suneetnangia: This is as "Message = new Message(payload)" is disposed and we are unable to surface the payload at test level for verification.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/PublisherIntegrationTestBase.cs
koepalex: ok

common/src/Microsoft.Azure.IIoT.Hub.Mock/src/Services/IoTHubServices.cs
koepalex: ok, because this is a mock and no production code

PR #1206, opened 2021-06-10
Add reversible encoding tests

simonjaeger:
Adding integration tests for reversible encoding and fix the pending alarms' test.


modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/BasicIntegrationTests.cs
```diff
@@ -27,18 +27,31 @@ public class IntegrationTests : PublisherIntegrationTestBase, IClassFixture<OPCU
...
             // Act
             var messages = await ProcessMessagesAsync(publishedNodesFile, new TimeSpan(0, 0, 1), new TimeSpan(0, 2, 0), 1);
 
             // Assert
             Assert.Single(messages);            
             Assert.Equal("i=2253", messages[0].RootElement[0].GetProperty("NodeId").GetString());
-            Assert.Equal("AlarmConditions", messages[0].RootElement[0].GetProperty("DisplayName").GetString());
-            Assert.Equal("http://opcfoundation.org/AlarmCondition#s=1%3aColours%2fEastTank", messages[0].RootElement[0].GetProperty("Value")[1].GetProperty("Value").GetProperty("SourceNode").GetString());
-            Assert.Equal("100", messages[0].RootElement[0].GetProperty("Value")[1].GetProperty("Value").GetProperty("Severity").ToString());          
+            Assert.Equal("PendingAlarms", messages[0].RootElement[0].GetProperty("DisplayName").GetString());
```
peterbryntesson, line 52: As you say, these paths are not very elegant. Any way of hiding this?

peterbryntesson (approved):
LGTM

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/BasicIntegrationTests.cs
simonjaeger: I looked around. The only way seem to be to serialize it again, and then deserialize it with another library (i.e. Newtonsoft). I didn't like that solution.

simonjaeger:
Looking into the failing build...

koepalex (approved):

PR #1224, opened 2021-06-11
Reading OPC Publisher standalone documentation

koepalex:
Having the split for OPC Publisher documentation is currently confusing for anyone. Therefor OPC Publisher will be handled again in this repository and the [iot-edge-opc-publisher](https://github.com/Azure/iot-edge-opc-publisher) is going to be achieved.

With this PR the updated documentation will be added as well as the tutorial for using local grafana dashboard

**Using Main as base branch**


docs/modules/publisher-commandline.md
```diff
@@ -0,0 +1,383 @@
+[Home](../../readme.md)
+
+### OPC Publisher Command Line Arguments for Version 2.5 and below
```
cristipogacean, line 3: >for Version 2.5 and below [](http://example.com/codeflow?start=41&length=25)

I think the order of the topics could be confusing for the user. This repo does not contain 2.5 or below code/versions. It would make more sense to start with 2.8 command line arguments and add "2.5 and below" arguments at the bottom of the doc.

koepalex: changed

marcschier (approved):

PR #1227, opened 2021-06-14
Check where clause 

peterbryntesson:
Feature that (asynchronously) checks whether the types referenced in the where clause is valid types - and writes a warning if it isn't.

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
@@ -1162,6 +1162,23 @@ namespace Microsoft.Azure.IIoT.OpcUa.Protocol.Services {
                         }
                     }
 
+                    Task.Run(() => {
+                        foreach (var element in eventFilter.WhereClause.Elements) {
+                            if (element.FilterOperator == FilterOperator.OfType) {
+                                foreach (var filterOperand in element.FilterOperands) {
+                                    var nodeId = default(NodeId);
+                                    try {
+                                        nodeId = (filterOperand.Body as LiteralOperand).Value.ToString().ToNodeId(messageContext);
+                                        nodeCache.FetchNode(nodeId.ToExpandedNodeId(messageContext.NamespaceUris)); // it will throw an exception if it doesn't work
+                                    }
+                                    catch (Exception ex) {
+                                        _logger.Warning($"Where clause is doing OfType({nodeId}) and we got this message {ex.Message} while looking it up");
+                                    }
+                                }
+                            }
+                        }
+                    });
+
                     // let's keep track of the internal fields we add so that they don't show up in the output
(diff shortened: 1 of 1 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -1162,6 +1162,23 @@ public class MonitoredItemWrapper {
                         }
                     }
 
+                    Task.Run(() => {
```
simonjaeger, line 1165: Why this?

peterbryntesson: Alex specifically wanted it done asynchronously

simonjaeger: @koepalex Thoughts on this?

koepalex: We already have issues for other customer if something isn't executed asynchrounsly. Because it blocks the execution (delivery of telemetry). Each feature that isn't mandatory to wait (like here the convenience feature to detect wrong configurations) should be done as default async and without waiting.

koepalex: Based on experience lookup on 1000 OPC UA nodes, adds around 10 seconds. But if this is pretty fast (< 250ms) in worse case scenario (each event points to unique type). Then it would be fine to execute synchronously.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -1162,6 +1162,23 @@ public class MonitoredItemWrapper {
...
+                        foreach (var element in eventFilter.WhereClause.Elements) {
+                            if (element.FilterOperator == FilterOperator.OfType) {
+                                foreach (var filterOperand in element.FilterOperands) {
+                                    var nodeId = default(NodeId);
+                                    try {
+                                        nodeId = (filterOperand.Body as LiteralOperand).Value.ToString().ToNodeId(messageContext);
```
koepalex, line 1171: >. [](http://example.com/codeflow?start=87&length=1)

safer to use ?.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
simonjaeger: I see. Thanks for the context.

simonjaeger (approved):
LGTM

koepalex (approved):
:shipit:

PR #1239, opened 2021-06-17
Update metrics collector to the released image

cristipogacean:
Update metrics collector to the released image

### DIFF
--- a/common/src/Microsoft.Azure.IIoT.Core/src/Deploy/ILogWorkspaceConfig.cs
+++ b/common/src/Microsoft.Azure.IIoT.Core/src/Deploy/ILogWorkspaceConfig.cs
@@ -19,5 +19,10 @@ namespace Microsoft.Azure.IIoT.Deploy {
         /// Log Analytics Workspace Key
         /// </summary>
         string LogWorkspaceKey { get; }
+
+        /// <summary>
+        /// The Resource Is of the IoT Hub
+        /// </summary>
+        string IoTHubResourceId { get; }
     }
 }
--- a/common/src/Microsoft.Azure.IIoT.Core/src/Deploy/ILogWorkspaceConfig.cs
+++ b/common/src/Microsoft.Azure.IIoT.Core/src/Deploy/ILogWorkspaceConfig.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Azure.IIoT.Deploy {
         string LogWorkspaceKey { get; }
 
         /// <summary>
-        /// The Resource Is of the IoT Hub
+        /// The Resource Id of the IoT Hub
         /// </summary>
         string IoTHubResourceId { get; }
     }
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Deploy/IoTHubDiscovererDeployment.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Deploy/IoTHubDiscovererDeployment.cs
@@ -95,15 +95,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Registry.Deploy {
                     Cmd = new[] {
                         "PkiRootPath=/mount/pki"
                     },
-                    NetworkingConfig = new {
-                        EndpointsConfig = new {
-                            host = new {
-                            }
-                        }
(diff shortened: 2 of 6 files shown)


prmathur-microsoft:
Two Questions - 
1. Is the discovery module Fix approved for LTS?
2. Is the metrics collector fix approved for LTS?

If so, can it be split into separate commits (for easier revert if needed)

koepalex (approved):
:shipit:

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Deploy/IoTHubDiscovererDeployment.cs
```diff
@@ -95,15 +95,7 @@ public sealed class IoTHubDiscovererDeployment : IHostProcess {
                     Cmd = new[] {
                         "PkiRootPath=/mount/pki"
                     },
-                    NetworkingConfig = new {
```
marcschier, line 98: Can we break this file into a separate PR?  is the discovery module still able to enumerate the subnets of the network adapters?  Or does this require now to also specify subnet mask?

common/src/Microsoft.Azure.IIoT.Core/src/Deploy/ILogWorkspaceConfig.cs
```diff
@@ -19,5 +19,10 @@ public interface ILogWorkspaceConfig {
         /// Log Analytics Workspace Key
         /// </summary>
         string LogWorkspaceKey { get; }
+
+        /// <summary>
+        /// The Resource Is of the IoT Hub
```
marcschier, line 24: typo

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Deploy/IoTHubMetricsCollectorDeployment.cs
```diff
@@ -108,11 +106,20 @@ public sealed class IoTHubMetricsCollectorDeployment : IHostProcess {
...
                                 ""value"": """ + _config.LogWorkspaceId + @"""
                             },
-                            ""AzMonWorkspaceKey"": {
+                            ""LogAnalyticsSharedKey"": {
                                 ""value"": """ + _config.LogWorkspaceKey + @"""
+                            },
+                            ""ResourceId"": {
+                                ""value"": """ + _config.IoTHubResourceId + @"""
+                            },
+                            ""MetricsEndpointsCSV"": {
```
marcschier, line 121: Should we add discovery module as well?  Does this value support wildcards?

marcschier (changes requested):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Deploy/IoTHubDiscovererDeployment.cs
cristipogacean: sure, I will move this to a separate PR. 
with this change the user needs to specify the subnet mask, in a same manner as it needs to be done for Windows containers. Do you think this is a blocker?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Deploy/IoTHubMetricsCollectorDeployment.cs
cristipogacean: For some reason the discovery was not part of this in the first place. That's the only reason I did not add it. 
I did not find any reference in the documentation on supporting wildcards.

marcschier (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Deploy/IoTHubDiscovererDeployment.cs
marcschier: Yes, let's understand the root cause a bit more before we make the other change.  We can deal with it in the other PR.

PR #1241, opened 2021-06-18
fix discovery not showing up bug

cristipogacean:


marcschier (changes requested):
Let's talk more about it, especially why this is failing now, while it worked fine for quite a while.

PR #1242, opened 2021-06-18
Operations manual converted to markdown

marcschier:


tools/scripts/acr-copy-release.ps1
```diff
@@ -130,6 +132,10 @@ $jobs = @()
 foreach ($Repository in $BuildRepositories) {
 
     $BuildTag = "$($Repository):$($script:ReleaseVersion)"
+    if (![string]::IsNullOrEmpty($script:BuildNamespace) -and `
```
karok2m, line 135: @marcschier, I do not see value of `BuildNamespace` being used anywhere else. And here, in the case it is not a already part of `Repository` it is again not used. Is this correct ? Don't we need to append it to repository name somewhere ?

docs/modules/publisher.md
```diff
@@ -10,7 +10,7 @@ It runs on [Azure IoT Edge](https://azure.microsoft.com/en-us/services/iot-edge/
 
 Please use our released containers for OPC Publisher available in the Microsoft Container Registry, rather than building from sources. The easiest way to deploy OPC Publisher is through the [Azure Marketplace](https://azuremarketplace.microsoft.com/en-us/marketplace/apps/microsoft_iot.iotedge-opc-publisher). 
 
-[<img src="image-20201028141833399.png" style="zoom:50%;" />](https://azuremarketplace.microsoft.com/en-us/marketplace/apps/microsoft_iot.iotedge-opc-publisher)
```
karok2m, line 13: The image does not seem to be present. Please check this document:

https://github.com/Azure/Industrial-IoT/blob/mschier%2Fopsmanual/docs/modules/publisher.md

image

tools/scripts/acr-copy-release.ps1
marcschier: No, this is just to pick a namespace from a build registry like industrialiotdev, otherwise the code runs through all namespaces and copies the same tag if available from all causing the wrong repositories (and images) to be picked.

docs/modules/publisher.md
marcschier: Correct, let me see if I can dig it up from the publisher repo.

marcschier: Done.

karok2m (approved):

PR #1243, opened 2021-06-21
Fix filtering of LocalizedText in EncodeableDictionary

simonjaeger:
Allow LocalizedText without Locale (only Text) to be encoded. For example, generated by OPC-PLC.


peterbryntesson (approved):
LGTM. Good catch!

mregen (approved):

PR #1244, opened 2021-06-21
set 'releases' branch ready

prmathur-microsoft:
*edit starts with in acr-build to capture 'releases' branch

### DIFF
--- a/tools/scripts/acr-build.ps1
+++ b/tools/scripts/acr-build.ps1
@@ -79,7 +79,7 @@ else {
         # dev feature builds
         $namespace = $namespace.Replace("feature/", "")
     }
-    elseif ($namespace.StartsWith("release") -or ($namespace -eq "releases")) {
+    elseif ($namespace.StartsWith("release/") -or ($namespace -eq "releases")) {
         $namespace = "public"
         if ([string]::IsNullOrEmpty($Registry)) {
             # Release and Preview builds go into staging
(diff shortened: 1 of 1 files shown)


tools/scripts/acr-build.ps1
```diff
@@ -79,7 +79,7 @@ else {
         # dev feature builds
         $namespace = $namespace.Replace("feature/", "")
     }
-    elseif ($namespace.StartsWith("release/") -or ($namespace -eq "master")) {
+    elseif ($namespace.StartsWith("release") -or ($namespace -eq "master")) {
```
marcschier, line 82: I would recommend testing for release/ and releases separately.  I would also remove "master", which then could be replaced by "releases".   That way it is clear to the reader which namespace is meant here.

marcschier (changes requested):

tools/scripts/acr-build.ps1
prmathur-microsoft: Agreed on removing "master". I can remove from here.
Since we were using StartsWith, shouldn't "release" suffice? What am I missing with 
```suggestion
    elseif ($namespace.StartsWith("release") -or ($namespace -eq "releases")) {
```
vs 
```suggestion
    elseif ($namespace.StartsWith("release")) {
```

marcschier (approved):

PR #1245, opened 2021-06-22
User Guide for output from events

peterbryntesson:
A guide detailing the output from events received in the publisher

### DIFF
--- a/docs/opc-event-output.md
+++ b/docs/opc-event-output.md
@@ -5,44 +5,44 @@ This section describes what the output looks like when listening for events in t
 Here is an example of the output we get when listening to events from the Simple Events sample:
 ```json
 {
-	"body": [
-		{
-			"MessageId": "43",
-			"MessageType": "ua-data",
-			"PublisherId": "SIMPLE-EVENTS",
-			"DataSetWriterGroup": "SIMPLE-EVENTS",
-			"Messages": [
-				{
-					"DataSetWriterId": "SIMPLE-EVENTS",
-					"MetaDataVersion": {
-						"MajorVersion": 1,
-						"MinorVersion": 0
-					},
-					"Payload": {
-						"SimpleEvents": {
-							"Value": {
-								"EventId": "+6CQjN1eqkO6+yHJnxMz5w==",
-								"EventType": "http://microsoft.com/Opc/OpcPlc/SimpleEvents#i=14",
-								"Message": "The system cycle '59' has started.",
-								"ReceiveTime": "2021-06-21T12:38:55.5814091Z",
-								"Severity": 1,
-								"SourceName": "System",
-								"SourceNode": "i=2253",
-								"Time": "2021-06-21T12:38:55.5814078Z"
-							},
-							"SourceTimestamp": "2021-06-21T12:38:55.5814091Z"
-						}
-					}
-				}
-			]
-		}
-	],
-	"enqueuedTime": "Mon Jun 21 2021 14:39:02 GMT+0200 (Central European Summer Time)",
-	"properties": {
-		"$$ContentType": "application/x-network-message-json-v1",
-		"iothub-message-schema": "application/ua+json",
-		"$$ContentEncoding": "utf-8"
-	}
+  "body": [
+    {
+      "MessageId": "43",
+      "MessageType": "ua-data",
+      "PublisherId": "SIMPLE-EVENTS",
+      "DataSetWriterGroup": "SIMPLE-EVENTS",
+      "Messages": [
+        {
+          "DataSetWriterId": "SIMPLE-EVENTS",
+          "MetaDataVersion": {
+            "MajorVersion": 1,
+            "MinorVersion": 0
+          },
+          "Payload": {
+            "SimpleEvents": {
+              "Value": {
+                "EventId": "+6CQjN1eqkO6+yHJnxMz5w==",
+                "EventType": "http://microsoft.com/Opc/OpcPlc/SimpleEvents#i=14",
+                "Message": "The system cycle '59' has started.",
+                "ReceiveTime": "2021-06-21T12:38:55.5814091Z",
+                "Severity": 1,
+                "SourceName": "System",
+                "SourceNode": "i=2253",
+                "Time": "2021-06-21T12:38:55.5814078Z"
+              },
+              "SourceTimestamp": "2021-06-21T12:38:55.5814091Z"
+            }
+          }
+        }
+      ]
+    }
+  ],
+  "enqueuedTime": "Mon Jun 21 2021 14:39:02 GMT+0200 (Central European Summer Time)",
+  "properties": {
+    "$$ContentType": "application/x-network-message-json-v1",
+    "iothub-message-schema": "application/ua+json",
+    "$$ContentEncoding": "utf-8"
+  }
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/opc-event-output.md
```diff
@@ -0,0 +1,199 @@
+# Detailing the output for OPC UA Events in OPC Publisher
+
+This section describes what the output looks like when listening for events in the OPC Publisher. Events are produced in the PubSub format specified in the OPC UA Standard. The payload is an event which consists of fields selected in the select clause and it's values.
```
simonjaeger, line 3: "it's values" => "its values".

docs/opc-event-output.md
```diff
@@ -0,0 +1,199 @@
...
+  ],
+  "enqueuedTime": "Mon Jun 21 2021 14:39:02 GMT+0200 (Central European Summer Time)",
+  "properties": {
+    "$$ContentType": "application/x-network-message-json-v1",
+    "iothub-message-schema": "application/ua+json",
+    "$$ContentEncoding": "utf-8"
+  }
+}
+```
+A few things to note here. Here the event in the payload is named "SimpleEvents" which it has been configured to by setting the DisplayName property of the events configuration. Also note that all fields+values reside under the Value key.
```
simonjaeger, line 48: "which it has" => "which has"

simonjaeger (approved):
LGTM

docs/opc-event-output.md
```diff
@@ -0,0 +1,199 @@
...
+                      "EventType",
+                      "Message",
+                      "ReceiveTime",
+                      "Severity",
+                      "SourceName",
+                      "SourceNode",
+                      "Time"
+                    ],
+                    "EventId": {
+                      "Type": "ByteString",
```
mregen, line 86: strange, builtin types should be encoded as number

peterbryntesson: Should this be encoded as a number?
image

peterbryntesson: Or do you mean some other field?

docs/opc-event-output.md
```diff
@@ -0,0 +1,199 @@
...
+          "MetaDataVersion": {
+            "MajorVersion": 1,
+            "MinorVersion": 0
+          },
+          "Payload": {
+            "SimpleEvents": {
+              "Value": {
+                "Type": "ExtensionObject",
+                "Body": {
+                  "TypeId": "s=EncodeableDictionary",
```
mregen, line 71: Here it says 's=', so its a custom type in namespace 0? But encodeabledictionary is not a standard OPC UA type?

docs/opc-event-output.md
```diff
@@ -0,0 +1,199 @@
...
+{
+  "body": [
+    {
+      "MessageId": "43",
+      "MessageType": "ua-data",
+      "PublisherId": "SIMPLE-EVENTS",
+      "DataSetWriterGroup": "SIMPLE-EVENTS",
+      "Messages": [
+        {
+          "DataSetWriterId": "SIMPLE-EVENTS",
```
mregen, line 16: Datasetwriterid is a number (I believe no yet fixed in iot repo)

docs/opc-event-output.md
```diff
@@ -0,0 +1,199 @@
...
+
+Here is an example of the output we get when listening to events from the Simple Events sample:
+```json
+{
+  "body": [
+    {
+      "MessageId": "43",
+      "MessageType": "ua-data",
+      "PublisherId": "SIMPLE-EVENTS",
+      "DataSetWriterGroup": "SIMPLE-EVENTS",
```
mregen, line 13: The name should be DataSetClassId

mregen: and a guid

mregen (commented):
Looking at your documentation I started some head scratching...

docs/opc-event-output.md
mregen: The type

peterbryntesson:
> Looking at your documentation I started some head scratching...

I can see that. Much (if not all) of what you are remarking on is already there in 2.7, nothing we added as part of the events feature

docs/opc-event-output.md
mregen: example for encoding of a double:
```json
 "#0": {
        "Value": {
          "Type": 11,
          "Body": 6628948.417515845
        },

Comment (peterbryntesson) [reply docs/opc-event-output.md:86]: ah, ok. I see.

File: docs/opc-event-output.md [hunk 7]
```diff
@@ -0,0 +1,199 @@
...
+  "properties": {
+    "$$ContentType": "application/x-network-message-json-v1",
+    "iothub-message-schema": "application/ua+json",
+    "$$ContentEncoding": "utf-8"
+  }
+}
+```
+A few things to note here. Here the event in the payload is named "SimpleEvents" which has been configured to by setting the DisplayName property of the events configuration. Also note that all fields+values reside under the Value key.
+
+The format produced here does not contain enough information to decode the JSON properly. If you need this there is a command-line switch in the OPC Publisher called "reversibleencoding", or "re" which can be set to either true or false. The default value is false. If you enable this setting the output will look like this:
```
mregen, line 50: This statement is not quite right.
Normally the receiver has metadata information about the dataset message, e.g. datasetwriter id and configured keys.
For each of the keys in a payload it knows the type, so it can do some reversible decoding, apparently not missing information like locale....
So when you chose reversible encoding, there is also a lot of redundant information being sent. I know thats not the way things are implemented here and it needs to be enhanced.

docs/opc-event-output.md
peterbryntesson: No, it's an implementation in the OPC publisher and needs to be registrered as a custom type

docs/opc-event-output.md
mregen: its correct here: https://github.com/Azure/Industrial-IoT/blob/master/docs/dev-guides/telemetry-messages-format.md
was it changed to datasetwritergroup?

peterbryntesson: I just ran the current code and generated some output and used that as documentation. We have used merged with main and this is how it looks. Don't know the state of master.

mregen:
Hi @peterbryntesson, I saw in your walkthrough that encodeableDictionary implementation was changed, maybe new implementation needs to be revisited.

docs/opc-event-output.md
```diff
@@ -0,0 +1,199 @@
...
+          },
+          "Payload": {
+            "SimpleEvents": {
+              "Value": {
+                "Type": "ExtensionObject",
+                "Body": {
+                  "TypeId": "s=EncodeableDictionary",
+                  "Encoding": "Json",
+                  "Body": {
+                    "_Count": 8,
```
koepalex, line 74: Please update to latest changes, _Count and _Keys have to be removed

docs/opc-event-output.md
```diff
@@ -0,0 +1,199 @@
...
+                    "_Keys": [
```
koepalex, line 75: Please update to latest changes, _Count and _Keys have to be removed

simonjaeger: Done!

docs/opc-event-output.md
simonjaeger: Done!

PR #1253, opened 2021-06-25
Fix missed vulnerable package in test code dependency

marcschier:
* And update test code to not spew warnings.

### DIFF
--- a/tools/e2etesting/TestEventProcessor/TestEventProcessor.Businesslogic/TelemetryValidator.cs
+++ b/tools/e2etesting/TestEventProcessor/TestEventProcessor.Businesslogic/TelemetryValidator.cs
@@ -164,11 +164,11 @@ namespace TestEventProcessor.BusinessLogic {
         /// Stop monitoring of events.
         /// </summary>
         /// <returns></returns>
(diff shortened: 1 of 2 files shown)


koepalex (approved):
:shipit:

tools/e2etesting/TestEventProcessor/TestEventProcessor.Businesslogic/TelemetryValidator.cs
```diff
@@ -164,11 +164,11 @@ public async Task<StartResult> StartAsync(ValidatorConfiguration configuration)
         /// Stop monitoring of events.
         /// </summary>
         /// <returns></returns>
-        public async Task<StopResult> StopAsync()
+        public Task<StopResult> StopAsync()
```
cristipogacean, line 167: Shouldn't we rename the method to reflect async removal?

koepalex: The method return a task object, therefor from caller side it is still asynchronously. the `async` keyword is only syntactic sugar. Therefor I wouldn't rename the method

PR #1255, opened 2021-06-29
Remove encoded Keys and Count from EncodeableDictionary

simonjaeger:
By taking a dependency on the JsonDecoderEx, which can decode a dictionary, we no longer need to encode Count and Keys separately.


peterbryntesson (approved):
LGTM!

koepalex (approved):
:shipit:

Issue #1306, opened 2021-07-29
Incorrect node is subscribed on aggregation server if the underlying server connects after the publisher.


Haeavar:
**Describe the bug**
We are currently using an aggregation OpcUa server (AS) to aggregate multiple underlying OpcUa server (US1, USn) using reverse connect. Each US1 server have its own namespace URIs which will be added to the AS namespace table. This happens only after the first time the US1 has connected to the AS.

If the publisher creates the subscription **before** the US1 has connected to AS, the namespaceUri is resolved to the namespace idx  65535 which is obviously not the correct one. The Idx 65535 comes from the `namespaces.GetIndex(nodeId.NamespaceUri)` and casting the result to `ushort`.

This wrong behavior will result in a wrong subscription. If the US1 connects after the publisher to the AS, the namespace will be added to the namespaceTable. But the publisher never refreshes its namespaceTable (on the opened session). Only a reconnect will solve the issue.

**To Reproduce**
1. Create a published nodes json for an aggregating OpcUa server with one node using the namespaceUri declaration `{ "Id": "nsu=urn:UnderlyingOpcUa:TRUMPF:UAInterfaces/http://trumpf.com/TRUMPF-Interfaces/;s=33" }`
2. Keep the underlying OPCUA server (`UnderlyingOpcUa` in this sample) disconnected
3. Start the Publisher
4. Observe, that the node `s=33` is subscribed with namespace idx `65535` 

**Expected behavior**
The subscription fails because the namespace is not found in the namespace table. As soon as the namespace can be found, the subscription is created.

**Screenshots**
This is the pn.json part:
![image](https://user-images.githubusercontent.com/31273449/127521891-44362754-da17-480c-a1f5-0a48795c4453.png)

And the publisher output:
![image](https://user-images.githubusercontent.com/31273449/127521792-476ed608-dca2-4854-a9f9-fa68b8a3178b.png)

**OpcPublisher:**
 - OpcPublisher version 2.8.0.

**Additional context**
I will create an PR to fix this bug.

marcschier:
Thanks for the contributed fix @Haeavar !

koepalex:
@Haeavar with the PR you submitted is this issue solved and can be closed?

koepalex:
Thanks for your contribution, I will close the issue for now. Please reopen if necessary

PR #1307, opened 2021-07-29
Fix bug, if namespaceUri is currently not available in the namespace …

Haeavar:
This PR will fix the bug, that the opcnode namespace idx is incorrect resolved to idx 65535 if the namespaceUri was not found in the namespace table.
The namespace table will be reloaded everytime the subscriptions needs to be refreshed.

See Bug #1306 

Please approve or comment the PR since this blocks us using the publisher in our environment.


marcschier (approved):
Lgtm

PR #1324, opened 2021-08-18
Simplified payload identifier detection in NetworkMessageEncoder.

karok2m:
Changes:
* Changed `NetworkMessageEncoder.GetPayloadIdentifier()` to also cache values that were not found by the lookup but were the fallback mechanism. This is usually expanded node id.

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
@@ -71,7 +71,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine {
                 var resultJson = EncodeAsJson(messages, maxMessageSize);
                 var resultUadp = EncodeAsUadp(messages, maxMessageSize);
                 var result = resultJson.Concat(resultUadp);
-                return Task.FromResult(result);
+                return Task.FromResult<IEnumerable<NetworkMessageModel>>(result.ToList());
             }
             catch (Exception e) {
                 _logger.Error(e, "Failed to encode {numOfMessages} messages", messages.Count());
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
@@ -441,20 +441,27 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine {
                 //do the long running lookup as less as possible
                 var dataSetWriter = message.Writer;
                 foreach (var publishedVariableData in dataSetWriter.DataSet.DataSetSource.PublishedVariables.PublishedData) {
-                    if ((publishedVariableData.PublishedVariableNodeId == notification.NodeId
-                            || publishedVariableData.PublishedVariableNodeId.ToExpandedNodeId(context).AsString(context) == notification.NodeId.ToExpandedNodeId(context.NamespaceUris).AsString(context)) &&
-                        publishedVariableData.Id != notification.NodeId) {
-                        _knownPayloadIdentifiers[notification.NodeId.ToString()] = publishedVariableData.Id;
-                        return publishedVariableData.Id;
+                    if (publishedVariableData.PublishedVariableNodeId == notification.NodeId
+                        || publishedVariableData.PublishedVariableNodeId.ToExpandedNodeId(context).AsString(context) == notification.NodeId.ToExpandedNodeId(context.NamespaceUris).AsString(context)) {
+                        if (publishedVariableData.Id != notification.NodeId) {
+                            _knownPayloadIdentifiers[notification.NodeId.ToString()] = publishedVariableData.Id;
+                            return publishedVariableData.Id;
+                        } else {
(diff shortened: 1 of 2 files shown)


karok2m:
@koepalex, with those changes the lookup times are similar for 1K and 20K scenarios. In both cases the first run that caches values takes 1.5-1.6 seconds and consecutive runs take 0.001-0.002 seconds.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
```diff
@@ -429,30 +434,35 @@ public class NetworkMessageEncoder : IMessageEncoder {
                 throw new ArgumentNullException(nameof(context));
             }
 
-            if (_knownPayloadIdentifiers.TryGetValue(notification.NodeId.ToString(), out var knownPayloadIdentifier)) {
-                if (!string.IsNullOrEmpty(knownPayloadIdentifier)) {
-                    return knownPayloadIdentifier;
-                }
+            if (_knownPayloadIdentifiers.TryGetValue(notification.NodeId.ToString(), out var knownPayloadIdentifier) && !string.IsNullOrEmpty(knownPayloadIdentifier)) {
```
koepalex: >notification.NodeId.ToString() [](http://example.com/codeflow?start=53&length=30)

Avoid multiple calls to notification.NodeId.ToString(), store it as local variable

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
```diff
@@ -74,7 +74,8 @@ public class NetworkMessageEncoder : IMessageEncoder {
                 return Task.FromResult(result);
```
koepalex: should we add result.ToList() here to ensure that the enumeration was iterated and the exception (if any) occurs directly and could be catched?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
```diff
@@ -429,30 +434,35 @@ public class NetworkMessageEncoder : IMessageEncoder {
...
-                    foreach (var publishedVariableData in dataSetWriter.DataSet.DataSetSource.PublishedVariables.PublishedData) {
-                        if ((publishedVariableData.PublishedVariableNodeId == notification.NodeId
-                              || publishedVariableData.PublishedVariableNodeId.ToExpandedNodeId(context).AsString(context) == notification.NodeId.ToExpandedNodeId(context.NamespaceUris).AsString(context)) &&
-                            publishedVariableData.Id != notification.NodeId) {
+            else {
+                //do the long running lookup as less as possible
+                var dataSetWriter = message.Writer;
+                foreach (var publishedVariableData in dataSetWriter.DataSet.DataSetSource.PublishedVariables.PublishedData) {
+                    if (publishedVariableData.PublishedVariableNodeId == notification.NodeId
+                        || publishedVariableData.PublishedVariableNodeId.ToExpandedNodeId(context).AsString(context) == notification.NodeId.ToExpandedNodeId(context.NamespaceUris).AsString(context)) {
```
koepalex: >notification.NodeId.ToExpandedNodeId(context.NamespaceUris).AsString(context) [](http://example.com/codeflow?start=120&length=77)

Avoid multiple calls to notification.NodeId.ToExpandedNodeId(context.NamespaceUris).AsString(context) store it as local variable

mregen (dismissed):
:shipit:

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
karok2m: I will have to use another transformation then to return the correct type, `result.ToList().AsEnumerable()`. So I would say that we keep it as it is to avoid those transformation on each encoding calls. What do you think ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
karok2m: Fixed.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
karok2m: Fixed.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/NetworkMessageEncoder.cs
mregen: A list is also enumerable...

karok2m: OK. I had a problem with wrapping it in the `Task.FromResult()`. It is done now.

koepalex (approved):
:shipit:

PR #1332, opened 2021-08-23
Installing .NET Core SDK for deploying the platform.

karok2m:
Azure DevOps Bug: [10621286](https://msazure.visualstudio.com/One/_workitems/edit/10621286)

We need to make sure that `dotnet` is present on the host machine for platform deployment stage in E2E Orchestrated pipeline as it has versioning step which requires it.

### DIFF
--- a/tools/e2etesting/steps/deployplatform.yml
+++ b/tools/e2etesting/steps/deployplatform.yml
@@ -1,4 +1,12 @@
 steps:
+- task: UseDotNet@2
+  displayName: 'Install .NET Core SDK for building'
+  inputs:
+    packageType: sdk
+    version: 3.1.x
+    includePreviewVersions: false
+    installationPath: $(Agent.ToolsDirectory)/dotnet
+
 - task: AzureCLI@2
   displayName: 'Set Service Principal Environment Variables'
   name: promoteserviceprincipal
--- a/tools/e2etesting/steps/runtests.yml
+++ b/tools/e2etesting/steps/runtests.yml
@@ -5,6 +5,14 @@ parameters:
   type: string
 
 steps:
+- task: UseDotNet@2
+  displayName: 'Install .NET Core SDK'
+  inputs:
+    packageType: sdk
+    version: 3.1.x
+    includePreviewVersions: false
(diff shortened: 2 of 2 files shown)


koepalex (dismissed):

monikavar (dismissed):

tools/e2etesting/steps/deployplatform.yml
```diff
@@ -1,4 +1,12 @@
 steps:
+- task: UseDotNet@2
+  displayName: 'Install .NET Core SDK for building'
+  inputs:
+    packageType: sdk
+    version: 3.1.x
+    includePreviewVersions: false
+    installationPath: $(Agent.ToolsDirectory)/dotnet
```
mregen, line 8: how come? did they change the default images again?

karok2m: This pipeline is running on self-hosted agents, that is why we have to make sure that all our dependencies are present. And I think that it is better to define our `dotnet` dependency in the pipeline and thus make sure that we get the latest one instead of installing a specific version on the host and forgetting about it.

mregen: Sounds fair, or you check the version and only install it if its not available yet? But I guess timing is not critical here...

karok2m: The task actually checks if desired SDK is already present:

```
Starting: Install .NET Core SDK
==============================================================================
Task         : Use .NET Core
Description  : Acquires a specific version of the .NET Core SDK from the internet or the local cache and adds it to the PATH. Use this task to change the version of .NET Core used in subsequent tasks. Additionally provides proxy support.
Version      : 2.191.0
Author       : Microsoft Corporation
Help         : https://aka.ms/AA4xgy0
==============================================================================
Tool to install: .NET Core sdk version 3.1.x.
Found version 3.1.412 in channel 3.1 for user specified version spec: 3.1.x
Version: 3.1.412 was found in cache.
Creating global tool path and pre-pending to PATH.
Finishing: Install .NET Core SDK
```

koepalex (approved):
:shipit:

PR #1337, opened 2021-08-24
Increased default size of outgress buffer in legacy options to 4096.

karok2m:


koepalex (approved):
Important change, now the orchestrated and standalone variant using the same default again

PR #1369, opened 2021-08-30
Removed bad nodes from E2E publisher test subscriptions.

karok2m:
OPC PLC now has bad nodes that drop messages by design. In this PR I have removed those bad nodes from E2E publisher tests to make sure that they don't interfere with our dropped and duplicate message detection logic.

### DIFF
--- a/e2e-tests/IIoTPlatform-E2E-Tests/Orchestrated/B_PublishMultipleNodesOrchestratedTestTheory.cs
+++ b/e2e-tests/IIoTPlatform-E2E-Tests/Orchestrated/B_PublishMultipleNodesOrchestratedTestTheory.cs
@@ -93,7 +93,11 @@ namespace IIoTPlatform_E2E_Tests.Orchestrated
 
             var cts = new CancellationTokenSource(TestConstants.MaxTestTimeoutMilliseconds);
             var testPlc = _context.SimulatedPublishedNodes[_context.ConsumedOpcUaNodes.First().Key];
-            _context.ConsumedOpcUaNodes.First().Value.OpcNodes = testPlc.OpcNodes.Skip(250).ToArray();
+
+            // We will filter out bad fast and slow nodes as they to drop messages by design.
+            _context.ConsumedOpcUaNodes.First().Value.OpcNodes = testPlc.OpcNodes
+                .Where(node => !node.Id.Contains("bad", StringComparison.OrdinalIgnoreCase))
+                .Skip(250).ToArray();
 
             var body = new {
                 NodesToAdd = _context.ConsumedOpcUaNodes.First().Value.OpcNodes.Select(node => new {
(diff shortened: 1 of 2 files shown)


tools/e2etesting/TestEventProcessor/TestEventProcessor.Businesslogic/TelemetryValidator.cs
```diff
@@ -289,6 +289,12 @@ private Task Client_ProcessEventAsync(ProcessEventArgs arg)
                     continue;
                 }
 
+                // OPC PLC contains bad fast and slow nodes that drop messages by design.
+                // We will ignore entries that do not have a value.
+                if (entryValue is null) {
```
koepalex, line 294: If we do not subscribe to bad nodes, that this should not be necessary right?

e2e-tests/IIoTPlatform-E2E-Tests/Orchestrated/B_PublishMultipleNodesOrchestratedTestTheory.cs
```diff
@@ -93,7 +93,11 @@ public class B_PublishMultipleNodesOrchestratedTestTheory {
 
             var cts = new CancellationTokenSource(TestConstants.MaxTestTimeoutMilliseconds);
             var testPlc = _context.SimulatedPublishedNodes[_context.ConsumedOpcUaNodes.First().Key];
-            _context.ConsumedOpcUaNodes.First().Value.OpcNodes = testPlc.OpcNodes.Skip(250).ToArray();
+
+            // We will filter out bad fast and slow nodes as they to drop messages by design.
```
monikavar: minor: omit "to"

tools/e2etesting/TestEventProcessor/TestEventProcessor.Businesslogic/TelemetryValidator.cs
monikavar: +1

karok2m: Well, yes, but that's if we assume that no-one will accidentally include them in a subscription again. And I would rather make sure that test event processor can handle those situations instead of hoping for that. So I will keep this check here.

e2e-tests/IIoTPlatform-E2E-Tests/Orchestrated/B_PublishMultipleNodesOrchestratedTestTheory.cs
karok2m: Removed.

karok2m:
Here is a run of E2E Orchestrated pipeline where you can see that the changes fixed publisher tests: https://msazure.visualstudio.com/One/_build/results?buildId=46192428&view=results

tools/e2etesting/TestEventProcessor/TestEventProcessor.Businesslogic/TelemetryValidator.cs
koepalex: ok, the first one that will really test BadNodeIds should add an setting to the configuration

koepalex (approved):

Issue #1060, opened 2021-03-11
publishednodes.json changes, log indicates that nothing has changed and nothing will be done, but system proceeds to process changes


derSchtefan:
**Describe the bug**
When the publishednodes.json changes, the message in the log indicates that a file change was detected but the content-hash is equal to last one, and nothing will be done. The system then however continues to correctly process the changes.

`[2021-03-11 19:04:46Z INF Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine.LegacyJobOrchestrator] File /appdata/publishednodes.json has changed and content-hash is equal to last one, nothing to do`

A full log can be found at the bottom. There  is also a message appearing that the contents of the file HAVE changed. It seems the second message is an artifact of some other change detector happening.

**To Reproduce**
1. Create a publishednodes.json
2. Start OPC Publisher
3. Change the json file and observe the two entries in the log

**Expected behavior**
Log message does not appear, or log message is amended to indicate that an earlier detection has already processed the change.

**Additional context**
Version is the current master branch dcd7a4a57d6e0e25936769f1496d687626d1f282.
Since there were quite some changes to the pn.json processing recently, it is probably a better approach to slightly reword the log message, than trying to suppress it .

**Impact**
We are monitoring the OPC Publisher logs for pn.json messages, and this triggered a monitoring rule that something went wrong updating the pn.json file. We can adjust our monitoring, but we are probably not the only ones running into this issue.

**Full log**

```
[2021-03-11 19:33:28Z INF Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine.LegacyJobOrchestrator] File /appdata/publishednodes.json has changed, last known hash 83B87FEE198B7A9DF9EEB354BE3FA734F6FF32B4CF825FC80A8E33FC0530E7BF, new hash E4B1F1E3552FEE2972BB99F27AE856088ECD63DB0A7781B433C9C9FD4B478F45, reloading...
[2021-03-11 19:33:28Z INF Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Models.PublishedNodesJobConverter] Read 2 items from published nodes file in 00:00:00.0130894
[2021-03-11 19:33:28Z INF Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Models.PublishedNodesJobConverter] Converted items to jobs in 00:00:00.0133206
[...]

PR #1371, opened 2021-08-31
avoid double events from FileSystemWatcher

dacolgit:
fix GitHub issue: [https://github.com/Azure/Industrial-IoT/issues/1060](https://github.com/Azure/Industrial-IoT/issues/1060)

ADO Tracking Item: [https://msazure.visualstudio.com/One/_workitems/edit/10737287](https://msazure.visualstudio.com/One/_workitems/edit/10737287)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -171,7 +171,12 @@ public class LegacyJobOrchestrator : IJobOrchestrator {
                 }
             }
             else {
-                RefreshJobFromFile();
+                //avoid double events from FileSystemWatcher
+                DateTime lastWriteTime = File.GetLastWriteTime(_legacyCliModel.PublishedNodesFile);
+                if (lastWriteTime.Ticks - _lastRead.Ticks > _interval10ms) {
```
karok2m: Can we please use difference of `DateTime` objects and compare that with a `TimeSpan` object, something like `TimeSpan.FromMilliseconds(10)` ? Using `Ticks` is quite confusing as they represent hundred nanoseconds, so having interval of `10ms` be equal to `100000` looks strange because of that.

karok2m: Also, what if the file has actually changed within last `10ms` ? Then we would ignore those changes. I would recommend to ignore refresh only if we are sure that file hash is the same.

dacolgit: changed using `TimeSpan.FromMilliseconds(10)` instead of ticks.
If I just ignore refresh when the file hash are the same I miss the event when you re-save the file for example (file overwritten with same hash). 
You are right that some events could be missed in theory, but I don't think we have two file changes in 10 ms

karok2m: @dacolgit, I still think that the code in the form that it is now is opening a possibility for a bug where we ignore a legitimate file change. And the problem with that bug will be that it will be super difficult to reproduce or debug. So I would still suggest to move this logic to the `RefreshJobFromFile()` method itself. We have an `else` case there where we log that the file has been changes but the hash is the same so there is nothing to do. And I would argue that that is the place where the check that you have here should happen.

koepalex: @dacolgit I agree with @karok2m we should move the code into the method `RefreshJobFromFile` in the case were we know that the hashes are identically. And I would increase the _debounce time_ to 10 seconds.

dacolgit: I moved the logic to `RefreshJobFromFile`.
Regarding increasing the debounce time, the duplicated event is fired just few ticks after the first one. This is a known issue of  FileSystemWatcher. Increasing the debounce to 10 seconds can lead to missing some entries in the log (for example if you re-save publishednodes.json with no changes within a 10 seconds interval)

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -258,8 +254,12 @@ public class LegacyJobOrchestrator : IJobOrchestrator {
                         _availableJobs = availableJobs;
                         _assignedJobs.Clear();
                     } else {
-                        _logger.Information("File {publishedNodesFile} has changed and content-hash is equal to last one, nothing to do", _legacyCliModel.PublishedNodesFile);
+                        //avoid double events from FileSystemWatcher
+                        if (lastWriteTime - _lastRead > TimeSpan.FromMilliseconds(10)) {
```
koepalex, line 258: please change to 1 second

koepalex (approved):
:shipit:

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
koepalex: OK

Issue #1153, opened 2021-05-06
Cloud services producing lots of logs


maalliet:
Installed the all-in-one solution of the Industrial IoT Cloud services.
The API services seem to produce a lot of logs - majority of them are informational, therefore costs increasing quite significantly.
Is there a way to chance the loglevel to warning without needing to rebuild the solution?

maalliet:
Hi @hansgschossmann 

From my point of view it is partially related. the LogFiles folder is on my side also increasing in size over time. However, I do not receive any error message in the logs. 
Underlying services are producing similar informational logs over and over again (see small sample below). Would be great if the loglevel of the application could be configured to warning or above. This will decrease the amount of logs significantly, although it does not resolve the increasing LogFiles directory size in the long term.

```
2021-05-17T21:39:57.458782466Z [21:39:57 INF Microsoft.Azure.IIoT.Agent.Framework.Jobs.JobOrchestratorEndpointSync] Updating orchestrator urls... 
2021-05-17T21:39:57.657452604Z [21:39:57 INF Microsoft.Azure.IIoT.Agent.Framework.Jobs.JobOrchestratorEndpointSync] Job orchestrator url updated to https://APISERVICENAME.azurewebsites.net/edge/publisher on all twins. 
2021-05-17T21:39:57.658315204Z [21:39:57 INF Microsoft.Azure.IIoT.Agent.Framework.Jobs.JobOrchestratorEndpointSync] Orchestrator Url update finished. 

2021-05-17T21:40:05.223900792Z [21:40:05 INF Microsoft.Azure.IIoT.Auth.IoTHub.TwinIdentityTokenUpdater] Updating identity tokens... 
2021-05-17T21:40:05.660330633Z [21:40:05 INF Microsoft.Azure.IIoT.Hub.Client.IoTHubServiceHttpClient] IOTEDGENAME (publisher) created or updated (AAAAAAAAVCY= -> AAAAAAAAVCc=) 
2021-05-17T21:40:05.779117681Z [21:40:05 INF Microsoft.Azure.IIoT.Auth.IoTHub.TwinIdentityTokenUpdater] Identity Token update finished. 

2021-05-17T21:40:57.659263260Z [21:40:57 INF Microsoft.Azure.IIoT.Agent.Framework.Jobs.JobOrchestratorEndpointSync] Updating orchestrator urls... 
[...]

PR #1372, opened 2021-09-01
LogLevel setting for Application Insights (#1153)

maalliet:
A PCS variable 'PCS_APPINSIGHTS_LOGLEVEL' is added to configure the loglevel of the Application Insights logger.
If the variable is not set or if naming does not match a LogEventLevel value (Fatal, Warning, Error,...) it will be defaulted to 'Information'


koepalex (approved):
Thank you for your contribution

Issue #1303, opened 2021-07-23
Error when deploying the version 2.8 via Microsoft.Azure.IIoT.Deployment


emulic:
Deployment aborts during the deployment step jumpbox-vm-setup with the following error:
'Long running operation failed with status 'Failed'. Additional Info:'At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/DeployOperations for usage details.''
I found following error description in the Azure portal for this deployment step:
[...]

PR #1383, opened 2021-09-22
 IAI: Added note about removing values with null from appsettings.json.

karok2m:
Changes:
* IAI: Fixed fallback mechanism to defaults for Helm settings.
* IAI: Added note about removing values with `null` from `appsettings.json`.


karok2m:
Fixes #1303

cristipogacean (approved):
:shipit:

PR #1386, opened 2021-09-27
Updated dependencies and remove projects that are not built

marcschier:
* Update dependencies and remove projects that are not built to not show up in CG.
* Update code to work with latest OPC UA stack.
* Unit and integration tests are passing locally.
* Update build system (under tools) to be the same as in main.


mregen (dismissed):

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher.csproj
```diff
@@ -29,13 +29,16 @@
...
-    <PackageReference Include="Microsoft.Azure.Devices" Version="1.18.2" />
-    <PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.21.2" />
-    <PackageReference Include="Mono.Options" Version="5.3.0.1" />
-    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
-    <PackageReference Include="OPCFoundation.NetStandard.Opc.Ua.Symbols" Version="1.4.357.28" />
+    <PackageReference Include="Microsoft.Azure.Devices" Version="1.35.0" />
+    <PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.38.0" />
+    <PackageReference Include="Mono.Options" Version="6.12.0.148" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Include="OPCFoundation.NetStandard.Opc.Ua.Symbols" Version="1.4.366.38" />
```
mregen: >OPCFoundation.NetStandard.Opc.Ua.Symbols [](http://example.com/codeflow?start=31&length=40)

better only use the OPCFoundation.NetStandard.Opc.Ua.Client lib here
to avoid the kestrel dependency

mregen (changes requested):
sorry, I found one more thing... if possible switch to the seperate ua libraries

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher.csproj
marcschier: I found that symbols brings in bouncy castle, which was used in SecureIoTHubToken.  I removed the entire code which we do not need in IoT Edge context anyway, and that is the only supported runtime context.

cristipogacean (approved):

mregen (approved):

hansgschossmann (approved):

PR #1418, opened 2021-09-30
Moving E2E pipelines to iotcat Azure DevOps organizatoin.

karok2m:
Moving E2E orchestrated and standalone pipelines to run in `iotcat` organization. The reason for the move is that for our integration tests we need custom build agents and in `One` organization those custom agents are no longer allowed.

Changes:
- Triggering of both E2E pipelines is now done via a REST API call since pipeline completion triggers that were used previously work only for pipeline within the same organization.
- Fixed paths of some `csproj` files.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -59,3 +59,5 @@ stages:
   - build
   jobs:
   - template: tools/templates/e2e_tests.yml
+    parameters:
+      branchName: ${{ replace(variables['Build.SourceBranch'], 'refs/heads/', '') }}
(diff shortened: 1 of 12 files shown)


azure-pipelines.yml
```diff
@@ -52,3 +52,12 @@ stages:
   - build
   jobs:
   - template: tools/templates/cc.yml
+- stage: e2e_tests
+  displayName: 'Triggering E2E tests'
+  condition: and(succeeded(), ne(variables['Build.Reason'], 'PullRequest'))
+  dependsOn:
+  - build
```
karok2m, line 59: E2E pipelines should be triggered only if all the above steps have passed and not only `build`.

karok2m: Added dependency on all previous stages.

cristipogacean (approved):

PR #1422, opened 2021-10-08
Modified publisher standalone E2E tests to also run for PubSub encoding.

karok2m:
Modified publisher standalone E2E tests to also run for `PubSub` encoding.


karok2m:
`PubSub` tests are currently disabled as we need `TestEventProcessor` to support `PubSub` encoding first.

karok2m:
`PubSub` tests are now enabled as support for processing `PubSub` messages in `TestEventProcessor` has been added in #1424.

cristipogacean (approved):
:shipit:

PR #1423, opened 2021-10-11
Improve encoding for events

simonjaeger:
Remove redundant fields (ServerTimestamp and SourceTimestamp). If wanted, they can be included in the SelectClauses to be part of the payload instead. Aligns better with specification (as there is no ServerTimestamp or SourceTimestamp on BaseEventType). 

The specification for BaseEventType is available [here](https://reference.opcfoundation.org/v104/Core/docs/Part5/6.4.2/). If Time and ReceiveTime is wanted, it should be retrieved with the appropriate SelectClauses (see [spec](https://reference.opcfoundation.org/v104/Core/docs/Part4/7.17.3/)) instead (and not mandatory by the OPC Publisher). For example:

```json
[
    {
        "EndpointUrl": "...",
        "UseSecurity": false,
        "OpcEvents": [
            {
                "Id": "ns=0;i=2253",
                "DisplayName": "...",
                "SelectClauses": [
                    {
                        "TypeDefinitionId": "i=2041",
                        "BrowsePath": [
                            "EventId"
                        ]
                    },
                    {
                        "TypeDefinitionId": "i=2041",
                        "BrowsePath": [
                            "Message"
                        ]
                    },
                    {
                        "TypeDefinitionId": "i=2041",
                        "BrowsePath": [
                            "Time"
                        ]
                    },
                    {
                        "TypeDefinitionId": "i=2041",
                        "BrowsePath": [
                            "ReceiveTime"
                        ]
                    }
                ],
                "WhereClause": {
                    "Elements": [
                        {
                            "FilterOperator": "OfType",
                            "FilterOperands": [
                                {
                                    "Value": "ns=2;i=235"
                                }
                            ]
                        }
                    ]
                }
            }
        ]
    }
]
```

PR #1427, opened 2021-10-15
Fix E2E tests for events after encoding change

simonjaeger:
Previous PR made a slight change to the encoding of events to align better with specification. See [here](https://github.com/Azure/Industrial-IoT/pull/1423). The E2E tests need to be changed to support this.


peterbryntesson (approved):
LGTM

karok2m (approved):

PR #1435, opened 2021-10-21
SignalR: removed deprecated calls.

karok2m:
Changes:
* Upgraded to `1.11.0` version of `Microsoft.Azure.SignalR.Management` NuGet package.
* Removed deprecated calls of `ServiceManagerBuilder` and changed the calls according to [`IServiceManager` to `ServiceManager` Migration Guidance](https://github.com/Azure/azure-signalr/blob/dev/docs/management-sdk-migration.md).


cristipogacean (approved):

karok2m:
I've run the following manual test on a deployment from this branch to validate that data is flowing through SignalR:
* Enable discovery for an asset. I could see that discovery results were shown on the page.
* Enabled publishing of several nodes. I could see that real-time values were continuously delivered to Engineering Tool.

PR #1447, opened 2021-11-02
Updated third party notice.

karok2m:
Changes:
* Updated third party notice.
* Manually added license information for `OPCFoundation` packages.


dacolgit (dismissed):
Because most of the licenses are of type 'Apache' or 'Mit', instead of reporting the full license for each package we can think to just  add the full license once at the end.

cristipogacean (dismissed):

karok2m:
> Because most of the licenses are of type 'Apache' or 'Mit', instead of reporting the full license for each package we can think to just add the full license once at the end.

The file is not generated manually but by a 3rd party notice generation step in our `SDL` job. So in general we should not manually modify it and keep as it is. For `OPCFoundation` packages I had to manually add the license terms as the tool was not able to add it automatically.

cristipogacean (approved):
:shipit:

dacolgit (approved):

marcschier (approved):

monikavar (approved):

PR #1454, opened 2021-11-04
Update release announcement for 2.8.1 release

monikavar:


docs/release-announcement.md
```diff
@@ -1,11 +1,17 @@
 # Release announcement
 
-## Azure Industrial IoT Platform 2.8 Release
+## Azure Industrial IoT Platform Release 2.8.1
+We are pleased to announce the release of version 2.8.1 of our Industrial IoT Platform components as the first patch update of the 2.8 Long-Term Support (LTS) release. This release contains important security updates, bugfixes and performance optimizations.
 
-We are pleased to announce the release of version 2.8.0 of our Industrial IoT Platform as well as the declaration of Long-Term Support (LTS) for this version.
+> IMPORTANT
+> Please note that OPC Publisher 2.8.1 is not backwards compatible with version 2.5.x. This will be addressed in the next patch release.
```
karok2m: Please remove `This will be addressed in the next patch release.`. Even though this is our plan at the moment, that might change and so should not commit to it in release announcements.

docs/release-announcement.md
```diff
@@ -1,11 +1,17 @@
...
+
+## Azure Industrial IoT Platform Release 2.8
+
+We are pleased to announce the release of version 2.8 of our Industrial IoT Platform as well as the declaration of Long-Term Support (LTS) for this version.
```
karok2m, line 11: Let's keep `2.8.0` versioning in the body of the announcement for `2.8` as this talks specifically about `2.8.0` release.

docs/release-announcement.md
```diff
@@ -1,11 +1,17 @@
...
 While we continue to develop and release updates to our ongoing projects on GitHub, we now also offer a branch that will only get critical bug fixes and security updates starting in July 2021. Customers can rely upon a longer-term support lifecycle for these LTS builds, providing stability and assurance for the planning on longer time horizons our customers require. The LTS branch offers customers a guarantee that they will benefit from any necessary security or critical bug fixes with minimal impact to their deployments and module interactions. At the same time, customers can access the latest updates in the main branch to keep pace with the latest developments and fastest cycle time for product updates.
 
 > IMPORTANT
-> We suggest to update from the version 2.6 or later to ensure secure operations of your deployment. 2.8.0 is not backwards compatible with version 2.5.
+> We suggest to update from the version 2.6 or later to ensure secure operations of your deployment. 2.8 is not backwards compatible with version 2.5.
```
karok2m: Same comment as above, let's keep `2.8.0` versioning instead of just `2.8` here as this announcement is specifically about `2.8.0`.

docs/release-announcement.md
```diff
@@ -1,11 +1,17 @@
...
 
-Version 2.8.0 includes an updated version of the IoT Edge Runtime, a new Linux base image for all Linux deployments, and several bug fixes. The detailed changes can be found [here](https://github.com/Azure/Industrial-IoT/releases/tag/2.8.0).
+Version 2.8 includes an updated version of the IoT Edge Runtime, a new Linux base image for all Linux deployments, and several bug fixes. The detailed changes can be found [here](https://github.com/Azure/Industrial-IoT/releases/tag/2.8.0).
```
karok2m: Same comment as above, let's keep `2.8.0` versioning instead of just `2.8` here as this announcement is specifically about `2.8.0`.

karok2m (changes requested):
See comments above.

docs/release-announcement.md
monikavar: Didn't we agree that we will use "2.8" for 2.8.0 release. Additionally, convention wise, the number after second decimal denotes the patch version, so it would make sense to use 2.8.1 for first patch version and only 2.8 for LTS release.
@cristipogacean, thoughts?

docs/release-announcement.md
monikavar: Same as above.

docs/release-announcement.md
marcschier: FWIW, I think the way it shows with 2.8 in the bottom, and 2.8.1 above works.

karok2m: @monikavar, I just do not want us to retroactively change our previous release announcements. That announcement was talking about `2.8` LTS release in general, but specifically about `2.8.0` version. So I don't think that we should change what was already there.

docs/release-announcement.md
karok2m: This one specifically talks about changes introduced in `2.8.0` and links to that specific tag. In my opinion this one should be reverted.

docs/release-announcement.md
karok2m: Here we talk specifically about `2.8.0` version. Future versions might (and should) be compatible with `2.5`, so this will not be generally true for `2.8` (current and future) version. I think that this one should also be reverted.

docs/release-announcement.md
karok2m: The change in this part is OK. Resolving this comment.

docs/release-announcement.md
monikavar: Reverted.

docs/release-announcement.md
monikavar: Reverted.

karok2m (approved):

PR #1456, opened 2021-11-05
[IAI] Fix version tag update in IAI deployment.

dacolgit:
Bug: [IAI deployment tags are outdated](https://msazure.visualstudio.com/One/_workitems/edit/12451621)

### DIFF
--- a/deploy/src/Microsoft.Azure.IIoT.Deployment/Configuration/HelmSettings.cs
+++ b/deploy/src/Microsoft.Azure.IIoT.Deployment/Configuration/HelmSettings.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Azure.IIoT.Deployment.Configuration {
         // Defaults
         public static string _defaultRepoUrl = "https://microsoft.github.io/charts/repo";
         public static string _defaultChartVersion = "0.3.2";
-        public static string _defaultImageTag = "2.7.206";
+        public static string _defaultImageTag = "2.8.1";
         public static string _defaultImageNamespace = "";
         public static string _defaultContainerRegistryServer = "mcr.microsoft.com";
         public static string _defaultContainerRegistryUsername = "";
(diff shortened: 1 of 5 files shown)


deploy/src/Microsoft.Azure.IIoT.Deployment/Configuration/HelmSettings.cs
```diff
@@ -13,7 +13,7 @@ class HelmSettings {
         // Defaults
         public static string _defaultRepoUrl = "https://microsoft.github.io/charts/repo";
         public static string _defaultChartVersion = "0.3.2";
-        public static string _defaultImageTag = "2.7.206";
```
karok2m: Please revert this back to `2.7.206`. Updating IAI for `2.8.1` release entails not only change of image tag and chart version but also cleanup of resources that were deployed for microservices that are removed from the deployment. And we have a separate PBI for doing all of that.

deploy/src/Microsoft.Azure.IIoT.Deployment/appsettings.json
```diff
@@ -112,7 +112,7 @@
     // azure-industrial-iot Helm chart version
     "ChartVersion": "0.3.2",
     // Azure IIoT components image tag
-    "ImageTag": "2.7.206",
```
karok2m: Please revert this back to `2.7.206`. Updating IAI for `2.8.1` release entails not only change of image tag and chart version but also cleanup of resources that were deployed for microservices that are removed from the deployment. And we have a separate PBI for doing all of that.

karok2m (changes requested):
Please revert changes for default image tag version. That change along with change of hem chart version and cleanup of resources that are no longer needed will be tackled in a separate PR. Everything else looks good.

deploy/src/Microsoft.Azure.IIoT.Deployment/appsettings.json
dacolgit: reverted

deploy/src/Microsoft.Azure.IIoT.Deployment/Configuration/HelmSettings.cs
dacolgit: reverted

karok2m (approved):

Issue #1385, opened 2021-09-24
Missing EndpointUrl in telemetry format


alevincenzi:
**Describe the bug**

With opcuaPublsher:2.8.0, the telemetry output (json) does not match [the online documentation](https://docs.microsoft.com/en-us/azure/industrial-iot/reference-opc-publisher-telemetry-format).

Moreover, changes in the command line option `--mm` has no effect on the telemetry messages.

**To Reproduce**

* Run a plc simulator with `docker run --rm -it -p 50000:50000 -p 8086:8086 --name opcplc mcr.microsoft.com/iotedge/opc-plc:latest --pn=50000 --autoaccept --sph --sn=5 --sr=10 --st=uint --fn=5 --fr=1 --ft=uint --ctb --scn --lid --lsn --ref--gn=5`
* Deploy the opcuaPublisher as explained in [this tutorial](https://docs.microsoft.com/en-us/azure/industrial-iot/tutorial-publisher-deploy-opc-publisher-standalone)
* Don't use the option `--mm`!
* Add routing to `$upstream`
* Configure the `pn.json` file with some nodes:

```
[
  {
    "EndpointUrl": "opc.tcp://10.1.0.2:50000",
    "UseSecurity": false,
    "OpcNodes": [
      { "Id": "ns=2;s=StepUp" },
      { "Id": "ns=2;s=AlternatingBoolean" },
      { "Id": "ns=2;s=RandomSignedInt32" },
      { "Id": "ns=2;s=RandomUnsignedInt32" }
    ]
  }
]
```

* Adapt the ip address according to the location of the plc simulator.

The telemetry messages will show up as follow

```
[
	{
		"NodeId": "http://microsoft.com/Opc/OpcPlc/#s=StepUp",
		"ApplicationUri": "urn:OpcPlc:b0803e843ad2",
		"Value": {
			"Value": 560934,
			"SourceTimestamp": "2021-09-24T05:56:12.202727Z"
		}
	},
	{
		"NodeId": "http://microsoft.com/Opc/OpcPlc/#s=RandomSignedInt32",
		"ApplicationUri": "urn:OpcPlc:b0803e843ad2",
		"Value": {
			"Value": -1001090625,
			"SourceTimestamp": "2021-09-24T05:56:12.2029103Z"
		}
	},
        ...
]
```

**Expected behavior**

[...]

PR #1462, opened 2021-11-11
fix NodeId telemetry format issue

cristipogacean:
* always use in telemetry the very same nodeId format as configured by the user
 * small refactoring & optimizations
 * fix for #1385

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Storage/PublishedNodesJobConverterTests.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Storage/PublishedNodesJobConverterTests.cs
@@ -231,9 +231,15 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Storage.Tests {
             Assert.Equal("testfieldid1", jobs
                 .Single().WriterGroup.DataSetWriters
                 .Single().DataSet.DataSetSource.PublishedVariables.PublishedData.First().Id);
+            Assert.Equal("i=2258", jobs
+                .Single().WriterGroup.DataSetWriters
+                .Single().DataSet.DataSetSource.PublishedVariables.PublishedData.First().PublishedVariableNodeId);
             Assert.Equal(null, jobs
                 .Single().WriterGroup.DataSetWriters
                 .Single().DataSet.DataSetSource.PublishedVariables.PublishedData.Last().Id);
+            Assert.Equal("i=2259", jobs
+                .Single().WriterGroup.DataSetWriters
+                .Single().DataSet.DataSetSource.PublishedVariables.PublishedData.Last().PublishedVariableNodeId);
             Assert.Equal("testgroup", jobs
                 .Single().WriterGroup.DataSetWriters
                 .Single().DataSet.DataSetSource.Connection.Group);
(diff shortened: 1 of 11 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Storage/PublishedNodesJobConverterTests.cs
```diff
@@ -231,7 +231,7 @@ private class StandaloneIdentity : IIdentity {
             Assert.Equal("testfieldid1", jobs
                 .Single().WriterGroup.DataSetWriters
                 .Single().DataSet.DataSetSource.PublishedVariables.PublishedData.First().Id);
-            Assert.Equal("i=2259", jobs
+            Assert.Equal(null, jobs
```
karok2m, line 237: Doesn't this invalidate the test itself ? Can we please check that `i=2258` and `i=2259` nodes are still present in `jobs` ?

cristipogacean: yes, that would make sense

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Storage/PublishedNodesJobConverterTests.cs
```diff
@@ -434,7 +434,7 @@ private class StandaloneIdentity : IIdentity {
 
             Assert.NotEmpty(jobs);
             Assert.Single(jobs);
-            Assert.Equal("testdisplayname1", jobs
+            Assert.Equal("testdatasetfieldid1", jobs
```
karok2m, line 443: @cristipogacean, based on the original and changed values it seems like values from `DataSetFieldId`, `DisplayName` and `Id` fields can all end up in `PublishedData.Single().Id`. What is the current priority order for those (and others if there are more) fields ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Storage/PublishedNodesJobConverterTests.cs
```diff
@@ -406,7 +412,7 @@ private class StandaloneIdentity : IIdentity {
 
             Assert.NotEmpty(jobs);
             Assert.Single(jobs);
-            Assert.Equal("i=2258", jobs
+            Assert.Equal(null, jobs
```
karok2m, line 415: Same comment as above, don't we expect that `i=2258` will be present in `jobs` ? Can we please check that it is still there ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Storage/PublishedNodesJobConverterTests.cs
cristipogacean: the priority order was DataSetFieldId / DisplayName / NodeId. This was actually defined when telemetry message was created, not when it was configured, therefore the test itself was incorrect. Now I refactored both to match the priority above.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Storage/PublishedNodesJobConverterTests.cs
cristipogacean: I changed the logic a bit so that Id stores DataSetFieldId or DisplayName if any is configured. The NodeId itself is stored into PublishedVariableNodeId. That's why here the test is to validate that the Id is null.

karok2m (approved):

karok2m:
@cristipogacean, can you please also run E2E tests to make sure that this change is not causing any issues there ?

PR #1305, opened 2021-07-29
Add Json Schema Validation for publishednodes.json file

WilliamBerryiii:
# Description
This PR begins the process of adding JSON-Schema based validation for the Azure OPC Publisher's [`publishednodes.json`](https://raw.githubusercontent.com/Azure/iot-edge-opc-publisher/main/opcpublisher/publishednodes.json) configuration file. A schema generation tool, [the OPC Publisher Schema Validator](https://github.com/WilliamBerryiii/opcpublisherschemavalidator), was created to programmatically generate a json-schema file that adheres to the three (3) supported (documented) formats for OPC UA node configuration within the [`publishednodes.json`](https://raw.githubusercontent.com/Azure/iot-edge-opc-publisher/main/opcpublisher/publishednodes.json) file.   The "default" generated schema, included in this PR, enforces not only proper object construction within the configuration file, but also validates property values such as NodeId, EndpointUrls, etc. (further details below)

Fixes #1301 
Partially addressed #1032 

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

## How Has This Been Tested?

This PR includes significant changes to existing OPC Publisher test code, including additions for 'schema validation error' validation, and ensuring that all pre-existing tests were/are exercised in accordance with their original conditional intent.

Further, the schema generator too covers both positive and negative case testing for the schema generation process as well as the output schema file (subsequently included in this PR).  Generated schema tests in the tooling can be found in the [generator tests](https://github.com/WilliamBerryiii/opcpublisherschemavalidator/tree/main/test).

## Contributors

@WilliamBerryiii - Developer of [OPC Publisher Schema Generator/Validator](https://github.com/WilliamBerryiii/opcpublisherschemavalidator) tool
@suneetnangia - Implemented all code changes within this PR

## Dependencies

This contribution adds an additional package dependency:

[JsonSchema.Net](https://github.com/gregsdennis/json-everything) @ Version 1.10.7

## Checklist

- [X] We have performed a self/peer-review of contributed code
- [X] We have commented code, particularly in hard-to-understand areas and updated method parameter descriptions
- [X] We have made corresponding changes to the documentation
- [X] Our changes generate no new warnings
- [X] We have added tests that prove the change is effective and works
- [X] New and existing unit tests pass locally with all changes

## Notes for Reviewers/Maintainers

There are a few points of consideration for this PR that would be useful to discuss:

[...]

PR #1464, opened 2021-11-12
temporarily disable default pn.json schema validation 

cristipogacean:
Temporarily disable default pn.json schema validation because of backwards compatibility issues:
* the validation schema is too restrictive with non-standard OPC UA namespace uris like the ones used by Kepserver. Even though the namespace uri definition is well defined in the OPC UA Spec, we find many opc ua servers in the field that use non-compliant uri strings. E.g.  "nsu=KEPServerEX;s=Sim.xxx". 
* the schema validation does not support NodeID in the IIoT Platform (twin defined) ExpandedNodeId format: "Namespaceuri#IdentifierUnion" e.g. `"Id": "http://test.org/UA/Data/#i=10847",`
* ExpandedNodeId is missing from OpcNode definition

Validation can enabled by specifying the right file using --pfs(publishfileschema) argument

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
@@ -202,7 +202,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine {
 
                             try {
                                 if (!File.Exists(_legacyCliModel.PublishedNodesSchemaFile)) {
-                                    _logger.Warning("File {PublishedNodesSchemaFile} does not exist, ignoring schema validation of {publishedNodesFile} file...",
+                                    _logger.Information("Validation schema file {PublishedNodesSchemaFile} does not exist or is disabled, ignoring validation of {publishedNodesFile} file...",
                                     _legacyCliModel.PublishedNodesSchemaFile, _legacyCliModel.PublishedNodesFile);
 
                                     jobs = _publishedNodesJobConverter.Read(fileReader, null, _legacyCliModel);
--- a/modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliConfigKeys.cs
+++ b/modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliConfigKeys.cs
@@ -20,7 +20,7 @@ namespace Microsoft.Azure.IIoT.Modules.OpcUa.Publisher.Runtime {
         /// <summary>
         /// Key for default published nodes schema file.
         /// </summary>
-        public const string DefaultPublishedNodesSchemaFilename = "schemas/publishednodesschema.json";
+        public const string DefaultPublishedNodesSchemaFilename = "";
 
         /// <summary>
         /// Key for the publisher site.
--- a/modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliOptions.cs
+++ b/modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliOptions.cs
@@ -53,7 +53,7 @@ namespace Microsoft.Azure.IIoT.Modules.OpcUa.Publisher.Runtime {
                     // Publisher configuration options
                     { "pf|publishfile=", "The filename to configure the nodes to publish.",
                         s => this[LegacyCliConfigKeys.PublishedNodesConfigurationFilename] = s },
-                    { "pfs|publishfileschema=", "The validation schema filename for publish file.",
+                    { "pfs|publishfileschema=", "The validation schema filename for publish file. Disabled by default.",
                         s => this[LegacyCliConfigKeys.PublishedNodesConfigurationSchemaFilename] = s },
                     { "s|site=", "The site OPC Publisher is working in.",
                         s => this[LegacyCliConfigKeys.PublisherSite] = s },


karok2m:
Should we open a GitHub Issue for this to add support of those formats in the validation scheme and keep track of the work ?

Tagging @suneetnangia and @WilliamBerryiii as they originally implemented this feature in #1305.

karok2m (approved):

suneetnangia:
> Should we open a GitHub Issue for this to add support of those formats in the validation scheme and keep track of the work ?
> 
> Tagging @suneetnangia and @WilliamBerryiii as they originally implemented this feature in #1305.



> Should we open a GitHub Issue for this to add support of those formats in the validation scheme and keep track of the work ?
> 
> Tagging @suneetnangia and @WilliamBerryiii as they originally implemented this feature in #1305.

@karok2m Agreed, we should open an issue and track it, not using schema validation should be exception than a norm, although from backward compatibility perspective we could just make validation disabled by default. Currently, if we want to bypass schema validation, we can do that explicitly by providing no value to --pfs param e.g. --pfs="" which also throws the warning to surface the fact that validation is skipped. If customer has something specific to themselves, they can add in the schema file to avoid validation error and warning and continue to take benefit of the schema validation, if this is more common than specific customer, we should/could add it in the schema. KepServer one above looks like falling in the latter case.

dacolgit (approved):

monikavar (approved):

WilliamBerryiii:
@cristipogacean - can you give some specifics about how you see stuff failing?

- the validation schema is too restrictive with non-standard OPC UA namespace uris like the ones used by Kepserver. Even though the namespace uri definition is well defined in the OPC UA Spec, we find many opc ua servers in the field that use non-compliant uri strings. E.g. "nsu=KEPServerEX;s=Sim.xxx".

If you have a recommendation for how far we should pull back from validating that the NSU is a valid URL/URN - I'm happy to update the schema. As an aside, the value validation for the ID property can be disabled by removing the "pattern property" from the ID field of the schema. 

https://github.com/Azure/Industrial-IoT/blob/c7184f7cbd8b507933722cbd6e995f871ed19b86/modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Schemas/publishednodesschema.json#L47

- the schema validation does not support NodeID in the IIoT Platform (twin defined) ExpandedNodeId format: "Namespaceuri#IdentifierUnion" e.g. "Id": "http://test.org/UA/Data/#i=10847",

The schema does support "nsu", "ns" and "nodeid" formats (see above pattern match). If is not working, can you provide an example pn.json section that we can test against? 

- ExpandedNodeId is missing from OpcNode definition

ExpandedNodeId is supported in the second sub-schema definition (see this line: https://github.com/Azure/Industrial-IoT/blob/c7184f7cbd8b507933722cbd6e995f871ed19b86/modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Schemas/publishednodesschema.json#L114). Can you provide an example pn.json section that we can test against? 

Thanks!

cristipogacean:
@WilliamBerryiii, @suneetnangia 

First-of-all we really appreciate your contributions! The schema validation is a really valuable addition to the OPC Publisher and we're looking forward to fill in the existing gaps and integrate it as a default enabled product feature.

The pn.json that I had issues with is the following: 
```
[
	{
	"EndpointUrl": "opc.tcp://crpogace01:49320",
	"UseSecurity": false,
	"OpcNodes": [
			{
				
				"DisplayName": "Kep_1_DisplayName",
				"DataSetFieldId": "Kep_1_DataSetFieldId",
				"Id": "nsu=KEPServerEX;s=Sim.CH1.SIM_CH1_TAG1\\234754a-c63-b9601",
				"OpcSamplingInterval": 1000,
				"OpcPublishingInterval": 1000
			},
			{
				"DisplayName": "Kep_2_DisplayName",
				"DataSetFieldId": "Kep_2_DataSetFieldId",
				"Id": "ns=2;s=Sim.CH1.SIM_CH1_TAG10\\2347798-c63-19401",
				"ExpandedNodeId":"nsu=KEPServerEX;s=Sim.CH1.SIM_CH1_TAG10\\2347798-c63-19401",
				"OpcSamplingInterval": 1000,
				"OpcPublishingInterval": 1000
			}
		]
	},
	{
	"EndpointUrl": "opc.tcp://crpogace01:51210/UA/DemoServer",
	"UseSecurity": false,
	"OpcNodes": [
			{
				"DisplayName": "Softing_1_DisplayName",
				"DataSetFieldId": "Softing_1_DataSetFieldId",
				"Id": "http://test.org/UA/Data/#i=10847",
				"OpcSamplingInterval": 1000,
				"OpcPublishingInterval": 1000
			},
			{
				"DisplayName": "Softing_2_DisplayName",
				"DataSetFieldId": "Softing_2_DataSetFieldId",
				"Id": "nsu=http://test.org/UA/Data/;i=10848",
				"OpcSamplingInterval": 1000,
				"OpcPublishingInterval": 1000
			},
			{
				"DisplayName": "Softing_3_DisplayName",
				"DataSetFieldId": "Softing_3_DataSetFieldId",
				"Id": "ns=3;i=10849",
				"ExpandedNodeId":"http://test.org/UA/Data/#i=10849",
				"OpcSamplingInterval": 1000,
				"OpcPublishingInterval": 1000
			}
		]
	}
]
```

On my original points:

* **OPC UA namespace uris** - I believe we should just validate that namespace uri is a string. I have seen too many uncompliant forms uris in market established OPC UA servers that I believe the best strategy to avoid support cases here is to expect the unexpected. I've seen, besides well formatted uris, uris starting with uri.xxx, urn.xxx, random (application specific) strings like KEPServerEX, I have even seen empty strings for namespace uris in some server.
* **Namespaceuri#IdentifierUnion** - this is the format that we use in the IIoT platform cloud based microservices as well as in the opc twin module, therefore we need to support it in the pn.json file as well. 
* **ExpandedNodeId** - just based on a quick review I had the impression that `Id` and `ExpandedNodeId` are mutual exclusive in the. Maybe I was wrong about this. Please find above the example I tested with. 

Thank you very much!

WilliamBerryiii:
@cristipogacean - thanks for the sample and I'll take a look at this today.  

I don't believe the project's unit tests or configuration files we've run across included any examples where `Id` and `ExpandedNodeId` were used in the same OpcNodes configuration set; so if the mixed use of `Id` and `ExpandedNodeId` is a real example (e.g. a config file that spanned multiple generations of OPC Publisher), then I'd propose we should: 

- update the schema to not be mutually exclusive arrays for `Id` and `ExpandedNodeId` - relatively trivial change. I intentionally kept them discrete to reflect the [generational config file comments](https://github.com/Azure/Industrial-IoT/blob/c7184f7cbd8b507933722cbd6e995f871ed19b86/docs/modules/publishednodes.json#L50) in the sample pn.json file
- add supporting unit tests to the [schema generator/validator](https://github.com/WilliamBerryiii/opcpublisherschemavalidator) to document the mixed use of `Id` and `ExpandedNodeId` as an operational requirement   
- add supporting unit tests to OPC Publisher to document the mixed use of `Id` and `ExpandedNodeId` as an operational requirement  
- update the example pn.json example file to note that it's permissible to mix them (not sure if this would encourage folks ... so maybe skip it?)
- back off the value pattern matches to just a string type match (we can do this, regardless, by simply removing the pattern element from the schema)

I can get you a PR pretty quickly on this, if you'd like help.

Few other points to note:

- We should absolutely prioritize not breaking any existing implementations of users; and it would be good to document these cases via unit tests so that future changes to (de)serialization routines don't accidentally break folks.  
- The [schema generator/validator](https://github.com/WilliamBerryiii/opcpublisherschemavalidator) can be used in local dev loops to update/upgrade existing config flies. The generator/validator uses AJV as the backing schema utility and can be run either as a local toolchain or in the context of a CI/CD pipeline to "upgrade" or enforce consistency of config files.
- I'd love some thoughts on if we can improve the project's documentation around config file format upgrading and how the schema can be leveraged to improve the quality of existing user's config files. I.E. would it help to have a tool built that could take old or mixed format pn.json config files, attempt an upgrade and show recommended changes for compatibility - like migration to NSU style ids?

Should we move the conversation to an issue?

Thanks again for the samples and happy to support dev cycles on this.

PR #1472, opened 2021-11-18
Diagnostic connection counters

peterbryntesson:
This PR adds three diagnostic counters per endpoint:
* Is the endpoint connection up and running?
* How many nodes have successfully connected and are receiving data?
* How many nodes have misconfiguration causing them to not connect?

Example of diagnostic output can be seen here:
image

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/WriterGroupMessageSource.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/WriterGroupMessageSource.cs
@@ -27,20 +27,17 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine {
         public string Id => _writerGroup.WriterGroupId;
 
         /// <inheritdoc/>
-        public int NumberOfConnectionRetries => _subscriptions?.FirstOrDefault()?.
-            Subscription?.NumberOfConnectionRetries ?? 0;
+        public int NumberOfConnectionRetries => _subscriptions?.Sum(x => x.Subscription?.NumberOfConnectionRetries) ?? 0;
 
         /// <inheritdoc/>
-        public bool IsConnectionOk => _subscriptions?.FirstOrDefault()?.
-            Subscription?.IsConnectionOk ?? false;
+        public bool IsConnectionOk => (_subscriptions?.Count == 0 || 
+            _subscriptions?.Where(x => x.Subscription?.IsConnectionOk == true).Count() < _subscriptions?.Count) ? false : true;
 
         /// <inheritdoc/>
-        public int NumberOfGoodNodes => _subscriptions?.FirstOrDefault()?.
-            Subscription?.NumberOfGoodNodes ?? 0;
+        public int NumberOfGoodNodes => _subscriptions?.Sum(x => x.Subscription?.NumberOfGoodNodes) ?? 0;
 
         /// <inheritdoc/>
-        public int NumberOfBadNodes => _subscriptions?.FirstOrDefault()?.
-            Subscription?.NumberOfBadNodes ?? 0;
+        public int NumberOfBadNodes => _subscriptions?.Sum(x => x.Subscription?.NumberOfBadNodes) ?? 0;
 
         /// <inheritdoc/>
         public ulong ValueChangesCountLastMinute {
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/WriterGroupMessageSource.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/WriterGroupMessageSource.cs
@@ -3,8 +3,7 @@
 //  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
 // ------------------------------------------------------------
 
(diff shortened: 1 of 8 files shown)


simonjaeger (dismissed):
LGTM

cristipogacean (dismissed):
nice improvement

karok2m:
@peterbryntesson, can you please add information about new diagnostics counters in `docs/manual/readme.md` ? There is a `IIoT Metrics` section about available metrics.

peterbryntesson:
> @peterbryntesson, can you please add information about new diagnostics counters in `docs/manual/readme.md` ? There is a `IIoT Metrics` section about available metrics.

Done

peterbryntesson:
Sorry, mistakenly closed the PR, but has reopened it.

cristipogacean (dismissed):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/WriterGroupMessageSource.cs
```diff
@@ -30,6 +30,18 @@ public class WriterGroupMessageTrigger : IMessageTrigger, IDisposable {
         public int NumberOfConnectionRetries => _subscriptions?.FirstOrDefault()?.
             Subscription?.NumberOfConnectionRetries ?? 0;
 
+        /// <inheritdoc/>
+        public bool IsConnectionOk => _subscriptions?.FirstOrDefault()?.
+            Subscription?.IsConnectionOk ?? false;
+
+        /// <inheritdoc/>
+        public int NumberOfGoodNodes => _subscriptions?.FirstOrDefault()?.
```
karok2m: @peterbryntesson, why are we getting `NumberOfGoodNodes` only from the first subscription and not summing over all of them ? I have the same question about all new counters in this class, as well as `NumberOfConnectionRetries` which I see was already present. Can we please test that if `WriterGroupMessageTrigger` has several subscriptions in it then the reported values would still be valid ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -64,6 +63,51 @@ public class DefaultSessionManager : ISessionManager, IDisposable {
...
+                }
+                return wrapper.State == SessionState.Running;
+            }
+            finally {
+                _lock.Release();
+            }
+        }
+
+        /// <inheritdoc/>
+        public int GetNumberOfGoodNodes(ConnectionModel connection) {
```
karok2m: @peterbryntesson, `GetNumberOfGoodNodes()` method is not used anywhere. Was it intended to be used in `SubscriptionWrapper` similar to `DefaultSessionManager.IsConnectionOk` ? Can we delete it if it is not referenced anywhere ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -64,6 +63,51 @@ public class DefaultSessionManager : ISessionManager, IDisposable {
...
+                }
+                return wrapper._subscriptions.Sum(x => x.Value.NumberOfGoodNodes);
+            }
+            finally {
+                _lock.Release();
+            }
+        }
+
+        /// <inheritdoc/>
+        public int GetNumberOfBadNodes(ConnectionModel connection) {
```
karok2m: @peterbryntesson, similar to the one above, `GetNumberOfBadNodes` method is not used anywhere. Was it intended to be used in `SubscriptionWrapper` similar to `DefaultSessionManager.IsConnectionOk` ? Can we delete it if it is not referenced anywhere ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/WriterGroupMessageSource.cs
peterbryntesson: I just followed the pattern with NumberOfConnectionRetries. Let me change that. I will change it for NumberOfConnectionRetries as well.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
peterbryntesson: Good spotting, Forgot to remove those

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
peterbryntesson: Good spotting, Forgot to remove those

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/WriterGroupMessageSource.cs
peterbryntesson: Done

karok2m (approved):

cristipogacean (approved):

PR #1478, opened 2021-11-23
Add DeviceAuthentication login option for Connect-AzAccount

ChenCheng368:
When the user's environment cannot prompting web browser for Connect-AzAccount login, the original code not able to catch that, no error popup but the output will be null.
An improvement here is to change line 75 -ErrorAction to -WarningAction to catch this case, and provide DeviceAuthentication login option for the user.

### DIFF
--- a/deploy/scripts/aad-register.ps1
+++ b/deploy/scripts/aad-register.ps1
@@ -72,7 +72,7 @@ Function Select-Context() {
     try {
         $azProfile = Connect-AzAccount `
             -Environment $environmentName @tenantIdArg `
-            -ErrorAction Stop
+            -WarningAction Stop
 
         $reply = Read-Host -Prompt "Save credentials in .user file? [y/n]"
         if ($reply -match "[yY]") {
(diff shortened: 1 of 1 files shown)


luiscantero (dismissed):

karok2m:
@ChenCheng368 , one question here, doesn't `Connect-AzAccount` use device authentication automatically when the platform doesn't support browser-based authentication ?

marcschier (dismissed):

ChenCheng368:
> @ChenCheng368 , one question here, doesn't `Connect-AzAccount` use device authentication automatically when the platform doesn't support browser-based authentication ?

@karok2m, No, "Connect-AzAccount" doesn't. While the interesting thing is "Connect-AzureAD" will use device authentication automatically.

deploy/scripts/aad-register.ps1
```diff
@@ -84,7 +84,26 @@ Function Select-Context() {
         return $context
     }
     catch {
-        throw "The login to the Azure account was not successful."
+        try {
+            # for the case where prompting web browser is not applicable, we will prompt for DeviceAuthentication login.
+            $azProfile = Connect-AzAccount `
+            -UseDeviceAuthentication `
+            -TenantId $script:tenantId `
```
karok2m: `-TenantId $script:tenantId` part might already be included by `@tenantIdArg` parameters below. It also should do the check for empty `$script:tenantId`. Please have a look.

karok2m:
> > @ChenCheng368 , one question here, doesn't `Connect-AzAccount` use device authentication automatically when the platform doesn't support browser-based authentication ?
> 
> @karok2m, No, "Connect-AzAccount" doesn't. While the interesting thing is "Connect-AzureAD" will use device authentication automatically.

@ChenCheng368, ah, OK. That clarifies it for me. Can you please have a look at the comment regarding `@tenantIdArg` ? Otherwise everything looks great. Thanks for your contribution.

deploy/scripts/aad-register.ps1
ChenCheng368: Yes, I agree, thanks for pointing out. I have edited again.

karok2m (approved):

karok2m:
Successful run of tests pipeline can be found [here](https://msazure.visualstudio.com/One/_build/results?buildId=49405675&view=results). Not sure why it is not linked automatically in the checks.

PR #1497, opened 2021-12-15
Reduce capabilities of edge module

dacolgit:
Because running edge modules as non root user is critical to the functionality of the service  the CAP_CHOWN and CAP_SETUID capabilities (provided by default to the root user in unprivileged containers) are removed via the --cap-drop within the container create options in the deployment manifest


docs/deploy/deployment-manifest.md
```diff
@@ -51,7 +51,7 @@ An example manifest for the released Industrial-IoT IoT Edge modules included in
             "restartPolicy": "always",
             "settings": {
               "image": "mcr.microsoft.com/iotedge/discovery:latest",
-              "createOptions": "{\"Hostname\":\"discovery\",\"NetworkingConfig\":{\"EndpointsConfig\":{\"host\":{}}},\"HostConfig\":{\"NetworkMode\":\"host\",\"CapAdd\":[\"NET_ADMIN\"]}}"
+              "createOptions": "{\"Hostname\":\"discovery\",\"NetworkingConfig\":{\"EndpointsConfig\":{\"host\":{}}},\"HostConfig\":{\"NetworkMode\":\"host\",\"CapAdd\":[\"NET_ADMIN\"],\"CapDrop\":[\"CHOWN\",\"SETUID\"]}"
```
karok2m: Should we also add this to the publisher and twin deployments bellow (in the same file) ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Twin/src/Deploy/IoTHubSupervisorDeployment.cs
```diff
@@ -94,6 +94,10 @@ public sealed class IoTHubSupervisorDeployment : IHostProcess {
                     HostConfig = new {
                         Binds = new[] {
                             "/mount:/mount"
+                        },
+                        CapDrop = new[] {
```
karok2m, line 98: What about windows deployment case ? Wouldn't those changes be applicable for that case as well ?

docs/deploy/deployment-manifest.md
dacolgit: added

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Twin/src/Deploy/IoTHubSupervisorDeployment.cs
dacolgit: These are Linux capabilities. they don't apply to Windows

karok2m: I was thinking here about IoT Edge runtime running through [EFLOW](https://docs.microsoft.com/en-us/azure/iot-edge/iot-edge-for-linux-on-windows?view=iotedge-2018-06&preserve-view=true), in which case containers will be effectively running on top of a Linux kernel where those capabilities would still be applicable. Though people might still be using IoT Edge on Windows version 1809 which does support Windows containers where I guess those capabilities will not be available.

karok2m: I will add a separate PBI to tackle this when we add support of EFLOW. For now this should be good for Windows containers.

dacolgit: As you said in EFLOW container will run in a Linux environment running on a VM in Windows. Then it's a Linux case

docs/deploy/howto-deploy-modules-portal.md
```diff
@@ -29,7 +29,8 @@ To deploy all required modules to the Gateway using the Azure Portal...
    As *create options* use the following JSON:
 
    ```json
-   {"NetworkingConfig":{"EndpointsConfig":{"host":{}}},"HostConfig":{"NetworkMode":"host","CapAdd":["NET_ADMIN"]}}
```
karok2m, line 32: @dacolgit, can you please also add those values to the deployment manifest in `docs/modules/publisher.md` ? It would also be great if you can add a few lines there explaining the values, so people understand why they should use `CapDrop` in production setups ?

dacolgit: done

karok2m (approved):

PR #1498, opened 2021-12-15
configurable default value for maximum nodes per dataset/subscription

cristipogacean:
add the option to configure the magic number 1000 for monitored items split into subscriptions


modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliOptions.cs
```diff
@@ -149,17 +147,17 @@ public class LegacyCliOptions : Dictionary<string, string>, IAgentConfigProvider
...
-                        (int i) => this[LegacyCliConfigKeys.ScaleTestCount] = i.ToString() },
                     { "mm|messagingmode=", "The messaging mode for messages " +
                         $"(allowed values: {string.Join(", ", Enum.GetNames(typeof(MessagingMode)))}).",
                         (MessagingMode m) => this[LegacyCliConfigKeys.MessagingMode] = m.ToString() },
                     { "me|messageencoding=", "The message encoding for messages " +
                         $"(allowed values: {string.Join(", ", Enum.GetNames(typeof(MessageEncoding)))}).",
                         (MessageEncoding m) => this[LegacyCliConfigKeys.MessageEncoding] = m.ToString() },
 
+                    // testing purposes
+                    { "sc|scaletestcount=", "The number of monitored item clones in scale tests.",
```
karok2m, line 158: Should we also have a dedicated command-line argument for maximum nodes per dataset ?

cristipogacean: Not really. I added it in the first place and then removed it. Let's keep this an an environment variable in the same manner we handle the "newly" added configuration defaults like `MaxOutgressMessages`. I think we should slowly start using these format of arguments and slowly deprecate the 2.5.x specific format. These are only available in the standalone (legacy) monde anyhow. 

btw,  `--DefaultMaxNodesPerDataSet=xxx` works perfectly well as a command line argument & cmd in container create options. We will only miss the short forms

karok2m: I am also thinking about losing mention of it in the help. But that is a minor thing, I guess.

karok2m (approved):

PR #1500, opened 2021-12-16
implementing response payload for Publish nodes direct method

monikavar:


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -396,12 +399,47 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
+                Array.Resize(ref result, result.Length > MaxResponsePayloadLength ? MaxResponsePayloadLength : result.Length);
+            }
+            MethodResponse methodResponse = new MethodResponse(result, (int)statusCode);
+            _logger.Information($"{prefix} completed with result {statusCode}");
+            return methodResponse;
+        }
+
         /// <inheritdoc/>
-        public async Task<List<string>> PublishNodesAsync(PublishedNodesEntryModel request, CancellationToken ct = default) {
+        public async Task<MethodResponse> PublishNodesAsync(PublishedNodesEntryModel request, CancellationToken ct = default) {
```
karok2m: @monikavar, please do not use `Microsoft.Azure.Devices.Client` specific `MethodResponse` here. That conversion should be done automatically by the controller wrapper. We do not want to have dependency on `Microsoft.Azure.Devices.Client` here. If you need to add new field or modify existing ones, please do that in `PublishedNodesResponseApiModel`. Erroneous cases should be propagated up via exceptions. There is a mechanism of translating them into corresponding HTTP status codes up the stack.

cristipogacean: +1
let's introduce for each direct method a XxxResponseModel class here as well as XxxResponseApiModel to properly control and encode the payload

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -445,18 +483,25 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
 
                 // fire config update so that the worker supervisor pickes up the changes ASAP
                 TriggerAgentConfigUpdate();
+                statusResponse.Add($"{nameof(PublishNodesAsync)}# Publishing successful for EndpointUrl: {request.EndpointUrl}");
             }
-            catch (MethodCallStatusException) {
-                throw;
+            catch (MethodCallStatusException e) {
+                statusMessage = $"{nameof(PublishNodesAsync)}# EndpointUrl: {request.EndpointUrl} Exception: {e.ResponsePayload}";
```
cristipogacean: The exception handling is already taken care of in the module router and works just fine. So simply rethrow the MethodCallStatusException here.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -445,18 +483,25 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
+                _logger.Error(statusMessage);
+                statusResponse.Add(statusMessage);
+                statusCode = (HttpStatusCode) e.Result;
             }
             catch (Exception e) {
-                throw new MethodCallStatusException((int)HttpStatusCode.BadRequest, e.Message);
+                statusMessage = $"{nameof(PublishNodesAsync)}# EndpointUrl: {request.EndpointUrl} Exception: {e.Message}";
```
cristipogacean: same as above, please revert to throwing a Bad request MethodCallStatusException

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -396,12 +399,47 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
             _agentConfig.TriggerConfigUpdate(this, new EventArgs());
         }
 
+        private static void AdjustResponse(ref List<string> statusResponse) {
```
cristipogacean: I'm not sure if this adjustment is the right approach here. For the response payloads that we expect to exceed the DM response capacity we have to implement a continuation token mechanism. That mechanism should be part of of the method router though, not in the orchestrator.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -396,12 +399,47 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
+        }
+        private MethodResponse BuildResponse(List<string> statusResponse, HttpStatusCode statusCode, string prefix) {
+            AdjustResponse(ref statusResponse);
+            string resultString = JsonConvert.SerializeObject(statusResponse);
+            byte[] result = Encoding.UTF8.GetBytes(resultString);
+            if (result.Length > MaxResponsePayloadLength) {
+                _logger.Error($"{prefix} Response size is too long");
+                Array.Resize(ref result, result.Length > MaxResponsePayloadLength ? MaxResponsePayloadLength : result.Length);
+            }
+            MethodResponse methodResponse = new MethodResponse(result, (int)statusCode);
```
cristipogacean: The result encoding is handled by the MethodRouter class. We just need to pass the response as an json encodable object. I believe the method router even has logic to handle the continuation tokens, we should investigate if that is enough for our needs.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -441,9 +442,10 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
         }
 
         /// <inheritdoc/>
-        public async Task<List<string>> PublishNodesAsync(PublishedNodesEntryModel request, CancellationToken ct = default) {
+        public async Task<PublishedNodesResponseApiModel> PublishNodesAsync(PublishedNodesEntryModel request, CancellationToken ct = default) {
```
karok2m: Let's keep the separation of abstractions here, so that the controller takes care of translation of response from here to `PublishedNodesResponseApiModel` and we work with `string` or `List<string>` here.

monikavar: Good point. Just changed.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -513,7 +515,7 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
                 _lockConfig.Release();
             }
 
-            return new List<string> { "Succeeded" };
+            return response;
```
cristipogacean, line 518: do we want to send an empty list here?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Controller/DmApiPublisherMethodControllerTests.cs
```diff
@@ -81,7 +81,7 @@ public class DmApiPublisherControllerTests {
 
                 publishNodesResult.Subject.StatusMessage.First()
                     .Should()
-                    .Be("Succeeded");
+                    .Contain("succeeded");
```
cristipogacean, line 84: is this test still successful? since we return an empty list?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
monikavar: We add the message in line above for successful publishing for the given endpoint URL.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Controller/DmApiPublisherMethodControllerTests.cs
monikavar: The list would not be empty so this test passes.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -495,11 +496,12 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
                     }
                 }
                 if (!found) {
-                    throw new MethodCallStatusException((int)HttpStatusCode.NotFound, "Endpoint not found.");
+                    throw new MethodCallStatusException((int)HttpStatusCode.NotFound, $"Endpoint: {request.EndpointUrl} not found.");
                 }
 
                 // fire config update so that the worker supervisor pickes up the changes ASAP
                 TriggerAgentConfigUpdate();
+                response.Add($"{nameof(PublishNodesAsync)}# Publishing succeeded for EndpointUrl: {request.EndpointUrl}");
```
karok2m: Let's not include name of the method (`nameof(PublishNodesAsync`) in response message. This is useful for logging, but for the caller it would not add much of information.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
cristipogacean: oh, I missed that, I thought it's a log entry. I'm not sure though if that response brings any useful information for the caller. I mean when the client calls the direct method, the DM's name is already known and so is the data set delta's  endpoint url. I think if we cannot add more than superfluous info in the response, we'd better leave it empty,

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: It would also be a bit confusing as the direct method name is not `PublishNodesAsync`. This is our internal naming for the handler and should not be exposed.

monikavar: removed.

karok2m (approved):

PR #1501, opened 2021-12-17
document the env variables/cli argument

cristipogacean:
* publisher: document all cli/env parametrization attributes
* publisher: cleaned-up unused legacy cli option
* twin: document all cli/env parametrization attributes
* discovery: document all cli/env parametrization attributes


dacolgit (approved):
:shipit:

karok2m (approved):

PR #1505, opened 2021-12-20
Publisher 2.5.6, fix for ICM ticket, refactor config and latest Nuget packages. 

mregen:
- bump up to use latest OPCF Nuget packages.
- fix issue that a new certificate on windows can not be used due to missing key set.
- refactor UA app config to use the OPCF config API instead of legacy init code, which caused issues with certs on .NET Core 3.1 and above.
- improve logging, to output at least error and security at info level.


modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/PublisherTelemetryConfiguration.cs
```diff
@@ -97,7 +98,8 @@ public string PatternMatch(string stringToParse)
             Match match = _patternRegex.Match(stringToParse);
             if (match.Groups[0].Success)
             {
-                foreach (var group in match.Groups.Values.Skip(1))
+                IEnumerable<Group> groups = match.Groups;
```
marcschier, line 101: For debugging?

marcschier (approved):

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/PublisherTelemetryConfiguration.cs
mregen: trick to get access to .skip otherwise build error

cristipogacean (approved):

PR #1507, opened 2021-12-20
Added stress test with 100 endpoints.

karok2m:
Added stress test with 100 endpoints.

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Engine/LegacyPublisherConfigServicesTests.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Engine/LegacyPublisherConfigServicesTests.cs
@@ -29,7 +29,25 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Tests.Engine {
     /// <summary>
     /// Tests the Direct methods configuration for the LegacyJobOrchestrator class
     /// </summary>
-    public class LegacyPublisherConfigServicesTests {
+    public class LegacyPublisherConfigServicesTests : IDisposable {
+
+        private readonly string _tempPublishedNodesFile;
+
+        public LegacyPublisherConfigServicesTests() {
+            _tempPublishedNodesFile = Path.GetTempFileName();
+        }
+
+        public void Dispose() {
+            try {
+                // Remove temporary published nodes file if one was created.
+                if (File.Exists(_tempPublishedNodesFile)) {
+                    File.Delete(_tempPublishedNodesFile);
+                }
+            }
+            catch (Exception) {
+                // Nothign to do.
+            }
+        }
 
         [Theory]
         [InlineData("Engine/publishednodes.json")]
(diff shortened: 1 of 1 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Engine/LegacyPublisherConfigServicesTests.cs
```diff
@@ -267,5 +270,139 @@ public class LegacyPublisherConfigServicesTests {
...
+                    .Should()
+                    .Be("Succeeded");
+            }
+
+            async Task CheckEndpointsAndNodes(
+                int expectedNumberOfEndpoints,
+                int expectedNumberOfNodes
+            ) {
+                var tasks = new List<Task<JobProcessingInstructionModel>>();
+                for (var i = 0; i < expectedNumberOfEndpoints; i++) {
```
cristipogacean: you can go expectedNumberOfEndpoints+1 to make sure no extra job was created.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Engine/LegacyPublisherConfigServicesTests.cs
```diff
@@ -267,5 +270,139 @@ public class LegacyPublisherConfigServicesTests {
...
+        [Fact]
+        public async Task PublishNodesStressTest() {
+            var legacyCliModelProviderMock = new Mock<ILegacyCliModelProvider>();
+            var agentConfigProviderMock = new Mock<IAgentConfigProvider>();
+            var identityMock = new Mock<IIdentity>();
+            var newtonSoftJsonSerializer = new NewtonSoftJsonSerializer();
+            var jobSerializer = new PublisherJobSerializer(newtonSoftJsonSerializer);
+            var publishedNodesJobConverter = new PublishedNodesJobConverter(TraceLogger.Create(), newtonSoftJsonSerializer);
+
+            string tempPublishedNodesFile = Path.GetTempFileName();
```
cristipogacean: we should probably delete this file as soon as the orchestrator is created and makes use of it. if we delete it only in the end of the testcase we might encounter exceptions and end-up with a bunch of undeleted empty files. 
Alternatively put the entire code in a try...finally block for cleanup.

cristipogacean (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Engine/LegacyPublisherConfigServicesTests.cs
karok2m: Good point. Done.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Engine/LegacyPublisherConfigServicesTests.cs
karok2m: Added removal of temporary file on test cleanup.

PR #1508, opened 2021-12-20
Direct method e2e publish and unpublish test

dacolgit:
Add call to publish and unpublish direct methods for current and legacy (2.5.5) publisher version in e2e test.
Fix creation of DataSetWriterId when is not defined in pn.json file


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesJobConverter.cs
```diff
@@ -259,6 +259,10 @@ public class PublishedNodesJobConverter {
         /// </summary>
         private static string GetUniqueWriterId(IEnumerable<PublishedDataSetSourceModel> set, PublishedDataSetSourceModel model) {
             var result = model.Connection.Id;
+            if (string.IsNullOrEmpty(model.Connection.Id)) {
```
cristipogacean: I see the bug here. The solution though is is not the right one.
it should be something like:
```
            var result = model.Connection.Id;
            var subset = set.Where(x => x.Connection.Id == model.Connection.Id).ToList();
            if (subset.Count > 1) {
                result += !string.IsNullOrEmpty(result) ? "_" : string.Empty;
                result += $"{model.SubscriptionSettings.PublishingInterval.GetValueOrDefault().TotalMilliseconds}";
                if (subset.Where(x => x.SubscriptionSettings.PublishingInterval == model.SubscriptionSettings.PublishingInterval).Count() > 1) {
                    result += $"_{model.PublishedVariables.PublishedData.First().PublishedVariableNodeId}";
                }
            }
            else {
                result ??= model.SubscriptionSettings.PublishingInterval.GetValueOrDefault().TotalMilliseconds.ToString();
            }
            return result;
```

e2e-tests/IIoTPlatform-E2E-Tests/Extensions/IConfigurationBuilderEx.cs
```diff
@@ -18,26 +18,26 @@ namespace IIoTPlatform_E2E_Tests.Extensions {
...
-        /// </summary>
-        /// <param name="configurationBuilder"></param>
-        /// <param name="environmentVariableTarget"></param>
-        /// <returns></returns>
-        public static IConfigurationBuilder AddEnvironmentVariables(
-            this IConfigurationBuilder configurationBuilder,
-            EnvironmentVariableTarget environmentVariableTarget
-        ) {
-            var envVarDict = Environment.GetEnvironmentVariables(environmentVariableTarget);
+        ///// <summary>
```
cristipogacean: why it this code commented out?

e2e-tests/IIoTPlatform-E2E-Tests/IIoTPlatform-E2E-Tests.csproj
```diff
@@ -14,7 +14,10 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Autofac" Version="6.3.0" />
     <PackageReference Include="Microsoft.Azure.Devices" Version="1.28.1" />
+    <PackageReference Include="Microsoft.Azure.IIoT.Core" Version="2.7.170" />
```
cristipogacean: is the 2.7.170 the right version here?

e2e-tests/IIoTPlatform-E2E-Tests/TestModels/PublishNodesRequestApiModel.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace IIoTPlatform_E2E_Tests.TestModels {
+    using System.Collections.Generic;
+    using System.ComponentModel.DataAnnotations;
+    using System.Runtime.Serialization;
+
+    /// <summary>
+    /// A monitored and published item
+    /// </summary>
+    public class PublishNodesRequestApiModel {
```
cristipogacean: do we really need to duplicate the API classes? can't we just take a dependency on the publisher api project?

e2e-tests/IIoTPlatform-E2E-Tests/TestModels/PublishedNodesEntryModel.cs
```diff
@@ -10,6 +10,21 @@ namespace IIoTPlatform_E2E_Tests.TestModels {
     /// Model with nodes to publish per Endpoint
     /// </summary>
     public class PublishedNodesEntryModel {
```
cristipogacean, line 12: I think this class is not part of the API, therefore should not be used in the E2E testing.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesJobConverter.cs
dacolgit: fixed

e2e-tests/IIoTPlatform-E2E-Tests/IIoTPlatform-E2E-Tests.csproj
dacolgit: It's the latest available

e2e-tests/IIoTPlatform-E2E-Tests/Extensions/IConfigurationBuilderEx.cs
dacolgit: code removed. Because now 
`public static IConfigurationBuilder AddEnvironmentVariables(
            this IConfigurationBuilder configurationBuilder,
            EnvironmentVariableTarget environmentVariableTarget)` is defined in  `Microsoft.Azure.IIoT.Core`

e2e-tests/IIoTPlatform-E2E-Tests/TestModels/PublishNodesRequestApiModel.cs
dacolgit: The nuget package is not updated and the publisher api project is in a different solution

e2e-tests/IIoTPlatform-E2E-Tests/Standalone/A_PublishSingleNodeStandaloneDirectMethodTestTheory.cs
```diff
@@ -79,16 +91,15 @@ IIoTMultipleNodesTestContext context
             ).GetAwaiter().GetResult());
             Assert.Null(exception);
 
-            //Todo
-            //Here instantiate _directMethod class and call the Publish direct method
-            //This is just a sample call
-            //_directMethods = new DirectMethodsAPIClient(_iotHubConnectionString, _iotHubPublisherDeviceName, _iotHubPublisherModuleName);
+            var request = model.ToApiModel();
```
cristipogacean, line 94: I think we should not apply this transformation, the payload needs to contain the api model, not the internal model.

cristipogacean (approved):

PR #1509, opened 2021-12-22
Added persistency to published nodes file.

karok2m:


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/IPublishedNodesProvider.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Storage {
+
+    /// <summary>
+    /// Interface for utilities provider for published nodes file.
+    /// </summary>
+    public interface IPublishedNodesProvider {
```
karok2m, line 14: Add method definitions.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -68,14 +63,11 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
             RefreshJobFromFile();
-
-            var file = Path.GetFileName(_legacyCliModel.PublishedNodesFile);
-            _fileSystemWatcher = new FileSystemWatcher(directory, file);
-            _fileSystemWatcher.Changed += _fileSystemWatcher_Changed;
-            _fileSystemWatcher.Created += _fileSystemWatcher_Created;
-            _fileSystemWatcher.Renamed += _fileSystemWatcher_Renamed;
-            _fileSystemWatcher.Deleted += _fileSystemWatcher_Deleted;
-            _fileSystemWatcher.EnableRaisingEvents = true;
+            _publishedNodesProvider.FileSystemWatcher.Changed += _fileSystemWatcher_Changed;
```
karok2m: Logs of file system notifications should be handled by `PublishedNodesProvider` class itself.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Controller/DmApiPublisherMethodControllerTests.cs
```diff
@@ -24,12 +24,32 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Tests.Engine {
...
+    using System;
 
     /// <summary>
     /// Tests the Direct Methods API for the pubisher
     /// </summary>
     public class DmApiPublisherControllerTests {
 
+        private readonly string _tempPublishedNodesFile;
+
+        public DmApiPublisherControllerTests() {
```
karok2m: Create a base class that will take care of providing temp file to inherited classes.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesProvider.cs
```diff
@@ -0,0 +1,159 @@
...
+                        FileAccess.Write,
+                        // We will require that there is no other process using the file.
+                        FileShare.None
+                     )) {
+                        fileStream.Write(Encoding.UTF8.GetBytes(content));
+                    }
+                }
+                catch (IOException e) {
+
+                    // We will fall back to writing with ReadWrite access.
```
karok2m, line 136: Let's log that there was a problem and we will try now with more relaxed share permissions.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -440,32 +556,66 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
-                if (!nodeFound) {
-                    existingGroup.Add(request);
+                if (!dataSetFound) {
+                    existingGroups.Add(request);
                     _publishedNodesEntrys.Add(request);
                 }
-                
+
+                // Update published nodes file.
+                var updatedContent = JsonSerializer.Serialize(_publishedNodesEntrys);
```
karok2m: Let's have a separate function that persists `_publishedNodesEntrys` to published nodes file.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -249,113 +243,235 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
+            }
+        }
+
+        private bool IsSameDataSet(PublishedNodesEntryModel entry1, PublishedNodesEntryModel entry2) {
+            return entry1.HasSameGroup(entry2)
+                && entry1.DataSetWriterId == entry2.DataSetWriterId
+                && entry1.DataSetPublishingInterval == entry2.DataSetPublishingInterval;
+        }
+
+        private IEnumerable<PublishedNodesEntryModel> AddNodes(
```
karok2m: Not used any more. Remove.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -249,113 +243,235 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
+                    entryToAddNodesSet.UnionWith(entryToAdd.OpcNodes);
+
+                    existingEntries.Add(Tuple.Create(entryToAdd, entryToAddNodesSet));
+                }
+            }
+
+            return existingEntries.ConvertAll(complexItem => complexItem.Item1);
+        }
+
+        private IEnumerable<PublishedNodesEntryModel> RemoveNodes(
```
karok2m: Not used any more. Remove.

karok2m: Removed.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: Removed.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: Added.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Controller/DmApiPublisherMethodControllerTests.cs
karok2m: Added.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/OpcNodeModel.cs
```diff
@@ -59,20 +63,22 @@ public class OpcNodeModel {
         [DataMember(EmitDefaultValue = false, IsRequired = false)]
         public string DisplayName { get; set; }
 
-        /// <summary> Heartbeat </summary>
+        /// <summary> Heartbeat interval in milliseconds </summary>
```
cristipogacean: >milliseconds [](http://example.com/codeflow?start=44&length=12)

in seconds

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/OpcNodeModel.cs
```diff
@@ -59,20 +63,22 @@ public class OpcNodeModel {
...
-        }
+        public int? HeartbeatInterval { get; set; }
 
         /// <summary>
-        /// Heartbeat interval as TimeSpan.
+        /// HeartbeatInterval as TimeSpan.
         /// </summary>
-        [DataMember(EmitDefaultValue = false, IsRequired = false)]
-        public TimeSpan? HeartbeatIntervalTimespan { get; set; }
+        [IgnoreDataMember]
```
cristipogacean: >IgnoreDataMember [](http://example.com/codeflow?start=9&length=16)

this was by intention the stored value so we can host milliseconds as well. the HeartBeat was supposed to be ignore-able, somehow it slipped.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -24,6 +23,8 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine {
     using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
+    using Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Storage;
+    using Microsoft.Azure.IIoT.Serializers;
```
cristipogacean: move these up to keep the alphabetical order guideline

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -37,11 +38,13 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
         /// <param name="jobSerializer">The serializer to (de)serialize job information.</param>
         /// <param name="logger">Logger to write log messages.</param>
-        /// <param name="identity">Module's identity provider.</param>
-
+        /// <param name="publishedNodesProvider">Published nodes provider.</param>
+        /// <param name="jsonSerializer">Json serializer.</param>
         public LegacyJobOrchestrator(PublishedNodesJobConverter publishedNodesJobConverter,
             ILegacyCliModelProvider legacyCliModelProvider, IAgentConfigProvider agentConfigProvider,
-            IJobSerializer jobSerializer, ILogger logger, IIdentity identity) {
+            IJobSerializer jobSerializer, ILogger logger, PublishedNodesProvider publishedNodesProvider,
```
cristipogacean: >PublishedNodesProvider [](http://example.com/codeflow?start=58&length=22)

shouldn't we inject the interface here?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -249,113 +246,134 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
+        /// </summary>
+        /// <param name="content"></param>
+        /// <returns></returns>
+        private IEnumerable<PublishedNodesEntryModel> DeserializePublishedNodes(string content) {
+            if (!File.Exists(_legacyCliModel.PublishedNodesSchemaFile)) {
+                _logger.Information("Validation schema file {PublishedNodesSchemaFile} does not exist or is disabled, ignoring validation of {publishedNodesFile} file...",
+                _legacyCliModel.PublishedNodesSchemaFile, _legacyCliModel.PublishedNodesFile);
+                return _publishedNodesJobConverter.Read(content, null);
+            }
+            else {
```
cristipogacean: >else { [](http://example.com/codeflow?start=12&length=6)

this is superfluous

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -249,113 +246,134 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
+                using (var fileSchemaReader = new StreamReader(_legacyCliModel.PublishedNodesSchemaFile)) {
```
cristipogacean: >StreamReader [](http://example.com/codeflow?start=50&length=12)

I guess we should put this in a try catch block and in case of exception we should fallback on converting the content without a schema.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -249,113 +246,134 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
+            return entry1.HasSameGroup(entry2)
+                && entry1.DataSetWriterId == entry2.DataSetWriterId
+                && entry1.DataSetPublishingInterval == entry2.DataSetPublishingInterval;
+        }
+
+        private void RefreshJobs(IEnumerable<PublishedNodesEntryModel> entries) {
+            var availableJobs = new Dictionary<string, JobProcessingInstructionModel>();
+            var assignedJobs = new Dictionary<string, JobProcessingInstructionModel>();
+
+            _lockJobs.Wait();
```
cristipogacean, line 293: >_lockJobs.Wait(); [](http://example.com/codeflow?start=12&length=17)

let's lock the jobs array after we do the conversion since this is quite time consuming and is independent of the jobs assignment

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -249,113 +246,134 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
+                    }
+                }
+
+                // Update local state.
+                _availableJobs = availableJobs;
+                _assignedJobs = assignedJobs;
+
+                AdjustMaxWorkersAgentConfig();
+
+                _publishedNodesEntries.Clear();
```
cristipogacean: this needs to be guarded by _lockConfig, ideally outside of the _lockJobs. Alternatively taken out of this method. Assuming the method is called inside of a the _lockConfig block is not nice...

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -440,32 +458,72 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
+
+            // This will throw SerializerException if values of request fields are not conformant.
+            DeserializePublishedNodes(requestJson);
+        }
+
+        /// <summary>
+        /// Persist _publishedNodesEntries to published nodes file.
+        /// </summary>
+        private void PersistPublishedNodes() {
+            var updatedContent = _jsonSerializer.SerializeToString(_publishedNodesEntries, SerializeOption.Indented);
```
cristipogacean, line 486: >_publishedNodesEntries [](http://example.com/codeflow?start=67&length=22)

I think we should put this in a _lockConfig block #Resolved

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -522,87 +581,54 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
-                            if (!found) {
-                                if (_availableJobs.ContainsKey(newJob.Job.Id)) {
-                                    _availableJobs[newJob.Job.Id] = newJob;
-                                    found = true;
-                                }
-                            }
-                        }
-                    }
+                _publishedNodesEntries.Clear();
+                _publishedNodesEntries.AddRange(updatedPublishedNodesEntries);
```
cristipogacean: >AddRange [](http://example.com/codeflow?start=39&length=8)

probably we should not do this if updatedPublishedNodesEntries is empty.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -522,87 +581,54 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
-                        }
-                        if (!found) {
-                            found = _availableJobs.Remove(entryJobId, out _);
-                        }
-                    }
-                }
-                finally {
-                    _lockJobs.Release();
-                }
+                PersistPublishedNodes();
```
cristipogacean, line 680: >PersistPublishedNodes [](http://example.com/codeflow?start=16&length=21)

probably we should call persist after we do refresh jobs since this will change the _publishedNodesEntries once again

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -522,87 +581,54 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
 
-                if (!found) {
-                    throw new MethodCallStatusException((int)HttpStatusCode.NotFound, "Endpoint not found.");
-                }
+                // Update local state.
+                RefreshJobs(updatedPublishedNodesEntries);
```
cristipogacean: >RefreshJobs [](http://example.com/codeflow?start=16&length=11)

Overall, I'm not convinced this approach is better, faster or more readable  than the existing implementation. Furthermore the publish/unpublish are now no longer symmetric.
I like the optimizations with the opc nodes hashset and simplification of dataset check.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesProvider.cs
```diff
@@ -0,0 +1,159 @@
...
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Models;
+using Serilog;
+using System;
+using System.IO;
+using System.Text;
+using System.Threading;
```
cristipogacean: using statements should be inside of the namespace block. like everywhere else

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -440,32 +458,72 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
...
+            var requestJson = _jsonSerializer.SerializeToString(requestList);
+
+            // This will throw SerializerException if values of request fields are not conformant.
+            DeserializePublishedNodes(requestJson);
+        }
+
+        /// <summary>
+        /// Persist _publishedNodesEntries to published nodes file.
+        /// </summary>
+        private void PersistPublishedNodes() {
```
cristipogacean, line 485: >PersistPublishedNodes [](http://example.com/codeflow?start=21&length=21)

I think we could refresh the _lastKnownFileHash in case the file was successfully written.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/IPublishedNodesProvider.cs
```diff
@@ -0,0 +1,18 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
```
cristipogacean: using inside the namespace

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesProvider.cs
```diff
@@ -0,0 +1,159 @@
...
+                        // Report and raise original exception if fallback also failed.
+                        _logger.Error(e, "Failed to update published nodes file at \"{path}\"",
+                            _legacyCliModel.PublishedNodesFile);
+
+                        throw e;
+                    }
+                }
+
+                // Retore state.
+                if (disableRaisingEvents) {
```
cristipogacean, line 160: >disableRaisingEvents [](http://example.com/codeflow?start=20&length=20)

perhaps we should restore the raise events setting in the finally block

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Storage/PublishedNodesJobConverterTests.cs
```diff
@@ -726,7 +726,7 @@ private class StandaloneIdentity : IIdentity {
         ""OpcNodes"": [
             {
                 ""Id"": ""i=2258"",
-                ""HeartbeatIntervalTimespan"": ""00:00:01.500""
+                ""HeartbeatInterval"": 1500
```
cristipogacean: >HeartbeatInterval [](http://example.com/codeflow?start=18&length=17)

let's keep the timespan test here. The test with interval is seconds is above

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Storage/PublishedNodesJobConverterTests.cs
```diff
@@ -713,7 +713,7 @@ private class StandaloneIdentity : IIdentity {
             Assert.Equal(2, jobs.Single()
                 .WriterGroup.DataSetWriters.Single()
                 .DataSet.DataSetSource.PublishedVariables.PublishedData.Single()
-                .HeartbeatInterval.Value.TotalSeconds);
+                .HeartbeatInterval.Value.TotalMilliseconds);
```
cristipogacean: >TotalMilliseconds [](http://example.com/codeflow?start=41&length=17)

I believe this was correct. we're testing for 2 seconds here

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: Yes. I need to fix the interface, then I will change this to get it.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Storage/PublishedNodesJobConverterTests.cs
karok2m: I was trying to fix the inconsistency that publishing and sampling intervals were using milliseconds while heartbeat uses seconds. Is this intentional ? The other inconsistency was that publishing and sampling intervals are using int to store the value and `TimeSpan` is just a wrapper for it. Heartbeat does opposite of that, which I want to fix. Any reasons not to do that ? I can revert it to store seconds, but otherwise I would like int to store the actual value.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesProvider.cs
karok2m: Good point.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: Done.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesProvider.cs
karok2m: Fixed.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/IPublishedNodesProvider.cs
karok2m: Fixed.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesProvider.cs
karok2m: Moved it to `finally` block.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: You mean the formatting of else or something else ? The formatting here is consistent with the rest of the file.

karok2m: Ah, you meant `else` since there is a return statement in the block before.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: Good catch, thanks.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: All calls to `PersistPublishedNodes()` are happening from direct method handlers which acquire `_lockConfig` lock. And it is a private method, so no calls from outside of the class are permitted.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: Passing an empty `IEnumerable` collection to `AddRange()` method should not add anything.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: I took update of `_publishedNodesEntries` out of `RefreshJobs()` method. Now `RefreshJobs()` only updates `_availableJobs` and `_assignedJobs`.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: Changed.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: Added exception handling for `IOException`s and fallback to deserialization without schema validation.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: Added.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: Took updates of `_publishedNodesEntries` out of `RefreshJobs()` method.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: First operation in `try` block reads values from `_assignedJobs`, so that should be under the lock. And releasing the lock in between those two operations may result in a change to `_assignedJobs` that will not be reflected in the resulting refreshed jobs, so I would like to keep the locking mechanism as it is here unless there are some other concerns.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Storage/PublishedNodesJobConverterTests.cs
cristipogacean: The heartbeat was inherited from 2.5.x as being expressed in seconds. To a certain extent it makes sense to have it different than the publishing/sampling intervals because here we're typically envision the users to have much bigger intervals than the sampling. We cannot simply change the semantic of HeartbeatInterval in this version because it will break lots of already existing configurations for both 2.5.x and 2.6+. The idea was to have the heartbeat persisted by default as timespan just to smoothly transition from the seconds int to the more comprehensive timespan - also to cover the situation when users wanted to use second subdivisions as well.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
cristipogacean: yeah, I meant we can move 
```var jobs = _publishedNodesJobConverter.ToWriterGroupJobs(entries, _legacyCliModel);```
 in the beginning - outside the lock, since it has no order dependency on 
```
 var jobIdDict = _assignedJobs
                    .Select(kvp => Tuple.Create(kvp.Value.Job.Id, kvp.Key))
                    .ToDictionary(tuple => tuple.Item1);
```

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
cristipogacean: well, I still believe the code inside a method (private or not) must not depend on external lock blocks. The code might be changed, refactored and such a design could easily lead to deadlocks or sync issues that are super difficult to debug and fix

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: OK. I am reverting back to original job refresh mechanism that was present here.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: Ah, sure. I will move job conversion out of the lock block.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
cristipogacean: that's right, but 2 lines below we call refresh jobs with an empty configuration that tries to do jobs conversions lots of memory allocations, locks and other funky stuff for nothing. We can simply clear the jobs here and not attempt a job refresh

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: Done.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: Done.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/OpcNodeModel.cs
karok2m: BTW, `OpcPublishingIntervalTimespan` and `OpcSamplingIntervalTimespan` having `[IgnoreDataMember]` set means that those fields will not be included in serialization and deserialization. So even if somebody sets `OpcPublishingIntervalTimespan` in published nodes file we will not be reading it. Is that what we want here ?

Similarly, if we set `[IgnoreDataMember]` for `HeartBeat`, then that value will also not be read from published nodes file. So we will have issues with reading old published nodes files. Which is also not desirable, right ?

karok2m: I will revert changes to heartbeat for now.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Storage/PublishedNodesJobConverterTests.cs
karok2m: Reverted changes to ` OpcNodeModel.HeartbeatInterval`.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Storage/PublishedNodesJobConverterTests.cs
karok2m: Reverted changes to ` OpcNodeModel.HeartbeatInterval`.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/OpcNodeModel.cs
karok2m: Fixed.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: I've reverted the parts bellow this to the original code that only processes `existingGroups`.

Regarding `updatedPublishedNodesEntries`: it is intended to contain full updated view of `_publishedNodesEntries`, so whether `updatedPublishedNodesEntries` is empty or not `_publishedNodesEntries` should be updated with it. The idea is that we remove nodes only if the request is not going to generate any errors, meaning that the endpoint and **all** nodes are present in our configuration. So if there are nodes that are not present, then we don't change anything. What do you think about this approach to `UnpublishNodesAsync` ? After a bit of testing I can see that this is not exactly what is happening since `_publishedNodesEntries` and `updatedPublishedNodesEntries` share same objects. So when I update `OpcNodes` of an entry in `updatedPublishedNodesEntries` it is also being updated in `_publishedNodesEntries`. I can fix this with either (1) creating full copies of entries for `updatedPublishedNodesEntries` or (2) do one pass over `_publishedNodesEntries` to see if it contains all the nodes (and throw an exception if it doesn't) and then if it does do the second (2) pass to actually remove them. I think that the second (2) approach should be cheaper since it does not create any copies and going over `_publishedNodesEntries` while having a hashset of what is there to remove is relatively fast.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: I think that the file change logs should stay in `LegacyJobOrchestrator` class since the client of `PublishedNodesProvider` class should decide what to do with the notifications and whether to act on them or not.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesProvider.cs
karok2m: Added log message.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/IPublishedNodesProvider.cs
karok2m: Added.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: Using interface.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: Implemented solution (2) with two passes.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/OpcNodeModel.cs
karok2m: Reverted.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: I will disagree here as the lock are needed for logical blocks where we change or access `_publishedNodesEntries`. This method is a part of that block just separated in helper method. I think that we should be able to have internal methods that assume being called under a lock. Of course that expectation should be documented  and enforced during reviews.

The other problem with the proposal is that we cannot achieve that with current locks as they are not reentrant, meaning that even from within the same thread the lock cannot be acquired twice. And that would be necessary as we have already acquired the lock from the calling block. Releasing the lock in between is not an option here as that introduces possibilities of inconsistent changes to `_publishedNodesEntries` and its serialized version in JSON file. We can use reentrant locks such as [Monitor](https://docs.microsoft.com/en-us/dotnet/api/system.threading.monitor?redirectedfrom=MSDN&view=net-6.0) here to work around that, but I would still argue that having internal methods that assume being called from within a lock block is a valid option.

cristipogacean: Got it. It's fine with me.

cristipogacean (approved):

PR #1514, opened 2021-12-23
adding exception handling in publishnodes dm

monikavar:


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -504,6 +505,9 @@ public class LegacyJobOrchestrator : IJobOrchestrator, IPublisherConfigServices,
             catch (MethodCallStatusException) {
                 throw;
             }
+            catch (UnknownJobTypeException ex) {
```
karok2m, line 508: Where would the exception be coming from ? I am wondering to better understand if it would be caused by a bad request or if it would be an internal error.

monikavar: in the **`ToJobProcessingInstructionModel`** method while serializing job, it might throw **`UnknownJobTypeException`**

karok2m: This would be out internal conversion error that does not come from an invalid request. That case would already be covered in `ExceptionsFilterAttribute` class, as any unknown exception will be reported as `HttpStatusCode.InternalServerError`. So we do not need to have a special handling for it here. Given that, I think that we can close this PR and move the task for adding exception handling for `UnpublishNodesAsync()` to done.

monikavar: Alright!

Issue #1203, opened 2021-06-04
Deployment failed when running script - Simulation devices


NilssonCJ:
**Describe the bug**
Deployment failed when running script - see **Additional context** for error message.
Happens when I run the regular script.sh and it tries to deploy the simulation devices.

**To Reproduce**
1. Go to Industrial-IoT folder
2. Run script.sh
3. See error

**Expected behavior**
That it would finish the deployment but I get errors when I deploy instead.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: MacOS

**Additional context**
8:35:10 AM - The deployment 'azuredeploy' failed with error(s). Showing 3 out of 7 error(s).
Status Message: At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/DeployOperations for usage details. (Code: DeploymentFailed)
 - {
  "error": {
    "code": "BadRequest",
    "message": "The selected VM size 'Standard_DC1s_v2' cannot boot Hypervisor Generation '1'. If this was a Create operation please check that the Hypervisor Generation of the Image matches the Hypervisor Generation of the selected VM Size. If this was an Update operation please select a Hypervisor Generation '1' VM Size."
  }
} (Code:BadRequest)

Status Message: The selected VM size 'Standard_DC1s_v2' cannot boot Hypervisor Generation '1'. If this was a Create operation please check that the Hypervisor Generation of the Image matches the Hypervisor Generation of the selected VM Size. If this was an Update operation please select a Hypervisor Generation '1' VM Size. (Code:BadRequest)

Status Message: Template output evaluation skipped: at least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/DeployOperations for usage details. (Code:DeploymentOperationFailed)

marcschier:
The deploy script does not use the hyper v gen as criteria for selecting the VM sku.  This filter needs to be added to the script.  Also, we might want a script param to allow overriding the automated selection of sku.

marcschier:
[...]

PR #1516, opened 2021-12-30
deploy.ps1: allow VM SKU sizes to be passed as parameter

saramonteiro:
# Description

This PR is a workaround for the issue https://github.com/Azure/Industrial-IoT/issues/1203. 
It only adds the feature of passing the VM SKU for the gateway host and OPC PLC server simulator as parameters.
A complete(and better) fix would include filter out the non Generation 1 SKU from the VM SKU list.

# Test

I got a successful deploy by using the following command:

`.\deploy -gatewayVMSKU Standard_D1_v2 -OPCPLCVMSKU Standard_D2_v2`

# Note:

If you choose to pass the SKU as parameters, make sure they're available on your chosen region by running the following snippet on your Power Shell:

``` bash
            Get-AzVMSize -Location "East US" `
            | Where-Object {
                ($_.NumberOfCores -ge 1) -and `
                ($_.MemoryInMB -ge 2048) -and `
                ($_.OSDiskSizeInMB -ge 1047552) -and `
                ($_.ResourceDiskSizeInMB -ge 4096)
            } `
```


deploy/scripts/deploy.ps1
```diff
@@ -70,6 +70,16 @@
 
  .PARAMETER numberOfWindowsGateways
     Number of Windows gateways to deploy into the simulation.
+
+ .PARAMETER gatewayVMSKU
```
karok2m: @saramonteiro, let's make name of the variable came-case to align with the rest. Something like `gatewayVmSku`.

deploy/scripts/deploy.ps1
```diff
@@ -70,6 +70,16 @@
...
+    Virtual machine SKU size that hosts simulated edge gateway.
+    Suggestion: use VM with at least 2 cores and 8 GB of memory.
+    Must Support Generation 1.
+
+ .PARAMETER OPCPLCVMSKU
```
karok2m: @saramonteiro, let's make name of the variable came-case to align with the rest. Something like `opcPlcVmSku`.

saramonteiro: Fixed, when it's ok, let me know that I squash my commits.

karok2m (approved):

cristipogacean (approved):

karok2m:
Here is the successful run of [Azure.Industrial-IoT pipeline](https://msazure.visualstudio.com/One/_build/results?buildId=51048751&view=results) for this PR.

PR #1519, opened 2022-01-13
Additional logging for errors of monitored items subscriptions.

karok2m:
Added logging in the following cases:
* When setting monitoring for an item returns not good status.
* When receiving a data change notification for an item that has `null` value  or value status code is not good.

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
@@ -808,15 +808,17 @@ namespace Microsoft.Azure.IIoT.OpcUa.Protocol.Services {
                         }
                     }
 
-                    var erroneousNotifications = message.Notifications?
-                        .Where(n => n.Value.Value == null
-                            || StatusCode.IsNotGood(n.Value.StatusCode))
-                        .ToList();
+                    if (_logger.IsEnabled(Serilog.Events.LogEventLevel.Debug)) {
+                        var erroneousNotifications = message.Notifications
+                            .Where(n => n.Value.Value == null
+                                || StatusCode.IsNotGood(n.Value.StatusCode))
+                            .ToList();
 
-                    if (erroneousNotifications.Any()) {
-                        _logger.Debug("Found {count} notifications with null value or not good status " +
-                            "code for '{subscription}' subscription.", erroneousNotifications.Count,
-                            subscription.DisplayName);
+                        if (erroneousNotifications.Count > 0) {
+                            _logger.Debug("Found {count} notifications with null value or not good status " +
+                                "code for '{subscription}' subscription.", erroneousNotifications.Count,
+                                subscription.DisplayName);
+                        }
                     }
 
                     if (message.Notifications?.Any() == true) {
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
@@ -543,7 +543,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Protocol.Services {
                         if (change.Key == null) {
                             continue;
                         }
-                        _logger.Information("Set monitoring to \"{value}\" for {count} items in subscription \"{subscription}\".",
+                        _logger.Information("Set monitoring to '{value}' for {count} items in subscription '{subscription}'.",
                             change.Key.Value, change.Count(), rawSubscription.DisplayName);
 
                         var itemsToChange = change.Select(t => t.Item).ToList();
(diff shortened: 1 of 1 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -760,43 +773,52 @@ internal sealed class SubscriptionWrapper : ISubscription {
...
                         }
                     }
 
+                    var erroneousNotifications = message.Notifications
+                        .Where(n => n.Value.Value == null
+                            || StatusCode.IsNotGood(n.Value.StatusCode))
+                        .ToList();
+
+                    if (erroneousNotifications.Any()) {
+                        _logger.Warning("Found {count} notifications with null value or not good status " +
```
karok2m: Consider switching to `_logger.Debug()` when merging to `main` as this log might be very noisy.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -543,16 +543,27 @@ internal sealed class SubscriptionWrapper : ISubscription {
...
+                                .Count(r => (r == null) ? false : StatusCode.IsNotGood(r.StatusCode));
+
+                            // Check the number of erroneous results and log.
+                            if (erroneousResultsCount > 0) {
+                                _logger.Warning("Failed to set monitoring for {count} items in subscription \"{subscription}\".",
+                                    erroneousResultsCount, rawSubscription.DisplayName);
+
+                                for (int i = 0; i < results.Count && i < itemsToChange.Count; ++ i) {
+                                    if (StatusCode.IsNotGood(results[i].StatusCode)){
+                                        _logger.Warning("Set monitoring for item \"{item}\" in subscription \"{subscription}\" failed with \"{status}\".",
```
karok2m: Consider switching to `_logger.Debug()` when merging to `main` as we might have several thousands of items.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
cristipogacean: indeed, we should use `Debug()` here

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
cristipogacean: I think this is an abnormal situation therefore a warning is appropriate here.

cristipogacean (dismissed):

cristipogacean (dismissed):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
karok2m: OK. Then we can leave it as it is.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -760,43 +773,52 @@ internal sealed class SubscriptionWrapper : ISubscription {
...
                             item.ValidateHeartbeat(publishTime);
                         }
                     }
 
+                    var erroneousNotifications = message.Notifications?
+                        .Where(n => n.Value.Value == null
+                            || StatusCode.IsNotGood(n.Value.StatusCode))
+                        .ToList();
+
+                    if (erroneousNotifications.Any()) {
```
marcschier: Use erroneousNotifications.Count > 0, more efficient here.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -760,43 +773,52 @@ internal sealed class SubscriptionWrapper : ISubscription {
...
+                                    heartbeatValue.PublishTime = publishTime;
+                                    message.Notifications?.Add(heartbeatValue);
+                                }
+                                continue;
                             }
                             item.ValidateHeartbeat(publishTime);
                         }
                     }
 
+                    var erroneousNotifications = message.Notifications?
```
marcschier: Since this is costly, maybe hide it behind a check for Debug Level being enabled on the logger?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
karok2m: Since `message.Notifications?` may be null, we need to check for null here as well by `erroneousNotifications?.Any()`.

karok2m: OK, I take this back as we generate `message` and make sure that `Notifications` is not null during construction. So we can even take `?` out from `message.Notifications?` when initializing `erroneousNotifications`. But it is fine as it is as well.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
karok2m: Done by @cristipogacean.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
karok2m: Removed `?`.

karok2m: Done.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
karok2m: Moved calculation behind a check.

cristipogacean (dismissed):

cristipogacean (approved):

marcschier (approved):

PR #1520, opened 2022-01-13
Add direct method GetConfiguredEndpoints

monikavar:
Adding direct method GetConfiguredEndpoints and tests to verify.


modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Models/PublisherExtensions.cs
```diff
@@ -70,5 +70,19 @@ public static class PublisherExtensions {
                 StatusMessage = model
             };
         }
+
+        /// <summary>
+        /// Create an api model from service model
+        /// </summary>
+        public static ConfiguredEndpointsResponseApiModel ToConfiguredEndpointsApiModel(
```
karok2m: Can the method be called `ToApiModel` to be consistent with the rest of the transformation calls in the controller ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/ConfiguredEndpointModel.cs
```diff
@@ -0,0 +1,53 @@
+﻿// ------------------------------------------------------------
```
cristipogacean: Do we need this new model? I think PublishedNodesEntryModel can be reused here

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/ConfiguredEndpointsResponseApiModel.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// </summary>
+    [DataContract]
+    public class ConfiguredEndpointsResponseApiModel {
+
+        /// <summary>
+        /// Configured endpoint urls
+        /// </summary>
+        [DataMember(Name = "endpoints", Order = 0)]
+        [Required]
+        public List<ConfiguredEndpointModel> Endpoints { get; set; }
```
cristipogacean: I think here we need an "API" class model, preferably the PublishNodesRequestApiModel

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -762,16 +762,34 @@ IJsonSerializer jsonSerializer
         }
 
         /// <inheritdoc/>
-        public async Task<List<PublishedNodesEntryModel>> GetConfiguredEndpointsAsync(
+        public async Task<List<ConfiguredEndpointModel>> GetConfiguredEndpointsAsync(
```
cristipogacean: I think we could avoid duplicate code by just using  here with PublishedNodesEntryModel here. Of course with no OpcNodes set.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Models/PublisherExtensions.cs
monikavar: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/ConfiguredEndpointModel.cs
monikavar: yes, reused PublishedNodesEntryModel

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Controller/PublisherMethodsController.cs
```diff
@@ -64,10 +64,10 @@ public class PublisherMethodsController : IMethodController {
         /// <summary>
         /// Handler for GetConfiguredEndpoints DM
         /// </summary>
-        public async Task<List<PublishedNodesResponseApiModel>> GetConfiguredEndpointsAsync() {
+        public async Task<List<PublishNodesRequestApiModel>> GetConfiguredEndpointsAsync() {
```
karok2m: Using a request API class for response kind of defeats the purpose of naming it request API model. Either the name of the class should be changed to reflect the entity that it holds and untie it from request/response parts or a new class should be created.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Models/PublisherExtensions.cs
```diff
@@ -70,5 +70,25 @@ public static class PublisherExtensions {
                 StatusMessage = model
             };
         }
+
+        /// <summary>
+        /// Create an api model from service model
```
karok2m: Need to specify that password will be ignored.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Controller/PublisherMethodsController.cs
karok2m: @monikavar, please add unit and integration tests for `GetConfiguredEndpoints_V1` direct method.

monikavar: Should we rather rename this class then?
as @cristipogacean mentioned in a comment above that we shouldn't be creating a duplicate class for this.

monikavar: missed to push the file in earlier commit. done now.

karok2m: @monikavar, I also think that we don't need to introduce duplicates of the classes. So let's rename it to `EndpointApiModel` so that we use it in both requests and responses without introducing any confusion.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Controller/DmApiPublisherMethodControllerTests.cs
```diff
@@ -221,5 +221,68 @@ public class DmApiPublisherControllerTests : TempFileProviderBase {
...
+                legacyCliModelProviderMock.Object,
+                agentConfigProviderMock.Object,
+                jobSerializer,
+                logger,
+                publishedNodesProvider,
+                newtonSoftJsonSerializer
+            );
+            var methodsController = new PublisherMethodsController(orchestrator);
+
+            using var publishPayloads = new StreamReader(publishedNodesFile);
```
karok2m, line 515: @monikavar, let's also call the `GetConfiguredEndpointsAsync()` method before we call `PublishNodesAsync` and make sure that it returns empty list.

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/PublishNodesEndpointApiModel.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Api.Publisher.Models {
     /// A monitored and published item
     /// </summary>
     [DataContract]
-    public class PublishNodesRequestApiModel {
+    public class PublishNodesEndpointApiModel {
```
karok2m, line 15: @monikavar, I would still suggest to rename the class to `EndpointApiModel` as initial naming of `PublishNodesRequestApiModel` implied that it is used for `PublishNodes` direct method. That is no longer true as now we are using this class in the majority of the direct methods whether it is as a request or a response. So we can drop `PublishNodes` part of its name.

karok2m:
@monikavar, please merge `feature/dmapi` branch into ` movarshn/configureendpoints` and resolve the conflicts that are showing up here.

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/PublishNodesEndpointApiModel.cs
karok2m: `EndpointApiModel` already exists. Let's keep it as `PublishNodesEndpointApiModel` then.

monikavar: `EndpointApiModel` already exists so I would it as is.

karok2m (approved):

PR #1522, opened 2022-01-14
Add GetConfiguredNodesOnEndpoint dm

dacolgit:
Add GetConfiguredNodesOnEndpoint dm
add unit test


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -775,18 +776,43 @@ IJsonSerializer jsonSerializer
...
         /// <inheritdoc/>
-        public async Task<PublishedNodesEntryModel> GetConfiguredNodesOnEndpointAsync(
+        public async Task<List<OpcNodeOnEndpointModel>> GetConfiguredNodesOnEndpointAsync(
             PublishedNodesEntryModel request,
             CancellationToken ct = default) {
 
             _logger.Information("{nameof} method triggered", nameof(GetConfiguredNodesOnEndpointAsync));
+            var sw = Stopwatch.StartNew();
             await _lockConfig.WaitAsync(ct).ConfigureAwait(false);
+            List<OpcNodeOnEndpointModel> response = new List<OpcNodeOnEndpointModel>();
```
karok2m: Please switch lines `785` and `786` so that acquiring the lock is immediately followed by the `try/catch/finally` block.

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/GetConfiguredNodesOnEndpointsRequestApiModel.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// <summary>
+    /// Model for a get configured nodes on endpoint request
+    /// </summary>
+    [DataContract]
+    public class GetConfiguredNodesOnEndpointsRequestApiModel {
+
+        /// <summary> Endpoint URL for the OPC Nodes to monitor. </summary>
+        [DataMember(Name = "endpointUrl", Order = 0)]
+        [Required]
+        public string EndpointUrl { get; set; }
```
karok2m: @dacolgit, please have a look at `PublishNodesRequestApiModel`. Here we will have to have all of its fields excluding `OpcNodes` as their combination as a whole would define the endpoint that we are looking for and not only the `EndpointUrl`.

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher.csproj
```diff
@@ -7,6 +7,7 @@
     <Description>Azure Industrial IoT OPC Publisher API</Description>
   </PropertyGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\..\..\components\opc-ua\src\Microsoft.Azure.IIoT.OpcUa\src\Microsoft.Azure.IIoT.OpcUa.csproj" />
```
marcschier: There should not be a dependency of the API on the service's core business logic.  Can we change this?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -775,18 +776,43 @@ IJsonSerializer jsonSerializer
...
             try {
-                throw new MethodCallStatusException((int)HttpStatusCode.NotImplemented, "Not Implemented");
+                var nodeFound = false;
+                if (!_publishedNodesEntries.Any()) {
```
karok2m: Let's get rid of this if and just do the `foreach` loop that is present in the `else` block. If `_publishedNodesEntries` list is empty `foreach` loop will detect it and we will not need to have to throw the very same `MethodCallStatusException((int)HttpStatusCode.NotFound, "Node not found in endpoint.")` exception in two places.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -775,18 +776,43 @@ IJsonSerializer jsonSerializer
...
+                    throw new MethodCallStatusException((int)HttpStatusCode.NotFound, "Node not found in endpoint.");
+                }
+                else {
+                    foreach (var entry in _publishedNodesEntries) {
+                        if (request.EndpointUrl == entry.EndpointUrl) {
```
karok2m: We need to have more complete comparison for determining correct endpoint similar to what we have in publish and unpublish nodes, which use `IsSameDataSet(entry, request)` (which internally also checks for `entry.HasSameGroup(request)`).

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/OpcNodeOnEndpointModel.cs
```diff
@@ -0,0 +1,46 @@
...
+    using System;
+    using Newtonsoft.Json;
+    using System.Text;
+    using System.Runtime.Serialization;
+
+    /// <summary>
+    /// Class describing a list of nodes
+    /// </summary>
+    [DataContract]
+    public class OpcNodeOnEndpointModel {
```
karok2m: `DataSetFieldId` and `QueueSize` properties are missing when compared to `OpcNodeModel` class.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/OpcNodeOnEndpointModel.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <summary> Display name </summary>
+        [DataMember(EmitDefaultValue = false, IsRequired = false)]
+        public string DisplayName { get; set; }
+
+        /// <summary> Heartbeat </summary>
+        [DataMember(EmitDefaultValue = false, IsRequired = false)]
+        public int? HeartbeatInterval { get; set; }
+
+        /// <summary> Queue Size for the monitored item </summary>
```
karok2m: The comment is for `QueueSize` property and not `SkipFirst`.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/OpcNodeOnEndpointModel.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <summary> Node Identifier </summary>
+        [DataMember(EmitDefaultValue = false, IsRequired = false)]
+        public string Id { get; set; }
+
+        /// <summary> Expanded Node identifier </summary>
+        [DataMember(EmitDefaultValue = false, IsRequired = false)]
+        public string ExpandedNodeId { get; set; }
+
+        /// <summary> Sampling interval </summary>
```
karok2m: Please specify whether interval is in seconds or milliseconds for `OpcSamplingInterval`, `OpcPublishingInterval` and `HeartbeatInterval` as they are different in `OpcNodeModel`.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
dacolgit: switched

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/GetConfiguredNodesOnEndpointsRequestApiModel.cs
dacolgit: added

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
dacolgit: removed

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/OpcNodeOnEndpointModel.cs
dacolgit: added

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/OpcNodeOnEndpointModel.cs
dacolgit: fixed

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/OpcNodeOnEndpointModel.cs
dacolgit: added

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher.csproj
dacolgit: removed

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/GetConfiguredNodesOnEndpointResponseApiModel.cs
```diff
@@ -0,0 +1,31 @@
+﻿// ------------------------------------------------------------
```
cristipogacean: I think we don't really need this model. For the response I think it would be enough to return the a list of PublishedNodeApiModel instances.

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/GetConfiguredNodesOnEndpointsRequestApiModel.cs
```diff
@@ -0,0 +1,56 @@
+﻿// ------------------------------------------------------------
```
cristipogacean: I think we don't really need this model. we already have the PublishNodesRequestApiModel that covers all the content here.

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/OpcNodeOnEndpointApiModel.cs
```diff
@@ -0,0 +1,55 @@
+﻿// ------------------------------------------------------------
```
cristipogacean: I think we can reuse PublishedNodeApiModel here

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -775,18 +776,40 @@ IJsonSerializer jsonSerializer
         }
 
         /// <inheritdoc/>
-        public async Task<PublishedNodesEntryModel> GetConfiguredNodesOnEndpointAsync(
+        public async Task<List<OpcNodeOnEndpointModel>> GetConfiguredNodesOnEndpointAsync(
```
cristipogacean: I believe we ca keep the PublishedNodesEntryModel here

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/OpcNodeOnEndpointModel.cs
```diff
@@ -0,0 +1,54 @@
+﻿// ------------------------------------------------------------
```
cristipogacean: can't we just use OpcNodeModel instead of thisone?

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/GetConfiguredNodesOnEndpointResponseApiModel.cs
dacolgit: removed. using PublishedNodeApiModel

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/GetConfiguredNodesOnEndpointsRequestApiModel.cs
dacolgit: removed. Using PublishNodesRequestApiModel

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/OpcNodeOnEndpointApiModel.cs
dacolgit: yes

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
dacolgit: removed. Using PublishedNodesEntryModel

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/OpcNodeOnEndpointModel.cs
dacolgit: removed

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Controller/DmApiPublisherMethodControllerTests.cs
```diff
@@ -221,5 +222,207 @@ public class DmApiPublisherControllerTests : TempFileProviderBase {
...
+            var endpointRequest = new PublishNodesRequestApiModel();
+            endpointRequest.EndpointUrl = endpointUrl;
+            var response = await FluentActions
+                    .Invoking(async () => await methodsController
+                    .GetConfiguredNodesOnEndpointAsync(endpointRequest).ConfigureAwait(false))
+                    .Should()
+                    .NotThrowAsync()
+                    .ConfigureAwait(false);
+
+            Assert.True(response.Subject.Count == 2);
```
cristipogacean: >Assert [](http://example.com/codeflow?start=12&length=6)

Let's stay consistent and use FluentAssertions overall in this file.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Extensions/PublishedNodesEntryModelEx.cs
```diff
@@ -35,7 +35,7 @@ public static class PublishedNodesEntryModelEx {
                 return false;
             }
 
-            if (model.OpcAuthenticationMode != that.OpcAuthenticationMode) {
+            if (model.OpcAuthenticationMode != that.OpcAuthenticationMode && that.OpcAuthenticationMode != OpcAuthenticationMode.Anonymous) {
```
cristipogacean: >&& that.OpcAuthenticationMode != OpcAuthenticationMode.Anonymous [](http://example.com/codeflow?start=74&length=64)

what is the usecase covered by this change?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Extensions/PublishedNodesEntryModelEx.cs
```diff
@@ -58,6 +58,69 @@ public static class PublishedNodesEntryModelEx {
             return true;
         }
 
+        /// <summary>
+        /// filter model
+        /// </summary>
+        public static bool IsFiltered(this PublishedNodesEntryModel model, PublishedNodesEntryModel that) {
```
cristipogacean: >IsFiltered [](http://example.com/codeflow?start=27&length=10)

what is the purpose of this method? how is it different from HasSameGroup()? 

we should probably rename it.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Extensions/PublishedNodesEntryModelEx.cs
```diff
@@ -58,6 +58,69 @@ public static class PublishedNodesEntryModelEx {
...
+            if (model == that) {
+                return true;
+            }
+
+            if (model == null || that == null) {
+                return false;
+            }
+
+            if (string.Compare(model.DataSetWriterGroup, that.DataSetWriterGroup, StringComparison.InvariantCulture) != 0 &&
+                that.DataSetWriterGroup != null) {
```
cristipogacean: I'm not sure if this is correct for all use-cases 
e.g.:
with
 model.DataSetWriterGroup = "1"
 that.DataSetWriterGroup = null 

the result will be true, meaning the fields are equal, which is not exactly right.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Extensions/PublishedNodesEntryModelEx.cs
dacolgit: this cover the case where OpcAuthenticationMode is not used as filtering option.
if OpcAuthenticationMode is not explicity set, it has the default value  OpcAuthenticationMode.Anonymous

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Extensions/PublishedNodesEntryModelEx.cs
dacolgit: This is used to filter the nodes on endpoint.
model is the actual entry, that.DataSetWriterGroup is the filtering mask. If that.DataSetWriterGroup is null it means that every DataSetWriterGroup is valid.
if we have 2 DataSetWriterGroup with same endpointUrl and just use the endpointUrl to filter the nodes, we get all nodes from both DataSetWriterGroup

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Extensions/PublishedNodesEntryModelEx.cs
cristipogacean: what happens when 
model.OpcAuthenticationMode = OpcAuthenticationMode.UserName
that.OpcAuthenticationMode = OpcAuthenticationMode.Anonymous
?
they look different to me, but the code sequence will conclude they are equal.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Extensions/PublishedNodesEntryModelEx.cs
dacolgit: the differences are:
added comparison also for DataSetWriterId
added for each element also the check that is not null or default (OpcAuthenticationMode)

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Extensions/PublishedNodesEntryModelEx.cs
cristipogacean: I'm not sure if I agree with this: 

> if we have 2 DataSetWriterGroup with same endpointUrl and just use the endpointUrl to filter the nodes, we get all nodes from both DataSetWriterGroup

I think we should not mix in the reply nodes from different DataSetWriterGroups. The problem is that the user will not be able to differentiate which node belongs in which group. If the request is ambiguous, we should just return an error so the user can refine it's request.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Controller/DmApiPublisherMethodControllerTests.cs
dacolgit: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Extensions/PublishedNodesEntryModelEx.cs
dacolgit: fixed

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Extensions/PublishedNodesEntryModelEx.cs
dacolgit: fixed

cristipogacean (approved):
:shipit:

PR #1532, opened 2022-01-25
Fixed data set detection bug in GetConfiguredNodesOnEndpointAsync().

karok2m:
Changed data set detection logic in `GetConfiguredNodesOnEndpointAsync()` method to use `IsSameDataSet()` instead of `HasSameGroup()` which also compares `DataSetWriterId` and `DataSetPublishingInterval` properties of data sets.
Added test which queries for nodes of specific endpoints where endpoints differ only in `DataSetWriterId` property.


dacolgit (approved):

cristipogacean (approved):
:shipit:

PR #1535, opened 2022-01-26
Adding two additional error codes as possible results of the call of the service ActivateSession (BadInvalidState and BadRequestTimeout), which do not have to result in cleanup, disposing and recreation of the current session

dimtix:
In some cases in our setup (publisher in standalone mode / pn.json), when there are a bit longer as normal disconnections between the publisher and our OPC UA Servers (e.g. 2-3 min) we experience data loss (even though SessionTimeout is set to 20 min and the QueueSizes of monitorerd items are sufficient). The reason for the data loss is - for some error codes returned at (re)ActivateSession, session cleanup, disposing and session recreation are performed, although the session could be reactivated and existing queues republished in the next refresh cycle without any problems.

These are the following error codes:
1) BadRequestTimeout - when connection loss lasts longer than configured request timeout (example attached)
2) BadInvalidState - the cause is unclear - doesn't occur always (example attached)

Both error codes were added to the error codes after which it must be tried again to make a reconnect in the next Retry cycle. 
After adding there were no more data losses in given cases and everything worked as expected. 
In addition, the description for the cli key ct|createsessiontimeout was adjusted cause it was not quite correct and understandable imho.

[publisher-log-sample-BadRequestTimeout.txt](https://github.com/Azure/Industrial-IoT/files/7946181/publisher-log-sample-BadRequestTimeout.txt)
[publisher-log-sample-BadInvalidState.txt](https://github.com/Azure/Industrial-IoT/files/7946182/publisher-log-sample-BadInvalidState.txt)

### DIFF
--- a/modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliOptions.cs
+++ b/modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliOptions.cs
@@ -95,12 +95,14 @@ namespace Microsoft.Azure.IIoT.Modules.OpcUa.Publisher.Runtime {
                         "sample values.",
                         (int i) => this[LegacyCliConfigKeys.OpcSamplingInterval] = TimeSpan.FromMilliseconds(i).ToString() },
                     { "op|opcpublishinginterval=", "Default value in milliseconds for the publishing interval " +
-                            "setting of the subscriptions against the OPC UA server.",
+                        "setting of the subscriptions against the OPC UA server.",
                         (int i) => this[LegacyCliConfigKeys.OpcPublishingInterval] = TimeSpan.FromMilliseconds(i).ToString() },
-                    { "ct|createsessiontimeout=", "Maximum amount of time that a session should remain open by the OPC server without any activity (session timeout) - to request from the OPC server at session creation.",
+                    { "ct|createsessiontimeout=", "Maximum amount of time in seconds that a session should " +
+                        "remain open by the OPC server without any activity (session timeout) " +
+                        "- to request from the OPC server at session creation.",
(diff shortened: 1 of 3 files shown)


modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliOptions.cs
```diff
@@ -97,7 +97,7 @@ public class LegacyCliOptions : Dictionary<string, string>, IAgentConfigProvider
                     { "op|opcpublishinginterval=", "Default value in milliseconds for the publishing interval " +
                             "setting of the subscriptions against the OPC UA server.",
                         (int i) => this[LegacyCliConfigKeys.OpcPublishingInterval] = TimeSpan.FromMilliseconds(i).ToString() },
-                    { "ct|createsessiontimeout=", "The timeout in seconds used when creating a session to an endpoint.",
+                    { "ct|createsessiontimeout=", "Maximum amount of time that a session should remain open by the OPC server without any activity (session timeout) - to request from the OPC server at session creation.",
```
cristipogacean: >Maximum amount of time that a session should remain open by the OPC server without any activity (session timeout) - to request from the OPC server at session creation. [](http://example.com/codeflow?start=51&length=167)

can you please split this on 2 or 3 lines so it fits the general code look and feel? Please also specify this value is in seconds

cristipogacean (commented):
Thanks a lot for this contribution! I added a small cosmetic comment. Other than that it looks good!

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliOptions.cs
dimtix: done

cristipogacean (approved):

karok2m:
Successful run of our build and test pipeline for this PR can be found [here](https://msazure.visualstudio.com/One/_build/results?buildId=51645716&view=results).

PR #1539, opened 2022-01-28
Added implementation of AddOrUpdateEndpoints_V1 direct method call.

karok2m:
Changes:
* Added implementation of `AddOrUpdateEndpoints_V1` direct method call.
* Null reference bug fixed in `PublishedNodesProvider` class.
* Fixed data set detection bug in `GetConfiguredNodesOnEndpointAsync()` method.
* Added tests for `AddOrUpdateEndpointsAsync()` method.
* Adjusting max worker count after execution of `AddOrUpdateEndpointsAsync()` method.
* Added test to check functionality of adding endpoints with `AddOrUpdateEndpointsAsync()` method.
* Added endpoint removal tests for `AddOrUpdateEndpointsAsync()` method.
* Added validation of request to ensure that it does not contain two entries for the same endpoint.


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -765,6 +773,169 @@ IJsonSerializer jsonSerializer
...
+                                }
+
+                                // We do not need to look for another matching data set in request.
+                                requestDataSetsFound[k] = true;
+                                break;
+                            }
+                        }
+                    }
+
+                    if (groupFound) {
```
cristipogacean, line 876: >groupFound [](http://example.com/codeflow?start=24&length=10)

isn't this the other way around (!groupFound)? at least the description implies this.  #Resolved

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -765,6 +773,169 @@ IJsonSerializer jsonSerializer
...
+                        var dataSetToUpdate = request[k];
+                        if (entry.HasSameGroup(dataSetToUpdate)) {
+                            groupFound = true;
+
+                            // We may have several entries with the same DataSetGroup definition,
+                            // so we will update nodes only if the whole DataSet definition matches.
+                            if (IsSameDataSet(entry, dataSetToUpdate)) {
+                                if (dataSetToUpdate.OpcNodes is null || dataSetToUpdate.OpcNodes.Count == 0 || requestDataSetsFound[k]) {
+                                    // In this case existing OpcNodes entries should be cleaned up.
+                                    entry.OpcNodes.Clear();
```
cristipogacean: >entry.OpcNodes.Clear() [](http://example.com/codeflow?start=36&length=22)

do we know for sure that the entry.OpcNodes is never null?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: yeah, it shows if a matching data set group was found (in `request`) for a specific entry in `_publishedNodesEntries`.

cristipogacean: now I got it

cristipogacean (approved):
:shipit:

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
karok2m: Let me see what will happen if we have an entry with null `OpcNodes` in JSON file.

karok2m: I've added tests to validate:
* that empty or null `OpcNodes` from published nodes JSON file do not result in entries in `_publishedNodesEntries`.
* that `null` requests or requests with `null` or empty `OpcNodes` are reported as errors.

karok2m: Also changed the call to be ` entry.OpcNodes?.Clear();` so that it does not throw exceptions in those cases. Entries with `null` `OpcNode` will be cleaned up in the next step.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -798,17 +989,25 @@ IJsonSerializer jsonSerializer
         /// <inheritdoc/>
         public async Task<List<OpcNodeModel>> GetConfiguredNodesOnEndpointAsync(
             PublishedNodesEntryModel request,
-            CancellationToken ct = default) {
-
+            CancellationToken ct = default
+        ) {
```
cristipogacean, line 1003: is this necessary? I think we should agree on a coding style and keep look and feel consistent. When in Rome ...

karok2m: Nop, not necessary. Will revert this change.

karok2m: Reverted.

cristipogacean (approved):

monikavar (approved):

PR #1540, opened 2022-01-30
Cert creation fix and test code for 2.5.7

mregen:
- the UA stack released Version 1.4.367.95 which contains the fix for the windows container cert loading issue
- add also some dockerfiles which were used to debug and test


marcschier (approved):

NuGet.Config
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <!--<add key="opcua-preview" value="https://opcfoundation.pkgs.visualstudio.com/opcua-netstandard/_packaging/opcua-preview/nuget/v3/index.json" />-->
```
karok2m, line 4: Shouldn't this file be removed since everything is commented out in it ?

cristipogacean (approved):

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/endlos.cmd
```diff
@@ -0,0 +1,4 @@
+REM Keep restarting docker to repro a certificate creation issue on windows containers
+REM Build the container with dockerbuild.cmd
+docker run opcpublisher:latest OpcPublisher >>endlesstest.log
```
karok2m, line 3: Do we need this script here given how test-case specific it is ?

NuGet.Config
mregen: Quite handy if you need to test preview releases

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/endlos.cmd
mregen: normally I would not check it in, but since this is a very special issue and difficult to repro better keep itz. And its not in main

mregen:
Ha, need a third review !

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/endlos.cmd
karok2m: OK, agreed.

NuGet.Config
karok2m: OK. I guess it is quite fine for `2.5` as we only do maintenance on it.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Dockerfile.standalone
```diff
@@ -0,0 +1,6 @@
+﻿FROM mcr.microsoft.com/dotnet/core/runtime:3.1
+
+COPY ./publish /publish
```
karok2m, line 3: Looks like this `Dockerfile.standalone` relies on the other `Dockerfile` to perform the build. Can this please be changed to to also contain the build step itself so that it is self-sufficient ?

mregen: no, it is just for dockerbuild.bat, no build steps needed here

karok2m: ah, ok. i see the build and publish steps there.

karok2m (approved):

PR #1542, opened 2022-02-01
update GetConfiguredEndpoints_V1 in docs

monikavar:
GetConfiguredEndpoints_V1


docs/modules/publisher.md
```diff
@@ -274,16 +274,35 @@ UnpublishNodes method enables a client to remove nodes from a previously configu
 
 #### **GetConfiguredEndpoints_V1**
 
-*TODO: add description*
+Returns the configured endpoints (Datasets)
 
-  *Request:*
+  *Request:* N/A
```
cristipogacean: I think we should just say here that the payload is empty json { }. N/A might be a bit confusing.

monikavar: done

cristipogacean (approved):

karok2m (approved):

PR #1543, opened 2022-02-02
Implement UnpublishAllNodes_V1 direct method

monikavar:
This PR adds implementation of "UnpublishAll" Direct Method with unit tests and e2e tests.

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
@@ -743,7 +743,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine {
                 throw new MethodCallStatusException((int)HttpStatusCode.BadRequest, e.Message);
             }
             finally {
-                _logger.Information("{nameof} method finished in {elapsed}", nameof(PublishNodesAsync), sw.Elapsed);
+                _logger.Information("{nameof} method finished in {elapsed}", nameof(UnpublishNodesAsync), sw.Elapsed);
                 sw.Stop();
                 _lockConfig.Release();
             }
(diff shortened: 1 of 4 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -756,13 +756,89 @@ IJsonSerializer jsonSerializer
...
+
+                // Report error if no matching endpoint was found.
+                if (matchingGroups.Count == 0) {
+                    throw new MethodCallStatusException((int)HttpStatusCode.NotFound,
+                        $"Endpoint not found: {request.EndpointUrl}");
+                }
+
+                // Report error if there were entries that did not have any nodes
+                if (nodesToRemoveSet.Count == 0) {
+                    request.OpcNodes = nodesToRemoveSet.ToList();
```
karok2m: The if condition here already makes sure that `nodesToRemoveSet` is going to be an empty list, so this conversion to list doesn't add any value. Also, why do we need to change the request here just before throwing an exception ? I think that we do not need to have `nodesToRemoveSet` which is a `HashSet` at all. A simple bool flag would be sufficient to check if we have found any nodes or not.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -756,13 +756,89 @@ IJsonSerializer jsonSerializer
...
+                // Perform first pass to determine if we can find all nodes to remove.
+                var matchingGroups = new List<PublishedNodesEntryModel>();
+                foreach (var entry in _publishedNodesEntries) {
+                    if (entry.HasSameGroup(request)) {
+                        // We may have several entries with the same DataSetGroup definition,
+                        // so we will remove nodes only if the whole DataSet definition matches.
+                        if (IsSameDataSet(entry, request)) {
+                            entry.OpcNodes.Select(n => nodesToRemoveSet.Add(n));
+                            entry.OpcNodes.Clear();
+                            matchingGroups.Add(entry);
```
karok2m: We need to add entries to `matchingGroups` from the if above, where we check that entries have the same group. All entries with the same group will go to the same job definition so we need this list to update the affected job.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -756,13 +756,89 @@ IJsonSerializer jsonSerializer
...
+
+                // Report error if there were entries that did not have any nodes
+                if (nodesToRemoveSet.Count == 0) {
+                    request.OpcNodes = nodesToRemoveSet.ToList();
+                    var entriesNotFoundJson = _jsonSerializer.SerializeToString(request);
+                    throw new MethodCallStatusException(entriesNotFoundJson, (int)HttpStatusCode.NotFound, "Nodes not found");
+                }
+
+                // Remove entries without nodes.
+                _publishedNodesEntries.RemoveAll(entry => entry.OpcNodes.Count == 0);
```
karok2m: Please also add a check for `entry.OpcNodes` being null.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -756,13 +756,89 @@ IJsonSerializer jsonSerializer
...
+
+                PersistPublishedNodes();
+
+                var found = false;
+                await _lockJobs.WaitAsync(ct).ConfigureAwait(false);
+                try {
+                    if (!found) {
```
karok2m, line 845: This if itself is redundant as `found` is assigned `false` above.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -756,13 +756,89 @@ IJsonSerializer jsonSerializer
...
+                        var entryJobId = _publishedNodesJobConverter.
+                            ToConnectionModel(request, _legacyCliModel).CreateConnectionId();
+                        foreach (var assignedJob in _assignedJobs) {
```
karok2m, line 848: In reality we might have deleted only one data set from a group as a result of this call and not the whole group. So we first need to generate a job from `matchingGroups` and if there is a matching job for that update it. Like this:

```
                var jobs = _publishedNodesJobConverter.ToWriterGroupJobs(existingGroups, _legacyCliModel);
```

And only after we failed to find a matching job we should consider to delete a job. Please consult code of `UnpublishNodesAsync()` method for a similar approach.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Controller/PublisherMethodsController.cs
```diff
@@ -52,11 +52,6 @@ public class PublisherMethodsController : IMethodController {
         /// </summary>
         public async Task<PublishedNodesResponseApiModel> UnpublishAllNodesAsync(
             PublishNodesEndpointApiModel request) {
-
-            if (request.OpcNodes != null) {
```
karok2m, line 57: Let's keep this check as we should not have any `OpcNodes` in the request. I would actually change the check to `if (request.OpcNodes != null && request.OpcNodes.Count > 0)` just to allow for empty lists as well.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -756,13 +756,109 @@ IJsonSerializer jsonSerializer
...
+
+                // Report error if no matching endpoint was found.
+                if (matchingGroups.Count == 0) {
+                    throw new MethodCallStatusException((int)HttpStatusCode.NotFound,
+                        $"Endpoint not found: {request.EndpointUrl}");
+                }
+
+                // Report error if there were entries that did not have any nodes
+                if (!found) {
+                    var entriesNotFoundJson = _jsonSerializer.SerializeToString(request);
```
karok2m: Does it bring much value to report the request itself in the response here ? I would suggest to just throw an endpoint not found exception similar to the one above and even remove the check above for `matchingGroups.Count == 0` and only throw based on whether we found exact data set or not.

monikavar: Indeed this check could be omitted.

cristipogacean (approved):
:shipit:

PR #1545, opened 2022-02-03
add support for incremental publisher reconfig

cristipogacean:
* changed the publisher behavior when reconfigured to incrementally change the existing subscriptions/monitored items instead of deleting & recreating completely the configuration tree. 
* fixed bugs and deadlock on publisher module exit
* several cosmetics changes


common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
```diff
@@ -38,13 +38,6 @@ public class Worker : IWorker, IDisposable {
         /// <summary>
         /// Create worker
         /// </summary>
-        /// <param name="jobManagerConnector"></param>
```
dacolgit, line 41: why the `///<param>` were removed?  #Resolved

dacolgit: I just seen that you removed `///<param>` everywhere

dacolgit (approved):

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
cristipogacean: indeed, if they have no descriptions, they are useless.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/IMessageTrigger.cs
```diff
@@ -74,8 +74,11 @@ public interface IMessageTrigger : IDisposable {
...
         /// Run the group triggering
         /// </summary>
-        /// <param name="ct"></param>
-        /// <returns></returns>
         Task RunAsync(CancellationToken ct);
+
+        /// <summary>
+        /// Allow takeover of newly configured subscriptions
+        /// </summary>
+        void Reconfigure(object config);
```
karok2m, line 82: Why not get `WriterGroupJobModel` object directly here ? #Resolved

cristipogacean: We're in the context of a generic assembly. We cannot take dependencies here on publisher related assembly.

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
```diff
@@ -94,22 +87,26 @@ public class Worker : IWorker, IDisposable {
 
         /// <inheritdoc/>
         public async Task StopAsync() {
-            await _lock.WaitAsync().ConfigureAwait(false);
```
karok2m, line 97: Given that `StartAsync()` and `GetStatus()` rely on the `_lock` to make sure that all operations are synchronous I would suggest to revert back the `_lock` related changes here. With the current changes we open-up opportunities for fast call to `StopAsync()` to not perform intended operations if there was a `StartAsynch()` executing at the same time. I understand that these kind of rapid calls might be exceptional cases, but the whole point of having `_lock` here seem to be to safeguard against that kind of situations. #Resolved

karok2m: The other possible error that I can see is that we might assign `_jobProcess` to `null` (on line `105`) at the same time as `GetStatus()` was evaluating `_jobProcess?.Status` operation. So the `?` check might see that `_jobProcess` is not null, but when it tries to access its `Status` member the `_jobProcess` might already be null and thus cause a `NullReferenceException`.

cristipogacean: nope, the reason for this change was a super ugly deadlock and exactly because of the lock on the `Status` in `SendHeartbeatWithoutResetTimerAsync`

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
monikavar: so should we add description of the parameters?

cristipogacean: only when the name is not self-describing. And even then, we should first try to find a better name :)

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
```diff
@@ -94,22 +87,26 @@ public class Worker : IWorker, IDisposable {
...
-                    await SendHeartbeatWithoutResetTimerAsync().ConfigureAwait(false); // need to be send before cancel the CancellationToken
-                }
+            // Inform services, that this worker has stopped working,
+            // so orchestrator can reassign the job
+            if (_jobProcess != null && _jobProcess.Status != WorkerStatus.Stopped) {
+                _jobProcess.Status = WorkerStatus.Stopped;
+                // need to be send before cancel the CancellationToken
+                await SendHeartbeatWithoutResetTimerAsync().ConfigureAwait(false);
+                _jobProcess.Dispose();
+                _jobProcess = null;
```
karok2m: Those 4 lines of code appear here and twice in `ProcessAsync()`. Maybe we can create a dedicated `ResetJobProcessAsync()` method for them ? #Resolved

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
```diff
@@ -157,20 +154,21 @@ public class Worker : IWorker, IDisposable {
...
         private async void HeartbeatTimer_ElapsedAsync(object sender) {
-
             if (!_cts.IsCancellationRequested) {
                 await SendHeartbeatWithoutResetTimerAsync().ConfigureAwait(false);
                 Try.Op(() => _heartbeatTimer.Change(_heartbeatInterval, Timeout.InfiniteTimeSpan));
             }
         }
 
+        /// <summary>
+        /// Send the a new heartbeat
```
monikavar: nit: "the" or "a" removal. #Resolved

monikavar (approved):
:shipit:

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
```diff
@@ -223,72 +220,66 @@ public class Worker : IWorker, IDisposable {
...
-                    using (_jobProcess = new JobProcess(this, jobProcessInstruction,
-                        _lifetimeScope, _logger)) {
-                        await _jobProcess.WaitAsync(ct).ConfigureAwait(false); // Does not throw
+                    if (_jobProcess == null) {
+                        _jobProcess = new JobProcess(this, jobProcessInstruction, _lifetimeScope, _logger);
                     }
+                    else {
+                        _jobProcess.ProcessNewInstruction(jobProcessInstruction);
+                    }
+                    await _jobProcess.WaitAsync(ct).ConfigureAwait(false); // Does not throw
```
karok2m, line 268: Is this operation going to complete at some point ? Or is it intended to wait until cancelled or reconfigured ? #Resolved

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
```diff
@@ -223,72 +220,66 @@ public class Worker : IWorker, IDisposable {
...
+                        _jobProcess?.JobContinuation?.ProcessMode == null) &&
+                        _jobProcess.Status != WorkerStatus.Stopped) {
                         _jobProcess.Status = WorkerStatus.Stopped;
                         await SendHeartbeatWithoutResetTimerAsync().ConfigureAwait(false);
+                        _jobProcess.Dispose();
                         _jobProcess = null;
                         break;
                     }
 
                     jobProcessInstruction = _jobProcess.JobContinuation;
```
karok2m: Why do we assign a new value to input `jobProcessInstruction` parameter ? If it is only for logging, I would suggest to use `_jobProcess.JobContinuation` instead. #Resolved

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
cristipogacean: it's intended to be cancelled

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
cristipogacean: I did not touch that. the PR was intended to fix an issue and I tried to keep the refactoring minimal. we can decide to refactor more in some other PR.

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
cristipogacean: yes, let's do that when we will decide to do a major refactoring of this stuff.

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/Worker.cs
karok2m: Let's just use `_jobProcess.JobContinuation` in the logs bellow and get rid of this assignment.

cristipogacean: that would crash since _jobProcess is disposed at that point in time ...

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/cli/Properties/launchSettings.json
```diff
@@ -1,7 +1,8 @@
 {
   "profiles": {
     "Microsoft.Azure.IIoT.Modules.OpcUa.Publisher.Cli": {
-        "commandName": "Project"
+        "commandName": "Project",
+        "commandLineArgs": "--PkiRootPath=/mount/pki --pf=/mount/pn.json --fd=false --fm=false --bs=1 --di=20 --to --aa --si=1 --mm=PubSub --dm=2"
```
karok2m: Looks like a remnant from debugging or testing. Please remove.

cristipogacean: yeah

PR #1554, opened 2022-02-08
Publish/unpublish/addorpupdate/removeall nodes  empty response

cristipogacean:
Changes
* simplify the response payload for publish/unpublish/addorpupdate/removeall nodes Direct methods. Now the response contains an empty json in case of success.


api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/PublishedNodesResponseApiModel.cs
```diff
@@ -4,23 +4,11 @@
...
-    using System.Collections.Generic;
-    using System.ComponentModel.DataAnnotations;
-    using System.Runtime.Serialization;
-    
 
     /// <summary>
     /// PublishNodes direct method response
     /// </summary>
-    [DataContract]
     public class PublishedNodesResponseApiModel {
```
dacolgit, line 11: why we need this empty class?

monikavar (commented):
I think it's better to keep List of string in return type. It might not provide critical info currently but in the future, we would have the liberty to add information related to publishing status of each Opcnode which might be useful for the customers. It would not be ideal to change the API later on. What do you think?

cristipogacean:
@monikavar, can you be a bit more specific with the use-case(s) you have in mind for each of the 4 methods?
How do you envision encoding 'information related to publishing status of each Opcnode' as a list of strings? Some examples for the different dm's payloads would be super helpful!

dacolgit (approved):

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/PublishedNodesResponseApiModel.cs
```diff
@@ -4,23 +4,11 @@
...
 
-        /// <summary>
-        /// Status messages
-        /// </summary>
-        [DataMember(Name = "statusMessage", Order = 0)]
-        [Required]
-        public List<string> StatusMessage { get; set; }
```
karok2m, line 23: Don't we at least want to keep one string to communicate exceptions ? Can you please also change the E2E tests in this PR as well so that we see hot this ends up on that end as well ? Would also be useful if we add some calls in E2E tests that we know should throw exceptions, such as removing nonexistent endpoints, and validate the responses in that cases.

cristipogacean: The DM call exceptions are handled exclusively by throwing `MethodCallStatusException`. This one has a field for description besides the (Bad) status. 
Right, extending E2E with bad things is a good idea.

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/PublishedNodesResponseApiModel.cs
cristipogacean: I believe the method router requires a class as response payload. Even if the class is empty.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -635,11 +633,10 @@ IJsonSerializer jsonSerializer
                 _lockConfig.Release();
             }
```
karok2m, line 613: Extra empty line can also be removed.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
```diff
@@ -803,17 +798,15 @@ IJsonSerializer jsonSerializer
...
```
karok2m, line 778: Empty line can be removed.

karok2m (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/LegacyJobOrchestrator.cs
cristipogacean: done

PR #1562, opened 2022-02-10
Enable opcpublisher to set ContentType to "application/json"

cristipogacean:
for the backwards compatibility with publisher 2.5.x, `legacycompatibility` cli argument was introduced.


modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/LegacyCliOptions.cs
```diff
@@ -160,6 +160,9 @@ public class LegacyCliOptions : Dictionary<string, string>, IAgentConfigProvider
                     { "me|messageencoding=", "The message encoding for messages " +
                         $"(allowed values: {string.Join(", ", Enum.GetNames(typeof(MessageEncoding)))}).",
                         (MessageEncoding m) => this[LegacyCliConfigKeys.MessageEncoding] = m.ToString() },
+                    { "lc|legacycompatibility=", "Run the publisher in legacy (2.5.x) compatibility mode. " +
```
karok2m: `legacycompatibility` sounds very generic. Let's use a name that would make it clear that this is going to change only content type that is set in the messages. With the current name some people might think that this is a full compatibility mode and expect too much, such as old direct methods or telemetry formatting.

cristipogacean: It is intended like this so we can host some other future 2.5.x particularities under the same umbrella. I was thinking about fullfeaturemessage for standalone, but it's a bit late to discontinue it now.

karok2m: OK, understood.

karok2m (dismissed):

karok2m (approved):

PR #1574, opened 2022-02-15
Change the diagnosticInfo.Id to an DataSetWriter/Endpoint 

dacolgit:
Added endpoint info in diagnostic model
added diagnosticApiModel
fix doc


api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/DiagnosticInfoApiModel.cs
```diff
@@ -0,0 +1,132 @@
...
+    /// <summary>
+    /// Model for a diagnostic info response.
+    /// </summary>
+    [DataContract]
+    public class DiagnosticInfoApiModel {
+ 
+        /// <summary> EndpointInfo /// </summary>
+        [DataMember(Name = "PublishedNodesEntryModel", Order = 0,
+            EmitDefaultValue = false)]
+        public PublishedNodesEntryModel EndpointInfo { get; set; }
```
karok2m: Response API model should be using `PublishNodesEndpointApiModel` instead of `PublishedNodesEntryModel`. `PublishedNodesEntryModel` is the internal class and `PublishNodesEndpointApiModel` is the one intended for external communication.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/StandaloneJobOrchestrator.cs
```diff
@@ -170,6 +170,24 @@ IJsonSerializer jsonSerializer
 
                         if (diagInfo.Id == assignedJob.Value.Job.Id) {
                             _publisherDiagnosticInfo.AddOrUpdate(assignedJob.Value.Job.Id, diagInfo);
+
+                            foreach (var entry in _publishedNodesEntries) {
+                                var id = _publishedNodesJobConverter.
+                                    ToConnectionModel(entry, _standaloneCliModel).CreateConnectionId();
+
+                                if (_publisherDiagnosticInfo[diagInfo.Id].Id == id) {
```
karok2m: `_publisherDiagnosticInfo[diagInfo.Id].Id` is the same as just `diagInfo.Id`.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/StandaloneJobOrchestrator.cs
```diff
@@ -170,6 +170,24 @@ IJsonSerializer jsonSerializer
...
+                                    _publisherDiagnosticInfo[id].EndpointInfo = new PublishedNodesEntryModel {
```
karok2m: The main diagnostic reporting logic is not using `EndpointInfo` at all. We only need it here where we try to find corresponding `PublishedNodesEntryModel` object for each `JobDiagnosticInfoModel` object. So I would suggest to remove `EndpointInfo` field from `JobDiagnosticInfoModel` and instead change `_publisherDiagnosticInfo` to be a dictionary from string `jobId` to `Tuple<JobDiagnosticInfoModel, PublishedNodesEntryModel>`.

This way we can keep the information about corresponding objects and not require cross-project dependencies that were introduced in the PR. In this case the return type from `GetDiagnosticInfoAsync()` would change to `List<Tuple<JobDiagnosticInfoModel, PublishedNodesEntryModel>>` and then the `ToApiModel()` method can take care of transforming those into `DiagnosticInfoApiModel`.

e2e-tests/IIoTPlatform-E2E-Tests/Standalone/C_PublishMultipleEndpointsStandaloneDirectMethodTestTheory.cs
```diff
@@ -234,7 +234,7 @@ IIoTMultipleNodesTestContext context
             ).ConfigureAwait(false);
 
             Assert.Equal((int)HttpStatusCode.OK, responseGetDiagnosticInfo.Status);
-            var diagInfoList = _serializer.Deserialize<List<JobDiagnosticInfoModel>>(responseGetDiagnosticInfo.JsonPayload);
+            var diagInfoList = _serializer.Deserialize<List<DiagnosticInfoApiModel>>(responseGetDiagnosticInfo.JsonPayload);
```
karok2m, line 237: Since response should contain endpoint definition now, we should validate it as well.

dacolgit: added

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/DiagnosticInfoApiModel.cs
dacolgit: changed

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/StandaloneJobOrchestrator.cs
dacolgit: changed

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/StandaloneJobOrchestrator.cs
dacolgit: removed `EndpointInfo` from `JobDiagnosticInfoModel`.
I also remove the  cross-project dependencies. 
`_publisherDiagnosticInfo` is now  `Dictionary<string, Tuple<JobDiagnosticInfoModel, PublishedNodesEntryModel>>`.
Using a Dictionary instead of List I was able to keep all the dictionary properties used to manage the JobDiagnosticInfoModel

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/StandaloneJobOrchestrator.cs
```diff
@@ -169,7 +169,24 @@ IJsonSerializer jsonSerializer
                     foreach (var assignedJob in _assignedJobs) {
 
                         if (diagInfo.Id == assignedJob.Value.Job.Id) {
-                            _publisherDiagnosticInfo.AddOrUpdate(assignedJob.Value.Job.Id, diagInfo);
+                            foreach (var entry in _publishedNodesEntries) {
+                                var id = _publishedNodesJobConverter.
```
karok2m: Now lookin at this again I understand that `jobId` is going to be determined only by the following fields:

* `EndpointUrl`
* `UseSecurity`
* `OpcAuthenticationMode` together with `OpcAuthenticationUsername` and `OpcAuthenticationPassword` 
* `DataSetWriterGroup`

So `DataSetWriterId` and `DataSetPublishingInterval` do not influence `jobId`. This makes sense given that only the above fields are going to define `jobId` and corresponding session details. And fields such as `DataSetWriterId` and `DataSetPublishingInterval` are only going to affect names of subscriptions within the session. What this means is that there might be several entries in `_publishedNodesEntries` list which would result in the same `jobId`. So only taking the first one does not seem to be the correct approach, neither does taking all of them. So maybe we should take only the fields whose values influence the `jobId`, meaning that we exclude `DataSetWriterId` and `DataSetPublishingInterval` from the assignment bellow. Then all endpoint definitions that match on the fields that we take would be part of the diagnostics info. @dacolgit, @cristipogacean, what you guys think about this ?

The other problem that we have here is that we are iterating over `_publishedNodesEntries` now, which means that we have to use `_lockConfig` together with `_lockJobs` instead of only having `_lockJobs` on line `117` above. This is problematic because we introduced `_lockJobs` so that heartbeat reporting can happen faster and not be affected as much by direct method calls. @cristipogacean, is there a way to get the meaningful fields that I've mentioned above from the job definition itself ?

dacolgit: Extended the messageTrigger to export the following info:
- `EndpointUrl
- UseSecurity
- OpcAuthenticationMode together with OpcAuthenticationUsername and OpcAuthenticationPassword
- DataSetWriterGroup`

As discussed the `DataSetWriterId` and 'DataSetPublishingInterval' are removed from the 'EndpointDiagnosticInfo'

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/IPublisherConfigServices.cs
```diff
@@ -6,6 +6,7 @@
 namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher {
     using Microsoft.Azure.IIoT.Agent.Framework.Models;
     using Microsoft.Azure.IIoT.OpcUa.Publisher.Config.Models;
+    using System;
```
karok2m: This seem to be the only addition to the file. Is it necessary ? My assumption is that it was compiling previously.

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Models/JobDiagnosticInfoModel.cs
```diff
@@ -11,11 +11,23 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Models {
     /// </summary>
     public class JobDiagnosticInfoModel {
 
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        public JobDiagnosticInfoModel() {
+            EndpointInfo = new EndpointDiagnosticModel();
+        }
+
```
karok2m: We can remove the constructor. `null` for `EndpointDiagnosticModel` does seem like a valid value until it is set.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
```diff
@@ -164,6 +163,11 @@ public class DataFlowProcessingEngine : IProcessingEngine, IDisposable {
             double estimatedMsgChunksPerDay = Math.Ceiling(chunkSizeAverage) * sentMessagesPerSec * 60 * 60 * 24;
             var diagnosticInfo = new JobDiagnosticInfoModel();
 
+            diagnosticInfo.EndpointInfo.EndpointUrl = _messageTrigger.EndpointUrl;
```
karok2m: Here you can first create a `EndpointDiagnosticModel` object, set its fields and then assign it to `diagnosticInfo.EndpointInfo`. That way there is no need to have a special constructor for `JobDiagnosticInfoModel`.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/IPublisherConfigServices.cs
dacolgit: removed

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Models/JobDiagnosticInfoModel.cs
dacolgit: I added the constructor to avoid null reference exception in `GetDiagnosticInfo` in DataFlowProcessingEngine

dacolgit: just read your comment below. Constructor removed

cristipogacean (approved):
:shipit:

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/WriterGroupMessageSource.cs
```diff
@@ -507,6 +537,14 @@ private sealed class DataSetWriterSubscription : IDisposable {
             private long _currentSequenceNumber;
         }
 
+        /// <summary>
+        /// Helper to deserialize credential info
+        /// </summary>
+        private class cred {
```
karok2m, line 543: Can we make this into a public class and use it whenever we serialize/deserialize `CredentialModel.Value` for `CredentialType.UserName` case ? I can see 3 places now where we hardcode the expected structure:
* in `PublishedNodesJobConverter.ToUserNamePasswordCredentialAsync()` method
* in `StackModelsEx.ToServiceModel()` method
* and now here

monikavar (approved):

PR #1580, opened 2022-02-18
Retry to set monitored items to reporting in case of failure

cristipogacean:
Changes:
* Retry to set monitored items to reporting in case of failure.
* Small cosmetic changes.

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
@@ -1008,8 +1008,8 @@ namespace Microsoft.Azure.IIoT.OpcUa.Protocol.Services {
 
             /// <inheritdoc/>
             public override string ToString() {
-                return $"Item {Template.Id ?? Template.StartNodeId}_{ServerId}: '{Template.StartNodeId}'" +
-                    $" - {(Item?.Status?.Created == true ? "" : "not ")}created";
+                return $"Item '{Template.StartNodeId}' with server id {ServerId} - " +
+                    $"{(Item?.Status?.Created == true ? "" : "not ")} created";
             }
 
             /// <summary>
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
@@ -76,7 +76,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Protocol.Services {
                 get {
                     return (IsConnectionOk && _currentlyMonitored != null) ?
                         _currentlyMonitored
-                            .Where(x => x.Template.MonitoringMode != Publisher.Models.MonitoringMode.Disabled)
+                            .Where(x => x.Item.Status.MonitoringMode != Opc.Ua.MonitoringMode.Disabled)
                             .Count() : 0;
                 }
             }
@@ -86,7 +86,7 @@ namespace Microsoft.Azure.IIoT.OpcUa.Protocol.Services {
                 get {
                     return (IsConnectionOk && _currentlyMonitored != null) ?
                         _currentlyMonitored
-                            .Where(x => x.Template.MonitoringMode == Publisher.Models.MonitoringMode.Disabled)
+                            .Where(x => x.Item.Status.MonitoringMode == Opc.Ua.MonitoringMode.Disabled)
                             .Count() : 0;
                 }
             }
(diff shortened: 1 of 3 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -76,7 +76,7 @@ internal sealed class SubscriptionWrapper : ISubscription {
                 get {
                     return (IsConnectionOk && _currentlyMonitored != null) ?
                         _currentlyMonitored
-                            .Where(x => x.Template.MonitoringMode != Publisher.Models.MonitoringMode.Disabled)
+                            .Where(x => x.Item.Status.MonitoringMode != Opc.Ua.MonitoringMode.Disabled)
```
marcschier: I think this is ok, but I would suggest testing it against the desired mode, i.e., if the desired mode is Disabled, it is actually ok.  If the desired mode != actual, then we are not monitoring.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -540,8 +540,10 @@ internal sealed class SubscriptionWrapper : ISubscription {
                 else {
                     // do a sanity check
                     foreach (var monitoredItem in _currentlyMonitored) {
-                        if (monitoredItem.Item.Status.Error != null &&
-                            StatusCode.IsNotGood(monitoredItem.Item.Status.Error.StatusCode)) {
+                        if (monitoredItem.Item.Status.MonitoringMode == Opc.Ua.MonitoringMode.Disabled ||
```
marcschier, line 543: See above.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -584,12 +587,15 @@ internal sealed class SubscriptionWrapper : ISubscription {
                                     if (StatusCode.IsNotGood(results[i].StatusCode)){
                                         _logger.Warning("Set monitoring for item '{item}' in subscription "
                                             + "'{subscription}'/'{sessionId}' failed with '{status}'.",
-                                            itemsToChange[i].StartNodeId, 
+                                            itemsToChange[i].StartNodeId,
                                             Id,
                                             Connection.CreateConnectionId(),
                                             results[i].StatusCode);
+                                        changeList[i].Template.MonitoringMode = Publisher.Models.MonitoringMode.Disabled;
```
marcschier, line 594: Again, all ok, but I wonder whether null is appropriate monitoring mode for "failure case"?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
cristipogacean: Ok, I'll change this, it makes sense. The original thought behind my implementation was that the publisher's desired state is Reporting all the time.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
cristipogacean: I guess the idea was that "Disabled" state is actually the "failure case" since the publisher shall always force all the monitored items to "Reporting".

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -1002,7 +1008,7 @@ internal class MonitoredItemWrapper {
 
             /// <inheritdoc/>
             public override string ToString() {
-                return $"Item {Template.Id ?? "<unknown>"}{ServerId}: '{Template.StartNodeId}'" +
+                return $"Item {Template.Id ?? Template.StartNodeId}_{ServerId}: '{Template.StartNodeId}'" +
```
karok2m: Do we need `_{ServerId}` part here any more ? Previously it was used to show the incremental number that server associated with the node and coupled with `<unknown>` provided a way to differentiate it from the others. But as we can see in the logs provided in the support bundle `Template.StartNodeId` has a valid value, so the log that previously looked like this:

```log
Item <unknown>1: 'ns=1;s=t|MC_EXP01-WTR-TMP-TC.Result_FD_Units' - created: Changing monitoring mode from Disabled to Reporting
```

Will now look like this:

```log
Item ns=1;s=t|MC_EXP01-WTR-TMP-TC.Result_FD_Units_1: 'ns=1;s=t|MC_EXP01-WTR-TMP-TC.Result_FD_Units' - created: Changing monitoring mode from Disabled to Reporting
```

Note the `_1:`, which would be confusing since that looks like a part of node id now. I would suggest to remove `_{ServerId}` part now that we are using `Template.StartNodeId` instead of `<unknown>`.

karok2m (dismissed):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
cristipogacean: Good point, it's suboptimal. I still think the ServerId is a relevant piece of information that can help troubleshooting, so I removed the `{Template.Id}` and kept the ServerId.

karok2m (approved):

PR #1583, opened 2022-02-21
Fix publisher heartbeat interval inconsistent handling

cristipogacean:
* make sure the default values for `HeartbeatInterval` respective `HeartbeatIntervalTimespan` are **not** persisted it in the pn.json.
* make sure only one of the `HeartbeatInterval` respectivelty `HeartbeatIntervalTimespan` (this wins when both values provided) is consistently used in job definitions. Timespan variant wins when both values are provided.
* report consistently the `HeartbeatInterval`, as originally configured in `GetConfiguredNodesOnEndpoint_V1` DM.

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Extensions/OpcNodeModelEx.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Extensions/OpcNodeModelEx.cs
@@ -58,7 +58,19 @@ namespace Microsoft.Azure.IIoT.OpcUa.Publisher.Config.Models {
                 return false;
             }
 
-            if (model.HeartbeatInterval != that.HeartbeatInterval) {
+            var modelHeartbeatInterval = model.HeartbeatIntervalTimespan.HasValue
+                ? model.HeartbeatIntervalTimespan
+                : model.HeartbeatInterval.HasValue
+                    ? TimeSpan.FromSeconds(model.HeartbeatInterval.Value)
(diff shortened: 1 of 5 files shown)


marcschier (dismissed):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Extensions/OpcNodeModelEx.cs
```diff
@@ -58,7 +58,19 @@ public static class OpcNodeModelEx {
                 return false;
             }
 
-            if (model.HeartbeatInterval != that.HeartbeatInterval) {
+            var modelHeartbeatInterval = model.HeartbeatIntervalTimespan.HasValue
+                ? model.HeartbeatIntervalTimespan
+                : model.HeartbeatInterval.HasValue
+                    ? TimeSpan.FromSeconds(model.HeartbeatInterval.Value)
+                    : (TimeSpan?)null;
```
karok2m: Can we have an extension method in `OpcNodeModelEx.cs` that will contain this logic ? That way we do not need to repeat it in several places.

karok2m (dismissed):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Extensions/OpcNodeModelEx.cs
cristipogacean: Good idea. Done

karok2m (approved):

PR #1587, opened 2022-02-21
Changed X509 certificate validation to use offline certificate revocation list.

karok2m:
Changed X509 certificate validation to use [`Offline`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509revocationmode?view=net-6.0) revocation check mode in which case the check is made using a cached certificate revocation list. Previously certificate revocation list check was disabled (by `validator.ChainPolicy.RevocationMode = X509RevocationMode.NoCheck`) and it was resulting in the following warning from Roslyn analyzers:

```
##[warning]components\opc-ua\src\Microsoft.Azure.IIoT.OpcUa\src\Core\Extensions\X509CertificateChainModelEx.cs(78,21): Warning IA5352: Do Not Change X509ChainPolicy. RevocationMode to NoCheck
```

Source for opportunistic revocation check is [here](https://stackoverflow.com/a/53378487/1451497).


karok2m:
Warning can be seen in this build from `main`: https://msazure.visualstudio.com/One/_build/results?buildId=52190183&view=results
Warning is fixed in the build form this branch: https://msazure.visualstudio.com/One/_build/results?buildId=52193188&view=results

cristipogacean (approved):

karok2m:
After a discussion with @marcschier we have decided to merge this PR after release of `2.8.2`.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Core/Extensions/X509CertificateChainModelEx.cs
```diff
@@ -75,7 +75,11 @@ public static class X509CertificateChainModelEx {
                 validator.ChainPolicy.RevocationFlag =
                     X509RevocationFlag.EntireChain;
                 validator.ChainPolicy.RevocationMode =
-                    X509RevocationMode.NoCheck;
+                    X509RevocationMode.Offline;
```
karok2m, line 78: Comment from @mregen: `X509RevocationMode.Offline` should not be used as there are strange particularities in the implementation, such as cert validator tries to reach a OCSP responder or CRL.

karok2m, line 78: Comment from @mregen: OPC Twin module should use certificate validator from OPC UA Stack. This custom certificate validator is very likely not up to date to security fixes

mregen: I think in this case its a simple check if a cert chain in a binary blob is really properly signed from top to bottom. So no need for the UA stack validator, also no need to check any revocation

mregen: see my comment below, NoCheck is the right choice here.

mregen:
I checked the codebase, the GetEndpoints may contain an array of certificates, but no revocation information. So the chain validation is ok here and the validator can remain in place but all revocation checks should be disabled..

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Core/Extensions/X509CertificateChainModelEx.cs
mregen: change back to NoCheck and disable the warning...

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Core/Extensions/X509CertificateChainModelEx.cs
```diff
@@ -75,7 +75,11 @@ public static class X509CertificateChainModelEx {
                 validator.ChainPolicy.RevocationFlag =
                     X509RevocationFlag.EntireChain;
                 validator.ChainPolicy.RevocationMode =
```
mregen, line 77: @karok2m on .NET6, set also DisableCertificateDownload property

mregen (changes requested):
leave the bits as NoCheck but set  DisableCertificateDownload property.

mregen (changes requested):
do not change revocation mode to offline, it may cause unexpected behavior and add disablecertificatedownload

marcschier (approved):

mregen (changes requested):
do not merge with Offline check!

PR #1593, opened 2022-02-23
Add non-versioned name to the publisher's DM API as fallback

cristipogacean:
The publisher shall respond to the DM names without _V1 as a fallback. The recommended way is still with "_V1" suffix

### DIFF
--- a/common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/MethodRouter.cs
+++ b/common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/MethodRouter.cs
@@ -99,7 +99,7 @@ namespace Microsoft.Azure.IIoT.Module.Framework.Hosting {
         /// <param name="target"></param>
         private void AddToCallTable(object target) {
             var versions = target.GetType().GetCustomAttributes<VersionAttribute>(true)
-                .Select(v => "_v" + v.Value)
+                .Select(v => v.Value)
                 .ToList();
             if (versions.Count == 0) {
                 versions.Add(string.Empty);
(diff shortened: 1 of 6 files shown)


monikavar:
could we verify via tests that DMs are working without "_V1" suffix?

cristipogacean:
> could we verify via tests that DMs are working without "_V1" suffix?

@monikavar, we could, but not in this PR. For now manual testing is enough.

common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/MethodRouter.cs
```diff
@@ -99,7 +99,7 @@ public sealed class MethodRouter : IMethodRouter, IMethodHandler {
         /// <param name="target"></param>
         private void AddToCallTable(object target) {
             var versions = target.GetType().GetCustomAttributes<VersionAttribute>(true)
-                .Select(v => "_v" + v.Value)
```
karok2m, line 102: Is direct method name comparison case-sensitive or not ? I am asking because `v` here is lowercase while the `V` in the new values in all controllers is uppercase.

karok2m: I see that in our E2E tests we have used uppercase `V` and it worked, so the comparison should be case-insesitive.

karok2m (approved):

common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/MethodRouter.cs
karok2m, line 102: @monikavar, can we please add in your documentation changes that both versioned and non-versioned DM names are supported, but we strongly recommend to use versioned names.

cristipogacean: it's insensitive

monikavar: I have already added this in the PR.

karok2m: Thank you.

PR #1597, opened 2022-02-24
Fix order of DataSetPublishing interval

cristipogacean:
* the Node Publishing Interval configuration prevails now the one at Data Set level
* unit-tests adapted


dacolgit:
> //  Copyright (c) Microsoft Corporation.  All rights reserved.

no changes in the file

---
Refers to: e2e-tests/IIoTPlatform-E2E-Tests/TestModels/PublishedNodesEntryModel.cs:2 in a769a6b. [](commit_id = a769a6bb9cb153f45348f83a484f9842c4df189b, deletion_comment = False)

dacolgit (dismissed):
:shipit:

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/StandaloneJobOrchestrator.cs
```diff
@@ -1087,7 +1087,10 @@ IJsonSerializer jsonSerializer
                     OpcAuthenticationUsername = model.OpcAuthenticationUsername,
                     DataSetWriterGroup = model.DataSetWriterGroup,
                     DataSetWriterId = model.DataSetWriterId,
-                    DataSetPublishingInterval = model.DataSetPublishingInterval,
+                    DataSetPublishingIntervalTimespan = model.DataSetPublishingIntervalTimespan,
+                    DataSetPublishingInterval = model.DataSetPublishingIntervalTimespan.HasValue
```
karok2m: Shouldn't this if/else be the other way around ? Now if `DataSetPublishingIntervalTimespan` has a value, then `DataSetPublishingInterval` would also be assigned a value. And if the first one does not have a value, then value of `model.DataSetPublishingInterval` would be ignored.

cristipogacean: right ... thanks!

karok2m (approved):

PR #1582, opened 2022-02-18
adding migration path and modifying other docs

monikavar:
Changes:
- Updated docs with direct methods for version `2.8.2` and included migration changes from `2.5.x` to `2.8.2`.
- Added requests and responses for all the supported direct methods for `2.5.x` and `2.8.2`.

PR #1598, opened 2022-02-24
Ensuring backward-compatible behavior for UnpublishNodes direct method.

karok2m:
Changes:
* Ensuring backward-compatible behavior for `UnpublishNodes` direct method with `2.5.x` version when `null` or empty `OpcNodes` is provided. In that case whole matching data set will be removed instead of throwing an exception.
* Added tests for new behacior.


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/StandaloneJobOrchestrator.cs
```diff
@@ -621,17 +621,18 @@ IJsonSerializer jsonSerializer
...
-
+            if (request is null) {
                 _logger.Information("{nameof} method finished in {elapsed}", nameof(UnpublishNodesAsync), sw.Elapsed);
                 sw.Stop();
 
-                throw new MethodCallStatusException((int)HttpStatusCode.BadRequest, message);
+                throw new MethodCallStatusException((int)HttpStatusCode.BadRequest, kNullRequestMessage);
             }
 
+            // When no node is specified then remove the whole data set.
```
karok2m, line 631: Update documentation and spec once #1582 is merged.

cristipogacean (dismissed):
:shipit:

monikavar (dismissed):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/StandaloneJobOrchestrator.cs
karok2m: Added note for `UnpublishNodes_V1` direct method detailing this behavior.

cristipogacean (approved):

PR #1600, opened 2022-02-25
fix sporadic timer ObjectDisposedException

cristipogacean:
* validate if the timer was disposed before using it
* tuned _healthCheckManager.IsReady for better reporting

### DIFF
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
@@ -49,8 +49,6 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Agent {
         /// <inheritdoc/>
         public async Task StartAsync() {
             await EnsureWorkersAsync();
-            _healthCheckManager.IsReady =
-                _instances.Count >= (_agentConfigProvider.Config?.MaxWorkers ?? kDefaultWorkers);
             _ensureWorkerRunningTimer.Enabled = true;
             _agentConfigProvider.OnConfigUpdated += ConfigUpdate_Triggered;
         }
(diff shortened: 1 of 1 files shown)


dacolgit (dismissed):
:shipit:

dacolgit (approved):

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
```diff
@@ -102,47 +104,53 @@ public class WorkerSupervisor : IWorkerSupervisor, IDisposable {
...
-            await worker.Key.StopAsync();
+            await worker.Key?.StopAsync();
             worker.Value?.Dispose();
         }
 
         /// <inheritdoc/>
         public void Dispose() {
             if (_ensureWorkerRunningTimer != null) {
-                _ensureWorkerRunningTimer.Stop();
+                _ensureWorkerRunningTimer.Enabled = false;
```
karok2m, line 119: Why is there no need to call `_ensureWorkerRunningTimer.Stop()` now ?

karok2m: Ah, because `Stop()` just sets `Enabled` to `false`.

karok2m (approved):

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Agent/Default/WorkerSupervisor.cs
cristipogacean: _ensureWorkerRunningTimer.Stop() is actually the same as: _ensureWorkerRunningTimer.Enabled = false. I made this change consistently through this class

PR #1606, opened 2022-02-25
Updated third party notice and wording in docs.

monikavar:
add third party notice separately and wording in docs

### DIFF
--- a/docs/modules/publisher-migrationpath.md
+++ b/docs/modules/publisher-migrationpath.md
@@ -1,9 +1,9 @@
 [Home](../../readme.md)
 
-# Migration path for publisher version 2.5.x to 2.8.2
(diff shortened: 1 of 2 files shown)


docs/modules/publisher-migrationpath.md
```diff
@@ -1,9 +1,9 @@
 [Home](../../readme.md)
 
-# Migration path for publisher version 2.5.x to 2.8.2
+# Migration path for OPC Publisher v2.5.x to v2.8.2
```
karok2m: Let's not use `v` in front of OPC Publisher version numbers. Our tags are without any prefixes.

monikavar: I was using for version short form only and not for tag names. Let me rephrase it.

karok2m (approved):

PR #1607, opened 2022-02-28
Update OPC Publisher documentation

hansgschossmann:
Increase AcroLinx score from < 80 to almost > 95 on all documents
Other changes
Fix typos


docs/modules/publisher-commandline.md
```diff
@@ -104,8 +107,8 @@ The following list of module parametrization settings can be provided either as
                                       Default: {00:00:00} meaning heartbeat is disabled
 
             DefaultSkipFirst=VALUE
-                                      Default for Skip First published node setting. When set to true the 
-                                      publisher will not send as telemetry the initial datachange event received
+                                      Default for SkipFirst published node setting. When set to true OPC Publisher
```
hansgschossmann: do we support SkipFirst in 2.8.2? I have not found any code for this. #Closed

hansgschossmann:
>       "QueueSize": "integer"

Is this still supported? I have not seen any code.

---
Refers to: docs/modules/publisher-directmethods.md:56 in 6f84244. [](commit_id = 6f842448c2a6dd12ffc04bcb7bc54de0e8311262, deletion_comment = False)

docs/modules/publisher-directmethods.md
```diff
@@ -65,46 +61,48 @@ The `_V1` direct methods  uses the  payload schema as described below:
...
+| `HeartbeatIntervalTimespan`     | No        | String  | null    | The interval used for the node to publish a value (a publisher <br>cached one) even if the value hasn't been <br>changed at the source. This value is represented in seconds. <br>Value expressed in Timespan string({d.hh:mm:dd.fff}).                                                                          |
+| `SkipFirst`                     | No        | Boolean | false   | Instructs the publisher not to add to telemetry the<br> Initial DataChange (after subscription activation) for this OpcNode.                                                                                                                                                                                      |
+| `QueueSize`                     | No        | Integer | 1       | The desired QueueSize for the monitored item to be published.                                                                                                                                                                                                                                                     |
```
hansgschossmann: Is this still supported? #Closed

karok2m: You mean `QueueSize` ?

docs/modules/publisher-directmethods.md
```diff
@@ -65,46 +61,48 @@ The `_V1` direct methods  uses the  payload schema as described below:
...
+| `DataSetPublishingInterval`         | no        | Integer | null                      | The publishing interval used for a grouped set of nodes <br>under a certain DataSetWriter. When defined it <br>overrides the OpcPublishingInterval value in the OpcNodes <br>if grouped underneath a DataSetWriter. <br>_Note:_ the DataSetPublishingInterval is overwritten if an<br> OpcPublishingInterval is set at the node level                                                                                                                                                                                                                                  |
```
hansgschossmann: ><br> [](http://example.com/codeflow?start=146&length=4)

Why is <br> used? Markdown formatter should to the right thing? Only if a line break makes sense it might be used. #Closed

docs/modules/publisher-migrationpath.md
```diff
@@ -17,63 +17,43 @@
...
+| **UnpublishAllNodes**            | EndpointUrl                                                     | Status, List\<StatusResponse\>                                                                                                                                                                 | Yes          |
+| **GetConfiguredEndpoints**       | -                                                               | List\<EndpointUrl\>                                                                                                                                                                            | Yes          |
+| **GetConfiguredNodesOnEndpoint** | EndpointUrl                                                     | EndpointUrl, List< OpcNodeOnEndpointModel > where OpcNodeOnEndpointModel contains: Id ExpandedNodeId OpcSamplingInterval OpcPublishingInterval DisplayName HeartbeatInterval SkipFirst | Yes          |
```
hansgschossmann: >SkipFirst [](http://example.com/codeflow?start=276&length=9)

Is this still supported? #Closed

docs/modules/publisher.md
```diff
@@ -125,29 +117,35 @@ A basic configuration file looks like this:
...
+
+OPC UA optimizes network bandwidth by only sending changes to OPC Publisher when the data item's value has changed. Some use cases require to publish data values in constant intervals. OPC Publisher supports a "heartbeat" for every configured telemetry event that can be enabled by specifying the `HeartbeatInterval` key in the data item's configuration. The interval is specified in seconds:
 
 ``` json
  "HeartbeatInterval": 3600,
 ```
 
-An OPC UA always send the current value of a data item when OPC Publisher first connects to it. To prevent publishing this data to IoT Hub, the SkipFirst key can be additionally specified in the data item's configuration:
+OPC UA sends the current data value when OPC Publisher connects to the OPC UA server. To prevent publishing this telemetry on startup to IoT Hub, the `SkipFirst` key can be additionally specified in the data item's configuration:
```
hansgschossmann, line 134: >SkipFirst [](http://example.com/codeflow?start=151&length=9)

Is this still supported? I have not seen any code. #Closed

docs/modules/publisher-directmethods.md
```diff
@@ -65,46 +61,48 @@ The `_V1` direct methods  uses the  payload schema as described below:
...
+| `DataSetPublishingIntervalTimespan` | no        | String | null                      | The publishing interval used for a grouped set of nodes <br> expressed as a Timespan string ({d.hh:mm:dd.fff}).<br>When both Intervals are specified, the <br> Timespan will win and be used for the configuration. <br>_Note:_ the DataSetPublishingInterval is overwritten if an<br> OpcPublishingInterval is set at the node level                                                                                                                                                                                                                                  |
```
hansgschossmann: >grouped set of nodes [](http://example.com/codeflow?start=124&length=20)

This is hard to understand for the user. Is not the DataSet itself a group of nodes identified by the same publishing interval? #Closed

docs/modules/publisher-directmethods.md
```diff
@@ -65,46 +61,48 @@ The `_V1` direct methods  uses the  payload schema as described below:
...
+| `OpcPublishingInterval`         | No        | Integer | 1000    | The publishing interval for the monitored item to be <br>published. Value expressed in milliseconds. <br>This value will be overwritten when a publishing interval <br>is explicitly defined in the DataSetWriter for the specified node. The value is used as defined in the OPC UA specification. Ignored when OpcPublishingIntervalTimespan is present.                                                                            |
+| `OpcPublishingIntervalTimespan` | No        | String  | null    | The publishing interval for the monitored item to be <br>published. Value expressed in Timespan <br>string({d.hh:mm:dd.fff}). <br>This value will be overwritten when a publishing interval <br>is explicitly defined in the DataSetWriter for the specified node.<br>The value is used as defined in the OPC UA specification. |
```
hansgschossmann: >This value will be overwritten when a publishing interval <br>is explicitly defined in the DataSetWriter for the specified node. [](http://example.com/codeflow?start=198&length=128)

This is confusing. The DataSetPublishing description says the opposite if i understand it correctly. From looking at the code, the explanation here is not correct but the one at the DataSetPublishingIntervalTimespan is correct and it makes also senses from my PoV. #Closed

docs/modules/publisher-directmethods.md
```diff
@@ -65,46 +61,48 @@ The `_V1` direct methods  uses the  payload schema as described below:
...
+| `OpcSamplingIntervalTimespan`   | No        | String  | null    | The sampling interval for the monitored item to be <br>published. Value expressed in Timespan <br>string({d.hh:mm:dd.fff}). <br>The value is used as defined in the OPC UA specification.                                                                                                                                     |
+| `OpcPublishingInterval`         | No        | Integer | 1000    | The publishing interval for the monitored item to be <br>published. Value expressed in milliseconds. <br>This value will be overwritten when a publishing interval <br>is explicitly defined in the DataSetWriter for the specified node. The value is used as defined in the OPC UA specification. Ignored when OpcPublishingIntervalTimespan is present.                                                                            |
```
hansgschossmann: >This value will be overwritten when a publishing interval <br>is explicitly defined in the DataSetWriter for the specified node. [](http://example.com/codeflow?start=173&length=128)

same as comment on next line. #Closed

docs/modules/publisher-directmethods.md
```diff
@@ -65,46 +61,48 @@ The `_V1` direct methods  uses the  payload schema as described below:
...
+| `UseSecurity`                       | no        | Boolean | false                     | Controls whether to use a secure OPC UA mode to establish a session to the OPC UA server endpoint                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      |
```
hansgschossmann: >false [](http://example.com/codeflow?start=62&length=5)

Has this changed over the versions? Formerly we where secure by design and we used a secure transport by default. Now we do not anymore? #Closed

hansgschossmann: >Controls whether to use a secure OPC UA mode to establish a session to the OPC UA server endpoint [](http://example.com/codeflow?start=90&length=97)

Not really related: I saw that the code does use Security.Best and Security.None. Is UseSecurity = false forcing Security.None and is Security.None enforcing SecurityProfile "None" or is it enforcing the lowest common SecurityProfile between client and server? #Closed

docs/modules/publisher.md
```diff
@@ -84,29 +75,30 @@ OPC Publisher has several interfaces that can be used to configure it.
...
+- encrypted communication between the systems
+- optionally for user authentication
 
-Security is enabled in the configuration file via the `"UseSecurity": true,` flag. The OPC Publisher will connect to the most secure endpoint available on the OPC UA servers.
-By default, OPC Publisher uses anonymous user authentication (in additional to the application authentication described above). However, OPC Publisher also supports user authentication using username and password. This can be specified via the REST API configuration interface (described below) or the configuration file as follows:
+OPC Publisher can be configured to store these certificates in a file system based certificate store. During startup, OPC Publisher checks if there's already a private certificate it can use. If not, a self-signed certificate is created. Self-signed certificates don't provide any trust value and we don't recommend using them in production.
+
+Encrypted communication can be enabled per endpoint via the `"UseSecurity": true,` flag. By default OPC Publisher will connect to an endpoint using the least secure mode OPC Publisher and the OPC UA server support.
```
hansgschossmann, line 88: >least secure mode OPC Publisher and the OPC UA server support [](http://example.com/codeflow?start=152&length=61)

Is this correct or is it really enforcing SecurityPolicy None and if the server does not support it, then the session will fail? #Closed

docs/modules/publisher.md
```diff
@@ -262,26 +269,26 @@ The `mq/om` parameter controls the upper limit of the capacity of the internal m
...
 
-To measure the performance of OPC Publisher, the `di` parameter can be used to print performance metrics to the trace log in the interval specified (in seconds).
+To measure the performance of OPC Publisher, the `di` parameter can be used to print performance metrics to the log in the interval specified (in seconds).
 
 ## Local metrics dashboard for OPC Publisher
 
-To learn more about how to create a local metrics dashboard for OPC Publisher V2.7, please refer to the tutorial [here](../tutorials/tut-publisher-local-metrics-dashboard/MetricsDashboard.md).
+To learn more about how to create a local metrics dashboard for OPC Publisher V2.7, refer to the tutorial [here](../tutorials/tut-publisher-local-metrics-dashboard/MetricsDashboard.md).
 
 ## Next steps
```
hansgschossmann, line 302: I think for advance users it would be great to understand how OPC Publisher is managing subscriptions internally: which nodes are put on which subscriptions and when is created a new one, .... #Closed

docs/modules/publisher-directmethods.md
```diff
@@ -65,46 +61,48 @@ The `_V1` direct methods  uses the  payload schema as described below:
...
+| `SkipFirst`                     | No        | Boolean | false   | Instructs the publisher not to add to telemetry the<br> Initial DataChange (after subscription activation) for this OpcNode.                                                                                                                                                                                      |
+| `QueueSize`                     | No        | Integer | 1       | The desired QueueSize for the monitored item to be published.                                                                                                                                                                                                                                                     |
 
 _Note_: __Id__ field may be omitted when ExpandedNodeIdId is present.
```
hansgschossmann, line 105: This is confusing for me: Id is Mandatory, how can it then be ommited? The explanation must be improved and the "Mandatory" flag aligned with the explanation. #Closed

hansgschossmann:
> **Please note: This feature is not available in version 2.5 or below.**

Must this be 2.5.x or should we do "in OPC Publisher versions < 2.6"?

---
Refers to: docs/modules/publisher.md:164 in 6f84244. [](commit_id = 6f842448c2a6dd12ffc04bcb7bc54de0e8311262, deletion_comment = False)

hansgschossmann:
>                                       Alternative: --bs, --batchsize

I do not understand this one. Maybe there should be somewhere in publisher.md a paragraph which talks about batching and how to configure it.
The first sentence does not make any sense to me.

---
Refers to: docs/modules/publisher-commandline.md:152 in 6f84244. [](commit_id = 6f842448c2a6dd12ffc04bcb7bc54de0e8311262, deletion_comment = False)

hansgschossmann:
>                                       Environment variable type: time span string {[d.]hh:mm:ss[.fffffff]}

This does not follow the general practice everywhere else to call the timespan version of the intervall "XxxxxxIntevalTimespan".

---
Refers to: docs/modules/publisher-commandline.md:162 in 6f84244. [](commit_id = 6f842448c2a6dd12ffc04bcb7bc54de0e8311262, deletion_comment = False)

hansgschossmann:
> 

This makes no sense to me. Is this an internal option used for testing? Either we should explain it or delete it.

---
Refers to: docs/modules/publisher-commandline.md:189 in 6f84244. [](commit_id = 6f842448c2a6dd12ffc04bcb7bc54de0e8311262, deletion_comment = False)

hansgschossmann:
>                                       Alternative: --fm, --fullfeaturedmessage

why is "legacy compatibility" mentioned here? this option is about "FullFeaturedMessage".

---
Refers to: docs/modules/publisher-commandline.md:144 in 6f84244. [](commit_id = 6f842448c2a6dd12ffc04bcb7bc54de0e8311262, deletion_comment = False)

hansgschossmann:
> 

The only occurence i have found the word "legacy" in this folder was in publishednodes.json.
I think it is important to explain what is meant here and controlled with this option.

---
Refers to: docs/modules/publisher-commandline.md:198 in 6f84244. [](commit_id = 6f842448c2a6dd12ffc04bcb7bc54de0e8311262, deletion_comment = False)

docs/modules/publisher-commandline.md
```diff
@@ -202,7 +206,7 @@ The following list of module parametrization settings can be provided either as
             MaxNodesPerDataSet=VALUE
                                       Maximum number of nodes within a DataSet/Subscription. 
                                       When more nodes than this value are configured for a 
-                                      dataSetWriter, they will be added in a separate DataSet/Subscription.
+                                      DataSetWriter, they will be added in a separate DataSet/Subscription.
```
hansgschossmann, line 201: >DataSet/Subscription [](http://example.com/codeflow?start=86&length=20)

Is DataSet == Subscription in the OPC UA specification? #Closed

docs/modules/publisher-commandline.md
karok2m: Doesn't seem so, I also cannot find any logic around it in the code. In `2.8.1` and `2.8.0` there was a todo item in the code. Seems like we have added the propagation of the value through job converter, but there is no actual logic to do the skipping.

docs/modules/publisher-directmethods.md
karok2m: I am assuming that you are referring to `<br>` as I cannot see the character in the comment. It helps a lot with formatting in tables and let's us avoid super long lines there. My markdown linter is also complaining about it locally, but GitHub renders it properly so we are making use of it.

docs/modules/publisher-commandline.md
hansgschossmann: How should we handle this? Remove from the doc and remove from code?

docs/modules/publisher-directmethods.md
hansgschossmann: ups. my  bad. i meant "SkipFirst". Should we remove "SkipFirst" here and in code?

docs/modules/publisher-directmethods.md
hansgschossmann: i understand, but it will unconditionally insert line breaks and will lead to not nicely formatted results when you have different with of screens.
I use Markdown All in One extension, which take care of the table formatting. I think it makes some sense when you want to start a new paragraph, like when showing Options or Default.
But at the end it is sort of style guide thing. On our spec i typically do not use it at all.

docs/modules/publisher-directmethods.md
```diff
@@ -2,34 +2,30 @@
...
-- __DataSetWriter__ has one DataSet and contains the elements required to successfully establish a connection to the OPC UA Server.
-- __DatSetWriterGroup__ is used to group several DataSetWriters for a specific OPC UA server.
+- __DataSet__ is a group of nodes within one OPC UA Server publishing of data value changes of those nodes is done with the same publishing interval.
+- __DataSetWriter__ has one DataSet and contains all information to establish a connection to an OPC UA Server.
+- __DatSetWriterGroup__ is used to group several DataSetWriter's for a specific OPC UA server.
 
 ## Payload Schema
 
-The `_V1` direct methods  uses the  payload schema as described below:
+The `_V1` direct methods use the payload schema as described below:
```
vslepakov, line 31: Looking at the schema here I am trying to better understand the usage of:

```
"DataSetWriterGroup": "string",
"DataSetWriterId": "string",
"DataSetPublishingInterval": "integer",
```
My current understanding is that there is one DataSetWriterGroup per server Endpoint and I can specify a friendly name for the session (overriding the default hash append behavior).
It looks like I can only specify only ONE DataSetWriterId per Group with ONE DataSetPublishingInterval (as group key?)

Would it be possible to specify the same endpoint multiple times with e.g., the same DataSetWriterGroup but different DataSetWriterId and DataSetPublishingInterval?
e.g.:

```
{
   "EndpointUrl": "opc.tcp://opcplc:50000",
   "DataSetWriterGroup": "Asset0",
   "DataSetWriterId": "DataFlow0",
   "DataSetPublishingInterval": 5000,
   "OpcNodes": [
      {
         "Id": "nsu=http://microsoft.com/Opc/OpcPlc/;s=FastUInt0"
      }
   ],
   "EndpointUrl": "opc.tcp://opcplc:50000",
   "DataSetWriterGroup": "Asset0",
   "DataSetWriterId": "DataFlow1",
   "DataSetPublishingInterval": 1000,
   "OpcNodes": [
      {
         "Id": "nsu=http://microsoft.com/Opc/OpcPlc/;s=FastUInt1"
      }
   ]
}
```

Or even different DataSetWriterGroup for the same endpoint?

If this is possible, what are the use cases/drivers for doing this? 
Are there any implications regarding the throughput or session count one would need to consider?

 #Closed

docs/modules/publisher-commandline.md
karok2m: We will add back the todo in the code and remove the parameter from the docs for now.

docs/modules/publisher-migrationpath.md
```diff
@@ -179,17 +161,17 @@ When there are configured endpoints:
...
 
-### GetConfiguredNodesOnEndpoint_V1
+### GetConfiguredNodesOnEndpoint (GetConfiguredNodesOnEndpoint_V1)
 
-`Request`:
+`Request`
 
 ```json
 {
    "EndpointUrl": "opc.tcp://sandboxhost-637811493394507132:50000"
```
MaidaB, line 168: Please add an example with all parameters and fix the corresponding explanation "..the other attributes are used to refine the result..." #Closed

docs/modules/publisher-commandline.md
cristipogacean: No, in the OPC UA Spec DataSet and Subscription are totally different concepts. We combine them though in publisher.

docs/modules/publisher-directmethods.md
cristipogacean: @vslepakov, this is a very good feedback! We need to explain this better.

The schema has certain limitations because we did not want to break the backwards compatibility for the pn.json. 

So basically your example is valid (grouping is a bit broken though) and the result is that the publisher will create 1 session, called `Asset0`. Within this session 2 subscriptions are created for `DataFlow0` and `DataFlow1` having 5000ms respectively 1000ms as publishing intervals.

If 2 DataSetWriterGroup are provided for the same endpoint, like in the example below:
```Json
[
	{
		"EndpointUrl": "opc.tcp://opcplc:50000",
		"DataSetWriterGroup": "Asset0",
		"DataSetWriterId": "DataFlow0",
		"DataSetPublishingInterval": 5000,
		"OpcNodes": [
			{
				"Id": "nsu=http://microsoft.com/Opc/OpcPlc/;s=FastUInt0"
			}
		]
	},
	{
		"EndpointUrl": "opc.tcp://opcplc:50000",
		"DataSetWriterGroup": "Asset1",
		"DataSetWriterId": "DataFlow1",
		"DataSetPublishingInterval": 1000,
		"OpcNodes": [
			{
				"Id": "nsu=http://microsoft.com/Opc/OpcPlc/;s=FastUInt1"
			}
		]
	}
]
```

the result would be that 2 sessions are created, using the very same endpoint & security options, `Asset0` and `Asset1`. Session `Asset0` has one subscription `DataFlow0` publishing 5000ms, respectively session `Asset1` has 1 subscription, `DataFlow1` publishing with 1000ms. 

One of the reasons for introducing these concepts were to give more control to the user on how the sessions/subscriptions/monitored items are created in order to better fit the various server setups we might find in the field. 
Another reason for this change is to introduce PubSub specific aliasing (`DataSetWriterGroup` and `DataSetWriter`) to allow decupling of protocol connectivity settings from logical semantics e.g. if the ip address of an opc ua server changes, but semantically represents the same asset, we can reconfigure the `endpoint url` without reconfiguring the `DataSetWriterGroup` therefore the telemetry format will not change for a seamless vertical consumption. This way redundancy/high availability scenarios can be easier architected. 

Regarding the last question, it is very very difficult to provide generic guidelines when it comes to load balancing sessions/subscriptions/monitored items for an optimal performance between OPC UA Client/server, because of the various profiles OPC UA was designed for. For example, a nano-embedded profile OPC UA Server may only allow 1 session. This is exactly the reason we (optionally) give more control to the user to manage and tune his configuration for best possible outcome. Please note that the new configuration settings are optional, therefore the backwards compatibility is ensured both in terms of configuration as well as behavior.

docs/modules/publisher-directmethods.md
cristipogacean: I don't think this changed lately, at least not from 2.7.x to 2.8.2. I believe the reason for this was that we wanted to be interoperable & easy to use on the first attempt first and then enable security. We can easily change this is required.

cristipogacean: I believe UseSecurity=False means "lowest" enforced by the server.

hansgschossmann:
>                                       Alternative: --fm, --fullfeaturedmessage

Do we explain what this means somewhere in the documentation?

---
In reply to: [1054192611](https://github.com/Azure/Industrial-IoT/pull/1607#issuecomment-1054192611) [](http://example.com/codeflow?ancestors=1054192611,1054192611)

---
Refers to: docs/modules/publisher-commandline.md:144 in 6f84244. [](commit_id = 6f842448c2a6dd12ffc04bcb7bc54de0e8311262, deletion_comment = False)

docs/modules/publisher-commandline.md
karok2m: Command line arguments related to `SkipFirst` are commented out in #1612.

docs/modules/publisher-directmethods.md
karok2m: Command line arguments and direct method parameters related to `SkipFirst` are commented out in #1612.

docs/modules/publisher-migrationpath.md
karok2m: This part talks about direct methods of `2.5` version of OPC Publisher. `SkipFirst` is supported there.

docs/modules/publisher.md
karok2m: Command line arguments and direct method parameters related to `SkipFirst` are commented out in #1612.

docs/modules/publisher-directmethods.md
karok2m: Changed the wording. @hansgschossmann, please have another look at it.

docs/modules/publisher-directmethods.md
karok2m: I've changed the wording in both places to indicate that `OpcPublishingInterval` will overwrite value of `DataSetPublishingInterval` for the specified node. Please have a look.

docs/modules/publisher-directmethods.md
karok2m: Changed wording to indicate that `OpcPublishingInterval` will overwrite value of `DataSetPublishingInterval` for the specified node. Please have a look.

docs/modules/publisher-commandline.md
hansgschossmann: removed

docs/modules/publisher-directmethods.md
hansgschossmann: removed

docs/modules/publisher.md
hansgschossmann: removed

docs/modules/publisher-directmethods.md
karok2m: Moved the note to the description parts of `Id` and `ExpandedNodeId` to be more explicit.

docs/modules/publisher-migrationpath.md
karok2m: Fixed description of request for `GetConfiguredNodesOnEndpoint_V1`. Added full example.

karok2m:
> > ```
> >   "QueueSize": "integer"
> > ```
> 
> Is this still supported? I have not seen any code.
> 
> In reply to: [1054159601](https://github.com/Azure/Industrial-IoT/pull/1607#issuecomment-1054159601)
> 
> In reply to: [1054159601](https://github.com/Azure/Industrial-IoT/pull/1607#issuecomment-1054159601)
> 
> Refers to: docs/modules/publisher-directmethods.md:56 in [6f84244](https://github.com/Azure/Industrial-IoT/commit/6f842448c2a6dd12ffc04bcb7bc54de0e8311262). [](commit_id = [6f84244](https://github.com/Azure/Industrial-IoT/commit/6f842448c2a6dd12ffc04bcb7bc54de0e8311262), deletion_comment = False)

@hansgschossmann, yes `QueueSize` is fully supported and tested.

docs/modules/publisher-directmethods.md
karok2m: I've added description of how DataSetWriters and DatSetWriterGroups are going to define subscriptions and sessions to OPC UA servers.

docs/modules/publisher.md
karok2m: I've added some explanation in documentation of direct methods about how DataSetWriters and DatSetWriterGroups are defining subscriptions and sessions to OPC UA servers.

docs/modules/publisher-directmethods.md
hansgschossmann: @cristipogacean  can this cause backward compatibility issues? This looks like it defaulted to true earlier: [https://github.com/Azure/iot-edge-opc-publisher-v1/blob/main/opcpublisher/Models/ConfigurationFileEntryModel.cs](https://github.com/Azure/iot-edge-opc-publisher-v1/blob/main/opcpublisher/Models/ConfigurationFileEntryModel.cs)

docs/modules/publisher-commandline.md
hansgschossmann: @cristipogacean  i think somewhere else we should describe how these both are combined by OPC Publisher. Maybe publisher.md is the place to do this.

hansgschossmann:
> **Please note: This feature is not available in version 2.5 or below.**

i have changed it to 2.5.x

---
In reply to: [1054183111](https://github.com/Azure/Industrial-IoT/pull/1607#issuecomment-1054183111) [](http://example.com/codeflow?ancestors=1054183111,1054183111)

---
Refers to: docs/modules/publisher.md:164 in 6f84244. [](commit_id = 6f842448c2a6dd12ffc04bcb7bc54de0e8311262, deletion_comment = False)

karok2m:
> > ```
> >                                   Alternative: --bs, --batchsize
> > ```
> 
> I do not understand this one. Maybe there should be somewhere in publisher.md a paragraph which talks about batching and how to configure it. The first sentence does not make any sense to me.
> 
> Refers to: docs/modules/publisher-commandline.md:152 in [6f84244](https://github.com/Azure/Industrial-IoT/commit/6f842448c2a6dd12ffc04bcb7bc54de0e8311262). [](commit_id = [6f84244](https://github.com/Azure/Industrial-IoT/commit/6f842448c2a6dd12ffc04bcb7bc54de0e8311262), deletion_comment = False)

@hansgschossmann, I've added notes in `publisher.md` about batching. Change the wording in description of command line options to say incoming instead of ingress. Not the description reads as `The number of incoming OPC UA data change messages to be cached for batching.`. Is this clearer ?

hansgschossmann:
>                                       Alternative: --bs, --batchsize

@cristipogacean maybe the first sentence should be something like: "The number of received OPC UA data-value change notifications from the OPC UA server to be cached before sending them.

---
In reply to: [1054186023](https://github.com/Azure/Industrial-IoT/pull/1607#issuecomment-1054186023) [](http://example.com/codeflow?ancestors=1054186023,1054186023)

---
Refers to: docs/modules/publisher-commandline.md:152 in 6f84244. [](commit_id = 6f842448c2a6dd12ffc04bcb7bc54de0e8311262, deletion_comment = False)

hansgschossmann:
> 

i have commented ScaleTestCount out.
cc: @cristipogacean , @karok2m 

---
In reply to: [1054187555](https://github.com/Azure/Industrial-IoT/pull/1607#issuecomment-1054187555) [](http://example.com/codeflow?ancestors=1054187555,1054187555)

---
Refers to: docs/modules/publisher-commandline.md:189 in 6f84244. [](commit_id = 6f842448c2a6dd12ffc04bcb7bc54de0e8311262, deletion_comment = False)

hansgschossmann:
yep, this sounds much better. i had also something like this in mind: "The number of received OPC UA data-value change notifications from the OPC UA server to be cached before sending them."

---
In reply to: [1061857357](https://github.com/Azure/Industrial-IoT/pull/1607#issuecomment-1061857357) [](http://example.com/codeflow?ancestors=1061857357,1061857357)

hansgschossmann:
i know, this was misplaced :( should be placed on SkipFirst.

---
In reply to: [1061627838](https://github.com/Azure/Industrial-IoT/pull/1607#issuecomment-1061627838) [](http://example.com/codeflow?ancestors=1061627838,1061627838)

docs/modules/publisher-commandline.md
```diff
@@ -552,11 +557,13 @@ The following list of module parametrization settings can be provided either as
                                    nodes without
                                    a heartbeat interval setting.
                                    Default: 0
+<!---
```
karok2m: @hansgschossmann, I believe that `2.5.x` version of OPC Publisher supported `skipfirstevent`. So we should uncomment this part back, in my opinion. #Closed

marcschier (dismissed):

docs/modules/publisher-directmethods.md
hansgschossmann: @cristipogacean i have updated this to reflect that even "UseSecurity" == false may result in a secure connection.

docs/modules/publisher.md
hansgschossmann: @cristipogacean i have updated this as well. pls check.

cristipogacean (dismissed):
:shipit:

karok2m (dismissed):

karok2m (approved):

cristipogacean (approved):

docs/modules/publisher.md
cristipogacean: looks good to me!

PR #1609, opened 2022-03-01
Fix the MethodCallStatusException payload

dacolgit:
fix bug [https://msazure.visualstudio.com/One/_workitems/edit/13250556](https://msazure.visualstudio.com/One/_workitems/edit/13250556)


common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/MethodRouter.cs
```diff
@@ -335,6 +342,22 @@ private class DefaultFilter : ExceptionFilterAttribute {
...
+            /// </summary>
+            /// <param name="ex"></param>
+            private MethodCallStatusExceptionModel CreateExceptionResponse(Exception ex) {
+                MethodCallStatusExceptionModel exceptionResponse = null;
+                if (!string.IsNullOrEmpty(ex.Message)) {
+                    exceptionResponse = new MethodCallStatusExceptionModel {
+                        Message = ex.Message.Split(":").First(),
+                        Description = ex.Message.Split(":", 2).Last(),
+                    };
+                }
```
karok2m: This logic relies heavily on the structure of the message that is generated in the constructor of `MethodCallStatusException`, which makes me think that this should happen immediately there instead of requiring a post-processing step here:

```
$"Response {result} {errorMessage ?? ""}: {responsePayload}"
```

I think that it would be better if in the constructor itself we create the payload structure that we are trying to add here. It would also be more maintainable i the long run since it would not require knowledge of the fact that the final message is constructed in two steps, in the constructor of `MethodCallStatusException` and here.

I am not sure, but I also think that this logic might not process the case when `MethodCallStatusException` is constructed without a response payload. In that case internally the constructor that accepts response payload is called with `{}` as payload. It seems like that might not be parsed correctly here. So we need to add tests that make sure that in both cases resulting `ResponsePayload` would be a valid JSON that can be deserialized back to an object from string. And on a side note, `null` might be a better value to pass from the first constructor to the one with response payload instead of `{}`.

dacolgit: I moved the creation of json paylosd in  `MethodCallStatusException`

common/src/Microsoft.Azure.IIoT.Abstractions/src/Exceptions/MethodCallStatusException.cs
```diff
@@ -39,7 +39,7 @@ public class MethodCallStatusException : MethodCallException {
         /// <param name="errorMessage"></param>
         public MethodCallStatusException(string responsePayload, int result,
             string errorMessage = null) :
-            base($"Response {result} {errorMessage ?? ""}: {responsePayload}") {
+            base($" {{ \"Message\":\"Response {result} {errorMessage ?? ""}\",\"Description\": {responsePayload} }}") {
```
karok2m: Please add tests that would validate that we can reserialize `ResponsePayload` as a proper JSON afterwards. The tests should also check the cases where passed `responsePayload` or `errorMessage` are null. In both cases we should still be able to deserialize. I think that the test should also make sure that message constructed from the other constructor above is also deserializable.

dacolgit: Added unit test

common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/MethodRouter.cs
```diff
@@ -300,8 +300,9 @@ private class DefaultFilter : ExceptionFilterAttribute {
                         }
                         _logger.Verbose(ex, "Method call error");
                         ex = _ef.Filter(ex, out var status);
+
                         if (ex is MethodCallStatusException) {
-                            throw ex;
+                            throw new MethodCallStatusException(ex.Message, status);
```
karok2m: Wouldn't `ex` already contain the `status` ? Looks like we extract `status` from `ex` itself a few lines above.

karok2m: Same comment for changes on line `331` below.

common/src/Microsoft.Azure.IIoT.Abstractions/src/Exceptions/MethodCallStatusException.cs
```diff
@@ -39,7 +39,7 @@ public class MethodCallStatusException : MethodCallException {
         /// <param name="errorMessage"></param>
         public MethodCallStatusException(string responsePayload, int result,
             string errorMessage = null) :
-            base($"Response {result} {errorMessage ?? ""}: {responsePayload}") {
+            base($"{{\"Message\":\"Response {result} {errorMessage ?? ""}\",\"Description\":{responsePayload}}}") {
```
karok2m: Please add a test for deserialization in the case when `responsePayload` is `null`.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Engine/StandaloneJobOrchestratorTests.cs
```diff
@@ -124,19 +124,96 @@ public class StandaloneJobOrchestratorTests : TempFileProviderBase {
...
+            };
+            serializeExceptionModel = _newtonSoftJsonSerializer.SerializeToString(exceptionModel);
+            serializeExceptionModel.Should().BeEquivalentTo(exceptionResponse);
+        }
+
+            [Fact]
+        public async Task Test_PublishNodes_NullOrEmpty() {
+            InitStandaloneJobOrchestrator();
+
+            var exceptionResponse = $"{{\"Message\":\"Response 400 null request is provided\",\"Description\":{{}}}}";
```
karok2m: `exceptionResponse` object does not seem to be used in `Test_PublishNodes_NullOrEmpty`. Can we remove it ?

common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/MethodRouter.cs
dacolgit: The exception response is not the same.
`throw new MethodCallStatusException(ex.Message, status);` generates:
`{
    "status":404,
    "payload":{
        "Message":"Response 404 Nodes not found",
        "Description":{
            "EndpointUrl":"opc.tcp://192.168.1.70:50000",
            "UseSecurity":false,
            "OpcAuthenticationMode":"Anonymous",
            "OpcNodes":[
                {
                    "Id":"ns=2;s=SlowUInt1",
                    "OpcPublishingInterval":1000
                }
            ]
        }
    }
}
`
while `throw ex` generates:
`{
    "status":404,
    "payload":{
        "EndpointUrl":"opc.tcp://192.168.1.70:50000",
        "UseSecurity":false,
        "OpcAuthenticationMode":"Anonymous",
        "OpcNodes":[
            {
                "Id":"ns=2;s=SlowUInt1",
                "OpcPublishingInterval":10000
            }
        ]
    }
}
`

karok2m: @dacolgit, is this the result of direct method invocation ?

common/src/Microsoft.Azure.IIoT.Abstractions/src/Exceptions/MethodCallStatusException.cs
dacolgit: added

common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/MethodRouter.cs
karok2m: I believe that this difference is caused by the fact that here you are transferring the value that has bees stored in `Message` to `ResponsePayload`, effectively overwriting it. And `ResponsePayload` is what is being picked up by `MethodRouter` when assembling the response to a direct method call. This makes me realize that we are incorrectly assigning assembled JSON to the `Message` field instead of `ResponsePayload` in the exception. So I think that we need to change the constructors that we have to contain the JSON message in `ResponsePayload`.

So this:

```csharp
        public MethodCallStatusException(int result, string errorMessage = null) :
            this("{}", result, errorMessage) {
            ResponsePayload = $"\"{errorMessage}\"";
        }

        ...
        public MethodCallStatusException(string responsePayload, int result,
            string errorMessage = null) :
            base($"{{\"Message\":\"Response {result} {errorMessage ?? ""}\",\"Description\":{responsePayload}}}") {
            Result = result;
            ResponsePayload = responsePayload;
        }
```

Should become something like this:

```csharp
        public MethodCallStatusException(int result, string errorMessage) :
            this("{}", result, errorMessage) { }

        ...
        public MethodCallStatusException(string responsePayload, int result, string errorMessage = null) :
            base($"Response {result} {errorMessage ?? ""}") {
            Result = result;
            ResponsePayload = $"{{\"Message\":\"Response {result} {errorMessage ?? ""}\",\"Description\":{responsePayload ?? "{}"}}}";
        }
```

common/src/Microsoft.Azure.IIoT.Abstractions/src/Exceptions/MethodCallStatusException.cs
```diff
@@ -39,7 +39,7 @@ public class MethodCallStatusException : MethodCallException {
...
             Result = result;
             ResponsePayload = responsePayload;
```
karok2m: Please note that `ResponsePayload` in the first constructor does not generate a proper JSON. We need to change that as `ResponsePayload` is picked up by `MethodRouter` to be the payload of direct method call response. So we need to fix that. Please see my previous comment about storing JSON payload in `Message` vs `ResponsePayload`, as one of the options would be to not overwrite `ResponsePayload` in the first constructor after it is being assigned here and make sure that we have a proper JSON here.

common/src/Microsoft.Azure.IIoT.Abstractions/src/Exceptions/MethodCallStatusException.cs
karok2m: [Microsoft REST API Guidelines](https://github.com/Microsoft/api-guidelines/blob/master/Guidelines.md#710-response-formats) recommend to use `Details` together with `Message`. So maybe we should rename `Description` to `Details` ?

A few more points from there:
* `Details` there is also a list of objects instead of a single object, but I think that we can keep our single object here.
* The top level object also contains `error` key with the whole object that we assemble here as its value. In our case the object will become the value for `payload` key. So I am happy with that part of our existing structure as well.

What do you think, @dacolgit, @cristipogacean ?

dacolgit: I'm ok for renaming `Description` to `Details`
@cristipogacean ?

common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/MethodRouter.cs
dacolgit: @dacolgit, is this the result of direct method invocation ?
Yes

common/src/Microsoft.Azure.IIoT.Abstractions/src/Exceptions/MethodCallStatusException.cs
dacolgit: yes changed

common/src/Microsoft.Azure.IIoT.Abstractions/src/Exceptions/MethodCallStatusException.cs
karok2m: Let's rename `Description` to `Details` then.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/tests/Utils/MethodCallStatusExeptionModel.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        /// <summary>
+        /// Exception message.
+        /// </summary>
+        [DataMember(Name = "Message", Order = 0,
+            EmitDefaultValue = true)]
+        public string Message { get; set; }
+
+        /// <summary>
+        /// Description.
```
karok2m: Let's change `Description` to something like `Details of the exception.`.

dacolgit: changed

karok2m (dismissed):

common/src/Microsoft.Azure.IIoT.Abstractions/src/Exceptions/MethodCallStatusException.cs
```diff
@@ -38,10 +37,10 @@ public class MethodCallStatusException : MethodCallException {
         /// <param name="result"></param>
         /// <param name="errorMessage"></param>
         public MethodCallStatusException(string responsePayload, int result,
-            string errorMessage = null) :
-            base($"Response {result} {errorMessage ?? ""}: {responsePayload}") {
+           string errorMessage = null) :
```
cristipogacean: please fix indentation

common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/MethodRouter.cs
```diff
@@ -300,6 +300,7 @@ private class DefaultFilter : ExceptionFilterAttribute {
                         }
                         _logger.Verbose(ex, "Method call error");
                         ex = _ef.Filter(ex, out var status);
+
```
cristipogacean: let's revert line changes here

common/src/Microsoft.Azure.IIoT.Hub.Module.Framework/src/Hosting/MethodRouter.cs
```diff
@@ -325,6 +326,7 @@ private class DefaultFilter : ExceptionFilterAttribute {
...
```
cristipogacean: let's revert line changes here

common/src/Microsoft.Azure.IIoT.Serializers.NewtonSoft/src/Default/NewtonSoftJsonSerializer.cs
```diff
@@ -534,6 +534,29 @@ internal sealed class JsonVariantConverter : JsonConverter {
             private readonly NewtonSoftJsonSerializer _serializer;
         }
 
+        /// <summary>
+        /// Json raw converter
+        /// </summary>
+        public sealed class RawJsonConverter : JsonConverter {
```
cristipogacean: since this is only used in the tests, can we move it in a Tests related class? next to MethodCallStatusExceptionModel maybe?

dacolgit: moved

common/src/Microsoft.Azure.IIoT.Abstractions/src/Exceptions/MethodCallStatusException.cs
dacolgit: fixed

karok2m (approved):

cristipogacean (approved):

PR #1613, opened 2022-03-08
Report invalid / unknown command line arguments

dacolgit:
fix [ICM 229489392: Report invalid / unknown command line arguments](https://msazure.visualstudio.com/One/_workitems/edit/9479357)


modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
```diff
@@ -161,39 +165,57 @@ public class StandaloneCliOptions : Dictionary<string, string>, IAgentConfigProv
...
+            try {
+                options.Parse(args);
+            }
+            catch (Exception e) {
+                logger.Warning("Parse args exception: " + e.Message);
+                Environment.Exit(0);
+            }
+
+            if (isLegacyOption) {
+                logger.Warning("Legacy option not supported, please use -h option to get all the supported options.");
```
karok2m, line 212: It would be great if we also log the option that caused this warning.

karok2m: It would be great if we also log the option that caused this warning.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
```diff
@@ -161,39 +165,57 @@ public class StandaloneCliOptions : Dictionary<string, string>, IAgentConfigProv
...
+                    { "i|install", "Legacy - do not use.", b => isLegacyOption = true },
+                    { "st|opcstacktracemask=", "Legacy - do not use.", b => isLegacyOption = true },
+                    { "sd|shopfloordomain=", "Legacy - do not use.", b => isLegacyOption = true },
+                    { "vc|verboseconsole=", "Legacy - do not use.", b => isLegacyOption = true },
+                    { "as|autotrustservercerts=", "Legacy - do not use.",b => isLegacyOption = true }
                 };
 
-            options.Parse(args);
+            try {
+                options.Parse(args);
```
karok2m, line 204: Would this report an error if unknown command line option (let's say `--unknown`) is provided ?

karok2m: Would this report an error if unknown command line option (let's say `--unknown`) is provided ?

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
```diff
@@ -161,39 +165,57 @@ public class StandaloneCliOptions : Dictionary<string, string>, IAgentConfigProv
...
+            }
+            catch (Exception e) {
+                logger.Warning("Parse args exception: " + e.Message);
+                Environment.Exit(0);
```
karok2m: We should have non-zero exit code in this case to indicate error. Maybe we can use `160` to follow guidelines [here](https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-?redirectedfrom=MSDN).

karok2m, line 208: We should have non-zero exit code in this case to indicate error. Maybe we can use `160` to follow guidelines [here](https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-?redirectedfrom=MSDN).

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
dacolgit: added

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
```diff
@@ -44,6 +44,11 @@ public class StandaloneCliOptions : Dictionary<string, string>, IAgentConfigProv
         /// <param name="args">The specified command line arguments.</param>
         public StandaloneCliOptions(string[] args) {
 
+            bool showHelp = false;
+            bool isLegacyOption = false;
+            List<string> legacyOptions = new List<string>();
+            var logger = ConsoleLogger.Create(LogEventLevel.Warning);
```
karok2m, line 52: Can we make sure that we have same log format as in the case of usual logs, such as having timestamp and information on source ?

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
karok2m: @dacolgit, documentation for [`Mono.Options.OptionSet.Parse()`](http://docs.go-mono.com/index.aspx?link=M%3AMono.Options.OptionSet.Parse(System.Collections.Generic.IEnumerable%7BSystem.String%7D)) method states that it returns a list of unhandled arguments. Can we please use this functionality to report warnings for all arguments that we do not recognize ?

karok2m: It would also be great if we had tests to cover this scenario.

cristipogacean: +1

dacolgit: added unit test

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
dacolgit: same output
image

Industrial-IoT.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29201.188
+# Visual Studio Version 17
```
cristipogacean: are we safe changing VS version in the solution now?

dacolgit: I restored the old version for safety

cristipogacean (dismissed):

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
```diff
@@ -165,39 +171,65 @@ public class StandaloneCliOptions : Dictionary<string, string>, IAgentConfigProv
...
+                    { "rc|removecert=", "Legacy - do not use.", b => {isLegacyOption = true; legacyOptions.Add("rc|removecert"); } },
+                    { "dt|devicecertstoretype=", "Legacy - do not use.", b => {isLegacyOption = true; legacyOptions.Add("dt|devicecertstoretype"); } },
+                    { "dp|devicecertstorepath=", "Legacy - do not use.", b => {isLegacyOption = true; legacyOptions.Add("dp|devicecertstorepath"); } },
+                    { "i|install", "Legacy - do not use.", b => {isLegacyOption = true; legacyOptions.Add("i|install"); } },
+                    { "st|opcstacktracemask=", "Legacy - do not use.", b => {isLegacyOption = true; legacyOptions.Add("st|opcstacktracemask"); } },
+                    { "sd|shopfloordomain=", "Legacy - do not use.", b => {isLegacyOption = true; legacyOptions.Add("sd|shopfloordomain"); } },
+                    { "vc|verboseconsole=", "Legacy - do not use.", b => {isLegacyOption = true; legacyOptions.Add("vc|verboseconsole"); } },
+                    { "as|autotrustservercerts=", "Legacy - do not use.", b => {isLegacyOption = true; legacyOptions.Add("as|autotrustservercerts"); } },
```
karok2m: Instead of having additional variable `isLegacyOption` and setting it to `true` for each deprecated option we can just check if `legacyOptions` list is empty or not.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/StandaloneCliOptionsTest.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    /// <summary>
+    /// Test class to override Exit method
+    /// </summary>
+    public class StandaloneCliOptionsTest : StandaloneCliOptions {
+
+        /// <summary>
+        /// Exit code
+        /// </summary>
+        public int ExitCode { get; set; }
```
karok2m: Shall we set it to `-1` by default here ? This way we can be sure in `HelpOptionParameterTest()` that the exit mechanism was used and we are not checking the default value of int that `ExitCode` would get anyways.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
dacolgit: done

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/StandaloneCliOptionsTest.cs
dacolgit: added init to -1

karok2m (approved):

Issue #1492, opened 2021-12-07
Unhandled error during Publish.  BadTimeout 'BadTimeout'


mwiot:
Hi, we are having some issues with the publisher. Getting bad timeout. Can you see from the logs what is causing this?
Had a stream running for the publisher for several hours, that suddenly stopped working. When checking the logs there were errors there regarding bad timeout. After the restart it is publishing data, but this error keeps coming on regular intervals. This is some timeout error related to OPC client (publisher) and OPC UA server?

```
[13:46:37 ERR Microsoft.Azure.IIoT.OpcUa.Protocol.Services.StackLogger] (Error) PUBLISH #15 - Unhandled error during Publish.
 BadTimeout 'BadTimeout'
```

Running 
Ubuntu 18.04
IIOT 2.8.1 all in one services
Publisher 2.8.1

```
[13:46:36 DBG Microsoft.Azure.IIoT.Agent.Framework.Agent.Worker] Sending heartbeat...
[13:46:36 VRB Microsoft.Azure.IIoT.Http.Default.HttpClient] Sending POST request to https://svc-bifrost-p2-weu.azurewebsites.net/edge/publisher/v2/heartbeat...
[13:46:36 VRB Microsoft.Azure.IIoT.Http.Default.HttpClient] POST to https://svc-bifrost-p2-weu.azurewebsites.net/edge/publisher/v2/heartbeat returned OK (took 00:00:00.0518594).
[13:46:36 DBG Microsoft.Azure.IIoT.Agent.Framework.Agent.Worker+JobProcess] Sending job processor heartbeat...
[13:46:36 VRB Microsoft.Azure.IIoT.Http.Default.HttpClient] Sending POST request to https://svc-bifrost-p2-weu.azurewebsites.net/edge/publisher/v2/heartbeat...
[13:46:36 DBG Microsoft.Azure.IIoT.OpcUa.Protocol.Services.DefaultSessionManager] Runner Keepalive reset due to checkAlive
[13:46:36 DBG Microsoft.Azure.IIoT.OpcUa.Protocol.Services.DefaultSessionManager] Refreshing session '2cc38d59fc2d08bc4bf6abe37aecde7d3ffe0640'
[13:46:36 VRB Microsoft.Azure.IIoT.OpcUa.Protocol.Services.StackLogger] (Service) Read Called. RequestHandle=394, PendingRequestCount=2
[...]

PR #1620, opened 2022-03-14
fix publish operation timeout exception during publishing

cristipogacean:
* fix for #1492
* ensure the configured opc stack transport arguments (OperationTimeout, LifeTimeCount, MaxKeepAliveCount) are passed through to the opc ua stack
* removed operation timeout from the connection model - it was never propagating to the DefaultSessionManager anyhow.


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -658,17 +691,19 @@ internal sealed class SubscriptionWrapper : ISubscription {
...
                         _logger.Error("Failed to add subscription '{name}' to session '{session}'",
                              Id, session.SessionName);
                         subscription = null;
                     }
                     else {
-                        subscription.Create();
-                        //TODO - add logs for the revised values
                         _logger.Debug("Added subscription '{name}' to session '{session}'",
                              Id, session.SessionName);
+                        subscription.Create();
```
karok2m, line 705: Should we have the log after we manages to create the subscription ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -538,64 +541,66 @@ internal sealed class SubscriptionWrapper : ISubscription {
...
-                        var itemsToChange = changeList.Select(t => t.Item).ToList();
-                        var results = rawSubscription.SetMonitoringMode(change.Key.Value, itemsToChange);
-                        if (results != null) {
-                            var erroneousResultsCount = results
-                                .Count(r => (r == null) ? false : StatusCode.IsNotGood(r.StatusCode));
 
-                            // Check the number of erroneous results and log.
-                            if (erroneousResultsCount > 0) {
-                                _logger.Warning("Failed to set monitoring for {count} items in subscription "
+                        if (activate) {
```
karok2m: Please move the logic under this `if` out of `else` case on line `543`.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
cristipogacean: fixing this in another PR

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
cristipogacean: fixed

karok2m (approved):

PR #1621, opened 2022-03-14
Added --overwrite true to az storage blob upload to fix upload errors.

karok2m:
We are getting "The specified blob already exists." errors even for blobs that do not exist:

```
ERROR: The specified blob already exists.
RequestId:c08a96fe-701e-0074-5f39-355d0d000000
Time:2022-03-11T11:15:14.5586036Z
ErrorCode:BlobAlreadyExists
If you want to overwrite the existing one, please add --overwrite in your command.
```

I've added `--overwrite true` to `az storage blob upload` command to fix upload errors. This is not yet working properly as macOS agents (both `10.15` and `11`) provided by Azure DevOps have not been updated yet to have `2.34.0+` version of Azure CLI which adds this flag. There are two PRs to fix this:

https://github.com/actions/virtual-environments/pull/5201
https://github.com/actions/virtual-environments/pull/5200

Once this is resolved on Azure DevOps side this PR should work.


cristipogacean (approved):

karok2m:
[Here](https://msazure.visualstudio.com/One/_build/results?buildId=53046619&view=results) is a sample successful run of the build pipeline.

PR #1624, opened 2022-03-14
add subscription diagnostic logs

cristipogacean:
* log revised values for settings of subscriptions and monitored items
* log warnings when sequence numbers are missed


monikavar (dismissed):

marcschier (dismissed):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -770,19 +817,26 @@ internal sealed class SubscriptionWrapper : ISubscription {
...
                         GetValueOrDefault(DateTime.UtcNow);
 
-                    _logger.Debug("DataChange for subscription '{subscription}'/'{sessionId}', sequence#: "
-                        + "{Sequence} isKeepAlive: {KeepAlive}, publishTime: {PublishTime}",
-                        Id,
-                        Connection.CreateConnectionId(),
-                        sequenceNumber,
-                        isKeepAlive,
-                        publishTime);
+                    if (_currentSequenceNumber == sequenceNumber) {
```
karok2m: Seems like `_expectedSequenceNumber` would be a better name than `_currentSequenceNumber` as it is not the one that we have received before but the one that we expect now. What  do you think ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -770,19 +817,26 @@ internal sealed class SubscriptionWrapper : ISubscription {
...
+                        _logger.Debug("DataChange for subscription '{subscription}'/'{sessionId}', sequence#: "
+                            + "{Sequence} isKeepAlive: {KeepAlive}, publishTime: {PublishTime}",
```
karok2m: Maybe we can have a `,` after `{Sequence}` ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -660,84 +681,110 @@ internal sealed class SubscriptionWrapper : ISubscription {
...
                         // do not deactivate an already activated subscription
                         subscription.SetPublishingMode(activate);
+                        logRevisedValues = true;
                     }
                 }
+
+                if (logRevisedValues && activate) {
+                    var diagInfo = new StringBuilder();
+                    diagInfo.Append("Subscription '{subscription}'/'{sessionId}' state actual(revised)/desired: ");
+                    diagInfo.Append("PublishingEnabled {currentPublishingEnabled}/{publishingEnabled} ");
```
karok2m: Should we add commas after reporting expected and actual values for each field ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
cristipogacean: agreed!

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
cristipogacean: indeed

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
cristipogacean: sure

karok2m (approved):

PR #1626, opened 2022-03-15
fix duplicate display names issue

cristipogacean:
* fix the name conflict when using same duplicate names or data-set field id's in the pn.json file


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Extensions/PublishedDataItemsModelEx.cs
```diff
@@ -31,33 +31,10 @@ public static class PublishedDataItemsModelEx {
...
-                monitoredItem.MonitoringMode = MonitoringMode.Sampling;
-                map.Add(monitoredItem.TriggerId,
-                    new MonitoredItemModel {
-                        MonitoringMode = MonitoringMode.Sampling,
-                        StartNodeId = "i=2258",
-                        SamplingInterval = monitoredItem.HeartbeatInterval,
-                        Id = monitoredItem.TriggerId,
-                    });
-                */
+                map.Add(monitoredItem.StartNodeId, monitoredItem);
```
karok2m, line 35: Should we use a LINQ statement here directly on `dataItems.PublishedData` to avoid dictionary creation and work with lists instead ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -493,8 +493,7 @@ internal sealed class SubscriptionWrapper : ISubscription {
                     rawSubscription.ApplyChanges();
                     _currentlyMonitored = nowMonitored;
 
-                    var map = _currentlyMonitored.ToDictionary(
-                        k => k.Template.Id ?? k.Template.StartNodeId, v => v);
+                    var map = _currentlyMonitored.ToDictionary(k => k.Template.StartNodeId, v => v);
```
karok2m, line 496: Would the usage of this dictionary on line `499` be unaffected by this change ? That one uses `item.Template.TriggerId` as search key.

dacolgit (approved):

karok2m (approved):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
cristipogacean: Yes, it's unaffected because triggering functionality is not used at all, it's not part of the API and was never tested. Actually I believe before the change was totally broken.

PR #1627, opened 2022-03-15
Added reporting of OPC Publisher restarts.

karok2m:
* Added reporting of runtime state.
* Added configuration option for enabling runtime state reporting.
* Moved state reporting logic to `Microsoft.Azure.IIoT.OpcUa.Edge.Publisher` project.
* Added tests for `RuntimeStateReporter` class.
* Made CMD configuration option a flag that does not require `true`/`false` value.
* Added documentation of CMD configuration option.

### DIFF
--- a/modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/ModuleProcess.cs
+++ b/modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/ModuleProcess.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Azure.IIoT.Modules.OpcUa.Publisher {
     using Microsoft.Azure.IIoT.Modules.OpcUa.Publisher.Agent;
     using Microsoft.Azure.IIoT.Modules.OpcUa.Publisher.Runtime;
     using Microsoft.Azure.IIoT.Modules.OpcUa.Publisher.Controller;
+    using Microsoft.Azure.IIoT.Modules.OpcUa.Publisher.State;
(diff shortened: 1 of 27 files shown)


modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/ModuleProcess.cs
```diff
@@ -120,6 +121,16 @@ public class ModuleProcess : IProcessControl {
                         kPublisherModuleStart.WithLabels(
                             identity.DeviceId ?? "", identity.ModuleId ?? "").Inc();
                         await workerSupervisor.StartAsync().ConfigureAwait(false);
+
+                        // Reporting runtime state on restart.
+                        // Reporting will happen only in stadalone mode.
+                        var standaloneCliOptions = new StandaloneCliOptions(_config);
```
cristipogacean: >var standaloneCliOptions = new StandaloneCliOptions(_config); [](http://example.com/codeflow?start=24&length=61)

I think we already have an instance of this. Can't we reuse it somehow?

karok2m: Changed the code to persist `RunInStandaloneMode` so that I do not need to recreate `StandaloneCliOptions`. Previously it was created in `ConfigureContainer()` method and later registered with DI but only in the case when we were running in standalone mode. And I actually just need to know if we are running in standalone mode or not, so doing that now.

cristipogacean (approved):
:shipit:

PR #1634, opened 2022-03-18
fix cli doc

dacolgit:
removed alternative option not supported anymore

### DIFF
--- a/docs/modules/publisher-commandline.md
+++ b/docs/modules/publisher-commandline.md
@@ -311,7 +311,7 @@ The `Alternative` field, where present, refers to the CLI argument applicable in
             TrustedIssuerCertificatesType=VALUE
                                       OPC UA Client Security Config - trusted issuer certificates
                                       store type
-                                      Alternative: --it, --issuercertstoretype
+                                      Alternative: N/A
                                       Mode: Standalone and Orchestrated
                                       Type: enum string : Directory, X509Store
                                       Default: Directory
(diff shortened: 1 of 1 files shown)


docs/modules/publisher-commandline.md
```diff
@@ -311,7 +311,7 @@ The `Alternative` field, where present, refers to the CLI argument applicable in
             TrustedIssuerCertificatesType=VALUE
                                       OPC UA Client Security Config - trusted issuer certificates
                                       store type
-                                      Alternative: --it, --issuercertstoretype
```
karok2m, line 314: @dacolgit, all of the options (`issuercertstoretype`, `trustedcertstoretype` and `rejectedcertstoretype`) are present in `StandaloneCliOptions` class. So why are they not supported any more ?

dacolgit: they are present as legacy options:
`{ "it|issuercertstoretype=", "Legacy - do not use.", b => {legacyOptions.Add("it|issuercertstoretype"); } },`
The alternative options --it, --issuercertstoretype are not supported. just the env variable is supported (TrustedIssuerCertificatesType). That for all the 3 options

karok2m: @cristipogacean, do you remember if there was a reason to removing those options from CLI ? I am wondering if we should bring them back instead of removing the documentation.

cristipogacean: yes, the reason was primarily useability: the idea was to reduce the huge number of options and deprecate the ones that are not exactly relevant ("less is more" in this case). I think we should not bring those back. There is a more generic alternative (that works for both standalone and orchestrated modes) available anyhow `TrustedIssuerCertificatesType`. we should slowly educate people to move to this format anyhow. besides this, I doubt anyone really uses a different certstore anyhow.

karok2m: OK, then this looks good.

karok2m (approved):

PR #1638, opened 2022-03-18
update telemetry messages doc

cristipogacean:
- update the documentation with the newly generated telemetry for different settings
- removed binary format telemetry from doc


docs/dev-guides/telemetry-messages-format.md
```diff
@@ -8,141 +8,91 @@ Telemetry messages used by the IIoT Platform
...
     "ExtensionFields": {
-      "EndpointId": "uat46f9f8f82fd5c1b42a7de31b5dc2c11ef418a62f",
-      "PublisherId": "uat46f9f8f82fd5c1b42a7de31b5dc2c11ef418a62f",
-      "DataSetWriterId": "uat46f9f8f82fd5c1b42a7de31b5dc2c11ef418a62f"
+      "PublisherId": "opc.tcp://opcplc:50000_D09D61EF",
+      "DataSetWriterId": "1000"
     }
   },
-  "enqueuedTime": "2020-03-25T00:31:00.078Z",
+  "enqueuedTime": "Fri Mar 18 2022 13:52:42 GMT+0100 (Central European Standard Time)",
```
karok2m, line 32: Is this really how `enqueuedTime` is reported ? Should we fix it to have the previous format which is in line with the rest of timestamps that we have ?

cristipogacean: yes this is copy/paste from AzureIoT Explorer. Probably they updated the way enqueuedTime is written

karok2m: So this is not a change in the formatting that we did, right ?

karok2m (approved):

PR #1659, opened 2022-03-29
Added full-name command line options to parsing.

karok2m:
Changes:
* Added keys from `SecurityConfig`.
* Added keys from `ModuleConfig`.
* Added keys from `ClientServicesConfig`.
* Added keys from `TransportQuotaConfig`.
* Added keys from `StandaloneCliConfigKeys`. Added some uppercase keys as the parsing is case-sensitive.


modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
```diff
@@ -46,82 +46,113 @@ public class StandaloneCliOptions : Dictionary<string, string>, IAgentConfigProv
         /// <param name="args">The specified command line arguments.</param>
         public StandaloneCliOptions(string[] args) {
 
+            _logger = ConsoleLogger.Create(LogEventLevel.Warning);
```
karok2m, line 51: We should initialize logger in the other constructor as well.

karok2m: Done.

cristipogacean (approved):
:shipit:

dacolgit (approved):

PR #1672, opened 2022-04-25
[Publisher] add support for message routing info in telemetry

cristipogacean:
- adding the dataSetWriterGroup info as routing information helper for telemetry generated by the publisher.
- make this optional setting via command line arguments.


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
```diff
@@ -388,6 +389,8 @@ public class DataFlowProcessingEngine : IProcessingEngine, IDisposable {
         /// </summary>
         private readonly int _maxOutgressMessages;
 
+        private readonly bool _hasRoutingInfo;
```
karok2m: `_hasRoutingInfo` does not seem to be used. Can we remove it ?

api/src/Microsoft.Azure.IIoT.Api/src/Publisher/Models/EngineConfigurationApiModel.cs
```diff
@@ -48,5 +47,12 @@ public class EngineConfigurationApiModel {
         [DataMember(Name = "maxOutgressMessages", Order = 4,
             EmitDefaultValue = false)]
         public int? MaxOutgressMessages { get; set; }
+
+        /// <summary>
+        /// /// Flag to determine if a telemetry routing info is required.
+        /// </summary>
+        [DataMember(Name = "hasRoutingInfo", Order = 5,
+            EmitDefaultValue = false)]
+        public bool? HasRoutingInfo { get; set; }
```
karok2m: Since it is a enable/disable flag, shall we call it `EnableRoutingInfo` instead ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/IEngineConfiguration.cs
```diff
@@ -37,5 +37,10 @@ public interface IEngineConfiguration {
         /// Default: 4096 messages with 256KB ends up in 1 GB memory consumed.
         /// </summary>
         int? MaxOutgressMessages { get; }
+
+        /// <summary>
+        /// Flag to determine if a telemetry routing info is required.
+        /// </summary>
+        bool? HasRoutingInfo { get; }
```
karok2m: Since it is a enable/disable flag, shall we call it `EnableRoutingInfo` instead ?

docs/modules/publisher-commandline.md
```diff
@@ -444,6 +444,16 @@ The `Alternative` field, where present, refers to the CLI argument applicable in
                                       Mode: Standalone
                                       Type: boolean
                                       Default: false
+
+            HasRoutingInfo=VALUE
+                                      adds the routing info to telemetry messages. The name of the property is
```
karok2m: Let's start the message with an uppercase character.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/MonitoredItemMessageEncoder.cs
```diff
@@ -163,7 +164,8 @@ public class MonitoredItemMessageEncoder : IMessageEncoder {
                         ContentEncoding = "utf-8",
                         Timestamp = DateTime.UtcNow,
                         ContentType = _jsonContentType,
-                        MessageSchema = MessageSchemaTypes.MonitoredItemMessageJson
+                        MessageSchema = MessageSchemaTypes.MonitoredItemMessageJson,
+                        RoutingInfo = routingInfo,
```
karok2m: It seems like in both `MonitoredItemMessageEncoder` and `NetworkMessageEncoder` classes we are **always** setting `RoutingInfo` property which is then **always** applied by `IoTHubMessageSink.CreateMessage()` method. Shouldn't this happen only when the feature is enabled ? I think that we need to add a condition either in encoders or sink to add IoT Hub message property only when requested.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Services/PublisherJobService.cs
```diff
@@ -275,7 +275,8 @@ public sealed class PublisherJobService : IPublishServices<string> {
                             BatchTriggerInterval = DefaultBatchTriggerInterval.Value,
                             DiagnosticsInterval = TimeSpan.FromSeconds(60),
                             MaxMessageSize = 0,
-                            MaxOutgressMessages = DefaultMaxOutgressMessages.Value
+                            MaxOutgressMessages = DefaultMaxOutgressMessages.Value,
+                            HasRoutingInfo = false,
                     };
```
karok2m: Seems like this `};` is missing one additional level of indentation.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Services/PublisherJobService.cs
```diff
@@ -275,7 +275,8 @@ public sealed class PublisherJobService : IPublishServices<string> {
                             BatchTriggerInterval = DefaultBatchTriggerInterval.Value,
                             DiagnosticsInterval = TimeSpan.FromSeconds(60),
                             MaxMessageSize = 0,
-                            MaxOutgressMessages = DefaultMaxOutgressMessages.Value
+                            MaxOutgressMessages = DefaultMaxOutgressMessages.Value,
+                            HasRoutingInfo = false,
```
karok2m: Just going to put it out there, though I guess the answer is no since we do not plan to add any feature to the cloud microservices ... Should we add an env var for configuring this value on the cloud side ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/IoTHubMessageSink.cs
```diff
@@ -67,11 +68,21 @@ public class IoTHubMessageSink : IMessageSink, IDisposable {
...
-                            await _clientAccessor.Client.SendEventBatchAsync(messageObjects).ConfigureAwait(false);
+                            if (messagesCount == 1) {
+                                await _clientAccessor.Client.SendEventAsync(routingInfo, messageObjects.First()).ConfigureAwait(false);
+                            }
+                            else {
+                                await _clientAccessor.Client.SendEventBatchAsync(routingInfo, messageObjects).ConfigureAwait(false);
+                            }
                         }
                     }
                     catch (Exception e) {
```
karok2m, line 88: Hm, don't we have a metrics counter that we should increase in exceptional cases when we fail to send the message ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/IoTHubMessageSink.cs
```diff
@@ -67,11 +68,21 @@ public class IoTHubMessageSink : IMessageSink, IDisposable {
...
+                                await _clientAccessor.Client.SendEventAsync(messageObjects.First()).ConfigureAwait(false);
+                            }
+                            else {
+                                await _clientAccessor.Client.SendEventBatchAsync(messageObjects).ConfigureAwait(false);
+                            }
                         }
                         else {
-                            await _clientAccessor.Client.SendEventBatchAsync(messageObjects).ConfigureAwait(false);
+                            if (messagesCount == 1) {
+                                await _clientAccessor.Client.SendEventAsync(routingInfo, messageObjects.First()).ConfigureAwait(false);
```
karok2m, line 81: I would argue that we should not use methods that accept `outputName` parameter here as that is not supported when `DeviceClient` is used. That means that OPC Publisher will not be able to send data when it is deployed using device credentials, which is often the case when running it outside of IoT Edge context. That is also the situation that we have when debugging locally, which would be very complicated with this change.

api/src/Microsoft.Azure.IIoT.Api/src/Publisher/Models/EngineConfigurationApiModel.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/DataFlowProcessingEngine.cs
cristipogacean: sure, done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/MonitoredItemMessageEncoder.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/IEngineConfiguration.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Services/PublisherJobService.cs
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Services/PublisherJobService.cs
cristipogacean: you're right: the answer is no :). as an emergency people can use the cosmosdb directly to control the EnabeRoutingInfo.

docs/modules/publisher-commandline.md
cristipogacean: done

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/IoTHubMessageSink.cs
cristipogacean: nop...

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/IoTHubMessageSink.cs
cristipogacean: fixed

common/src/Microsoft.Azure.IIoT.Hub.Module.Client/src/Default/IoTSdkFactory.cs
```diff
@@ -11,6 +11,7 @@ namespace Microsoft.Azure.IIoT.Module.Framework.Client {
     using Microsoft.Azure.Devices.Client;
     using Microsoft.Azure.Devices.Client.Transport.Mqtt;
     using Microsoft.Azure.Devices.Shared;
+    using Prometheus;
```
karok2m, line 14: Is this using needed ? I do not see changes here related to metrics.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/IoTHubMessageSink.cs
```diff
@@ -8,15 +8,14 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine {
     using Microsoft.Azure.IIoT.Module.Framework.Client;
     using Microsoft.Azure.IIoT.Hub;
     using Microsoft.Azure.Devices.Client;
+    using Prometheus;
```
karok2m, line 11: Same comment as above. I do not see changes related to metrics. So is this using needed ?

karok2m (approved):

PR #1686, opened 2022-05-04
removing timestamps from metrics and updating dashboard queries

monikavar:
_Context:_
Earlier the timestamps in the label were added so that we could get the exact timestamp of reconnections/disconnections. 

_Functioning:_
Although the reconnections are in few hundreds, since each of them is treated as a new metric, the metrics collector module sends all of them at every scraped interval,  resulting in hundreds of thousands of entries which in turn increases the cost of storage.

_Action:_
Removing the timestamps from the affected queries and updating the dashboard queries for the same.


cristipogacean (approved):
LGTM

dacolgit (approved):

PR #1697, opened 2022-05-12
Skipping reversible encoding tests for events.

karok2m:


modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/tests/ReversibleEncodingIntegrationTests.cs
```diff
@@ -23,18 +23,23 @@ public class ReversibleEncodingIntegrationTests : PublisherIntegrationTestBase,
...
 
-        [Theory]
+        // ToDo: Parts of job converter for OpcEvents are commented out for now.
+        [Theory(Skip = "PublishedNodesJobConverter does not parse OpcEvents now.")]
         [InlineData(@"./PublishedNodes/SimpleEvents.json")]
         public async Task CanEncodeWithoutReversibleEncodingTest(string publishedNodesFile) {
             // Arrange
             // Act
-            var result = await ProcessMessagesAsync(publishedNodesFile, arguments: new[] { "--mm=PubSub", "--re=False" });
+            var result = await ProcessMessagesAsync(publishedNodesFile, arguments: new[] { "--mm=PubSub", "--UseReversibleEncoding=False" });
```
cristipogacean, line 32: shouldn't we have this as a definition flag, like we have the "EnableRoutingInfo" instead of a boolean?

karok2m: I prefer having parameters where we can set the value to both `true` and `false`, specifically for the long-form ones such as `UseReversibleEncoding`. This is actually the mechanism that we use for the majority of the parameters (9 against 4).

cristipogacean: agreed, probably I should then change the EnableRoutingInfo to a bool argument for consistency.

cristipogacean (approved):

dacolgit (approved):

PR #1700, opened 2022-05-13
Fixed HttpTunnelTests.TestGetWebAsync() test.

karok2m:
Changes:
* Checking for some other data in `payload` as `<!DOCTYPE html>` does not seem to always be present.


cristipogacean (approved):
good choice! `Microsoft` will stay there even after they replace html with whatever metaverse will bring :)

dacolgit (approved):

monikavar (approved):

Issue #1701, opened 2022-05-16
OPC Publisher 2.8.2: Could not send worker heartbeat - eventually crashing and not restarting


gregeva:
**Describe the bug**
Repeatedly receiving the error message "**Could not send worker heartbeat. 
System.NullReferenceException: Object reference not set to an instance of an object.**" from the OPC Publisher (version 2.8.2).

```
[16:31:47 ERR Microsoft.Azure.IIoT.Agent.Framework.Agent.Worker+JobProcess] Could not send worker heartbeat. 
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine.WriterGroupMessageTrigger.get_AuthenticationUsername() in D:\a\1\s\components\opc-ua\src\Microsoft.Azure.IIoT.OpcUa.Edge.Publisher\src\Engine\WriterGroupMessageSource.cs:line 61
   at Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine.DataFlowProcessingEngine.GetDiagnosticInfo() in D:\a\1\s\components\opc-ua\src\Microsoft.Azure.IIoT.OpcUa.Edge.Publisher\src\Engine\DataFlowProcessingEngine.cs:line 171
   at Microsoft.Azure.IIoT.Agent.Framework.Agent.Worker.JobProcess.GetProcessDiagnosticInfo() in D:\a\1\s\common\src\Microsoft.Azure.IIoT.Agent.Framework\src\Agent\Default\Worker.cs:line 374
   at Microsoft.Azure.IIoT.Agent.Framework.Agent.Worker.JobProcess.SendHeartbeatAsync(CancellationToken ct) in D:\a\1\s\common\src\Microsoft.Azure.IIoT.Agent.Framework\src\Agent\Default\Worker.cs:line 465
   at Microsoft.Azure.IIoT.Agent.Framework.Agent.Worker.JobProcess.OnHeartbeatTimerAsync() in D:\a\1\s\common\src\Microsoft.Azure.IIoT.Agent.Framework\src\Agent\Default\Worker.cs:line 514

```

**To Reproduce**

1. I had subscribed to around 55 new OPC tags with a 500ms poll rate
2. Updated changes received in the publisher and started scanning the OPC UA server
3. Started receiving published updates in VS Code event hub monitoring
4. OPC Publisher logs started throwing the worker heartbeat error
[...]

PR #1702, opened 2022-05-16
fix for #1701

cristipogacean:
fix crash in case of invalid user authentication configuration

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/WriterGroupMessageSource.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/WriterGroupMessageSource.cs
@@ -54,18 +54,18 @@ namespace Microsoft.Azure.IIoT.OpcUa.Edge.Publisher.Engine {
                 true : false;
 
         /// <inheritdoc/>
-        public OpcAuthenticationMode AuthenticationMode =>_subscriptions?.First()?.Subscription?.Connection.User != null ?
+        public OpcAuthenticationMode AuthenticationMode =>_subscriptions?.First()?.Subscription?.Connection?.User != null ?
                 OpcAuthenticationMode.UsernamePassword : OpcAuthenticationMode.Anonymous;
 
         /// <inheritdoc/>
-        public string AuthenticationUsername =>_subscriptions?.First()?.Subscription?.Connection.User != null ?
+        public string AuthenticationUsername =>_subscriptions?.First()?.Subscription?.Connection?.User?.Value != null ?
                 _serializer.Deserialize<cred>(_subscriptions?
                     .First()?
                     .Subscription?
-                    .Connection
-                    .User
-                    .Value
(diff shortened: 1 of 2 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/WriterGroupMessageSource.cs
```diff
@@ -54,18 +54,18 @@ public class WriterGroupMessageTrigger : IMessageTrigger, IDisposable {
                 true : false;
 
         /// <inheritdoc/>
-        public OpcAuthenticationMode AuthenticationMode =>_subscriptions?.First()?.Subscription?.Connection.User != null ?
+        public OpcAuthenticationMode AuthenticationMode =>_subscriptions?.First()?.Subscription?.Connection?.User != null ?
```
karok2m: Should we use `_subscriptions?.First()?.Subscription?.Connection?.User?.Value` to keep the condition of `OpcAuthenticationMode` symmetric to that of `AuthenticationUsername` ?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/WriterGroupMessageSource.cs
```diff
@@ -54,18 +54,18 @@ public class WriterGroupMessageTrigger : IMessageTrigger, IDisposable {
...
-        public string AuthenticationUsername =>_subscriptions?.First()?.Subscription?.Connection.User != null ?
+        public string AuthenticationUsername =>_subscriptions?.First()?.Subscription?.Connection?.User?.Value != null ?
                 _serializer.Deserialize<cred>(_subscriptions?
                     .First()?
                     .Subscription?
-                    .Connection
-                    .User
-                    .Value
-                    .ToJson())
+                    .Connection?
```
karok2m: The condition above makes sure that `_subscriptions.First().Subscription.Connection.User.Value` is not `null` so we can remove the `?` checks up to `ToJson()` here.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/WriterGroupMessageSource.cs
cristipogacean: agreed

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/WriterGroupMessageSource.cs
cristipogacean: agreed, thought originally better safe than sorry, but you're right.

karok2m (approved):

dacolgit (approved):

PR #1719, opened 2022-05-24
Bugfix/orchestratorbadrequest

cristipogacean:
fix CosmosDB continuation token handling

### DIFF
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Jobs/Default/DefaultJobOrchestrator.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Jobs/Default/DefaultJobOrchestrator.cs
@@ -49,6 +49,7 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Jobs {
             do {
                 var jobList = await _jobRepository.QueryAsync(query, continuationToken, null, ct);
                 if (jobList?.Jobs == null) {
+                    continuationToken = null;
                     continue;
                 }
                 System.Diagnostics.Debug
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Jobs/Default/DefaultJobOrchestrator.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Jobs/Default/DefaultJobOrchestrator.cs
@@ -49,8 +49,7 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Jobs {
             do {
                 var jobList = await _jobRepository.QueryAsync(query, continuationToken, null, ct);
                 if (jobList?.Jobs == null) {
-                    continuationToken = null;
-                    continue;
+                    break;
                 }
                 System.Diagnostics.Debug
                     .Assert(!jobList.Jobs.Any(j => j.LifetimeData.Status != JobStatus.Active));
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Jobs/Default/DefaultJobOrchestrator.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Jobs/Default/DefaultJobOrchestrator.cs
@@ -48,6 +48,9 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Jobs {
             string continuationToken = null;
             do {
                 var jobList = await _jobRepository.QueryAsync(query, continuationToken, null, ct);
+                if (jobList?.Jobs == null) {
+                    continue;
+                }
                 System.Diagnostics.Debug
                     .Assert(!jobList.Jobs.Any(j => j.LifetimeData.Status != JobStatus.Active));
 
--- a/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Storage/Database/Services/JobDatabase.cs
+++ b/common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Storage/Database/Services/JobDatabase.cs
@@ -27,6 +27,7 @@ namespace Microsoft.Azure.IIoT.Agent.Framework.Storage.Database {
             var dbs = databaseServer.OpenAsync(databaseJobRepositoryConfig.DatabaseName).Result;
             var cont = dbs.OpenContainerAsync(databaseJobRepositoryConfig.ContainerName).Result;
             _documents = cont.AsDocuments();
+            _client = _documents.OpenSqlClient();
         }
 
         /// <inheritdoc/>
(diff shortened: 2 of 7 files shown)


dacolgit (dismissed):

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Jobs/Default/DefaultJobOrchestrator.cs
```diff
@@ -48,6 +48,9 @@ public class DefaultJobOrchestrator : IJobOrchestrator {
             string continuationToken = null;
             do {
                 var jobList = await _jobRepository.QueryAsync(query, continuationToken, null, ct);
+                if (jobList?.Jobs == null) {
+                    continue;
```
karok2m: We should also assign value to `continuationToken` here so that loop is broken when there are no more entries to retrieve.

```
ontinuationToken = jobList.ContinuationToken;
```

cristipogacean: agreed. actually, the continuationToken should be reset to null. done that.

common/src/Microsoft.Azure.IIoT.Storage.CosmosDb/src/Services/DocumentQuery.cs
```diff
@@ -60,7 +60,8 @@ internal sealed class DocumentQuery : ISqlClient {
 
                     EnableCrossPartitionQuery = pk == null
                 });
-            return new DocumentInfoFeed<T>(query.AsDocumentQuery(), _logger);
+            _query = query.AsDocumentQuery();
```
karok2m: `DocumentQuery` object is not tied to any specific query on initialization, which means/implies that it can be used to retrieve elements for different queries. If that is done, then those different queries will interfere with each other and `Continue()` for one of them may return results for the other query. And with the current interface of the class there is nothing in place to stop that scenario from happening. I think that we have a few options here:

* couple `DocumentQuery` object to a specific query
* query string should be passed to `Continue()` method and corresponding `IDocumentQuery<Document>` object should be recreated
* query string should be passed to `Continue()` method and corresponding `IDocumentQuery<Document>` object can be stored and retrieved from an internal cache that would tie it with the original query string

The other problem that I see with storing `_query` is that its continuation token can be used only once. So if there are many matching elements in the database, then 2nd call to `Continue()` bellow will most probably still get the results for 1st call as the continuation token has not been updated. The fact that `Continue()` bellow doesn't actually use passed `continuationToken` also points in that direction, unless I am missing bigger picture here.

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Storage/Database/Services/JobDatabase.cs
```diff
@@ -134,10 +135,9 @@ public class JobDatabase : IJobRepository {
         /// <inheritdoc/>
         public async Task<JobInfoListModel> QueryAsync(JobInfoQueryModel query,
             string continuationToken, int? maxResults, CancellationToken ct) {
-            var client = _documents.OpenSqlClient();
             var results = continuationToken != null ?
-                client.Continue<JobDocument>(continuationToken, maxResults) :
-                client.Query<JobDocument>(CreateQuery(query, out var queryParameters),
+                _client.Continue<JobDocument>(continuationToken, maxResults) :
```
karok2m: Since the `_client` stores the query internally it seems like we need to have a set of them each for a specific query. Otherwise two separate calls to `JobDatabase.QueryAsync()` method will cause conflicts and one of them will return elements for the other query on 2nd call.

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Jobs/Default/DefaultJobOrchestrator.cs
karok2m: Can we actually exclude cases where `jobList.Jobs` would be `null` but it has a continuation token ? To be honest, that would be a weird scenario and wouldn't make much sense, but I am wondering if we can rely on that never happening. Also, if we set `continuationToken` to null, isn't that effectively a `break` instead of `continue` ?

cristipogacean: ok, I guess break will do it.

common/src/Microsoft.Azure.IIoT.Agent.Framework/src/Storage/Database/Services/JobDatabase.cs
cristipogacean: no, the client shall be reused for multiple queries.

common/src/Microsoft.Azure.IIoT.Storage.CosmosDb/src/Services/DocumentQuery.cs
karok2m: @cristipogacean, one more point, `JobDatabase.QueryAsync()` is called from several places. One path is `WorkersController.GetAvailableJobAsync()` -> `DefaultJobOrchestrator.GetAvailableJobAsync()`, and the others are through `JobsController` -> `DefaultJobService` via `ListJobsAsync()` and `QueryJobsAsync()` methods. And in case of services-all they are all going to use the same instance of `JobDatabase`. So we should make it and the client that it uses stateless as there will be concurrent calls from those API controllers to the `JobDatabase` with different queries.

cristipogacean: I believe the `Continue()` logic continuation token logic is designed to be executed on the very same _query object without the need to interfere with it's internal continuationToken state. The actual implementation was build on different assumptions.

the actual implementation works just fine with multiple calls with continuation tokens on the same query.

cristipogacean: @karok2m, yes, you're right, I switched back to stateless.

common/src/Microsoft.Azure.IIoT.Storage.CosmosDb/src/Services/DocumentQuery.cs
```diff
@@ -60,26 +60,35 @@ internal sealed class DocumentQuery : ISqlClient {
...
-        public IResultFeed<IDocumentInfo<T>> Continue<T>(string continuationToken,
+        public IResultFeed<IDocumentInfo<T>> Continue<T>(string queryString,
+            string continuationToken, IDictionary<string, object> parameters,
             int? pageSize, string partitionKey) {
             if (string.IsNullOrEmpty(continuationToken)) {
                 throw new ArgumentNullException(nameof(continuationToken));
             }
             var pk = _partitioned || string.IsNullOrEmpty(partitionKey) ? null :
-                new PartitionKey(partitionKey);
+                            new PartitionKey(partitionKey);
```
karok2m, line 75: I am a bit confused by this condition, not here specifically, but in the whole class. If `_partitioned` is `true` then we would set `pk` to `null`. I was thinking that we would set the partition key (`pk`) in case when `_partitioned` is `true` and non-empty `partitionKey` is provided. Not an issue for this PR but wanted to raise the question so that we test this logic at some point.

karok2m (approved):

common/src/Microsoft.Azure.IIoT.Storage.CosmosDb/src/Services/DocumentQuery.cs
cristipogacean: it's not clear t me as well, we should handle that in a separate PR though.

cristipogacean: it's not clear to me as well, we should handle that in a separate PR though.

PR #1725, opened 2022-05-30
Enabling OPC Events e2e tests

dacolgit:


cristipogacean (approved):
:shipit:

karok2m (approved):

karok2m:
@dacolgit, can you please also run the Standalone E2E pipeline and post a link here ? Let's make sure that it is passing without any issues.

e2e-tests/IIoTPlatform-E2E-Tests/TestExtensions/IIoTPlatformTestContext.cs
```diff
@@ -106,27 +106,27 @@ public class IIoTPlatformTestContext : IDisposable, IDeviceConfig, IIoTHubConfig
         /// <summary>
         /// Urls for the dynamic ACI containers
         /// </summary>
-        public string[] PlcAciDynamicUrls { get; set; }
```
karok2m, line 109: What were the build warnings here ? With nullable values we may need to start checking if they actually have value. Might be better if we can avoid that.

dacolgit: these for example are the warnings:

image

karok2m: Understood after discussing this with @dacolgit. This uses the same pattern as other similar variables in `IIoTPlatformTestContext` class.

monikavar (approved):

PR #1727, opened 2022-06-01
fix version in operation manual

dacolgit:


version.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "2.8.3-rc.{height}",
```
karok2m, line 3: Please revert this change. Main should use `2.8.3-rc.{height}` format.

readme.md
```diff
@@ -1,96 +1,2323 @@
-# Azure Industrial IoT Platform
+[Home](../../readme.md)
```
karok2m, line 1: @dacolgit, I believe that the intent was to change operations manual which is `docs/manual/readme.md`, right ? Seems like the updated content from there has been copy-pasted to `readme.md`. Can you please fix this and apply the changes in `docs/manual/readme.md` ?

version.json
karok2m: BTW, this change can be coming from the fact that your branch is based on `release/2.8.3_rc` instead of `main`.

dacolgit: yes

PR #1748, opened 2022-06-14
Helm: Change the chart to deploy 2.8.3 version of Azure IIoT platform.

karok2m:

PR #1763, opened 2022-07-26
Update to .net 6 and latest dependencies where applicable

marcschier:
* Update all dependencies and .net runtime to .net 6., including msgpack
* Rate limiting for edge publisher to limit concurrent requests to max 400 (can be changed using env, with 0 to disable)
  * Retry with exponential backoff from OPC Publisher to orchestration endpoint in conjunction with rate limiting
  * Better logging and additional metrics
* IoT Edge 1.4 support including eflow, including in simulation
* Robustness changes in TestEventProcessor, fix discovery end to end tests, update e2e deployment of PLC, automate variable settings in e2e testing to get tests passing reliably


common/src/Microsoft.Azure.IIoT.AspNetCore/src/RateLimiting/RateLimitingSetup.cs
```diff
@@ -0,0 +1,74 @@
...
+                await _next.Invoke(context);
+            }
+            else {
+                try {
+                    if (Interlocked.Increment(ref _concurrentCalls) > _maxConcurrentCalls) {
+                        if (0 == Interlocked.CompareExchange(ref _throttling, 1, 0)) {
+                            _logger.Information("Start throttling requests due to too many concurrent calls.");
+                        }
+                        kRateLimitedCalls.Inc();
+                        context.Response.StatusCode = 429;
```
cristipogacean: is there an enum that could be used instead 429?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Deploy/IoTHubDiscovererDeployment.cs
```diff
@@ -115,20 +115,19 @@ public sealed class IoTHubDiscovererDeployment : IHostProcess {
                 });
             }
             else {
-                // Windows
+                // Windows (Eflow)
```
cristipogacean, line 118: Does this mean we discontinue windows containers with this version? if so, we should probably add a little migration tutorial.

cristipogacean: On a second thought, if we discontinue windows containers, then the Linux vs Windows configurations should be identical, therefore the entire logic here could be optimized.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Twin/src/Deploy/IoTHubSupervisorDeployment.cs
```diff
@@ -103,20 +104,19 @@ public sealed class IoTHubSupervisorDeployment : IHostProcess {
                 });
             }
             else {
-                // Windows
+                // Eflow
```
cristipogacean, line 107: I guess since eflow configuration is identical the Linux one, so we can simplify the code here.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Registry/src/Deploy/IoTHubDiscovererDeployment.cs
marcschier: I had it simplified, but /mount is readonly on eflow.  I dont want to break existing linux deployments.

common/src/Microsoft.Azure.IIoT.AspNetCore/src/RateLimiting/RateLimitingSetup.cs
marcschier: Good one, will change to HttpStatusCode enum.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Twin/src/Deploy/IoTHubSupervisorDeployment.cs
marcschier: I had it simplified, but /mount is readonly on eflow. I dont want to break existing linux deployments.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
```diff
@@ -859,22 +859,23 @@ internal sealed class SubscriptionWrapper : ISubscription {
                     var publishTime = (notification?.MonitoredItems?.First().Message?.PublishTime).
                         GetValueOrDefault(DateTime.UtcNow);
 
-                    if (isKeepAlive || _expectedSequenceNumber == sequenceNumber) {
+                    if (isKeepAlive) {
```
marcschier, line 862: @cristipogacean, could you also review this change?  This should fix erroneous logs in current publisher where we were complaining about the sequence number in the first message after a keep alive message.

deploy/scripts/default.yml
```diff
@@ -1,19 +1,19 @@
 version: '3'
 services:
   opcserver0:
-    image: mcr.microsoft.com/iotedge/opc-plc:latest
+    image: mcr.microsoft.com/iotedge/opc-plc:2.2.0
```
karok2m, line 4: Why not use `latest` here ?

deploy/src/Microsoft.Azure.IIoT.Deployment/Configuration/HelmSettings.cs
```diff
@@ -13,7 +13,7 @@ class HelmSettings {
         // Defaults
         public static string _defaultRepoUrl = "https://azure.github.io/Industrial-IoT/helm";
         public static string _defaultChartVersion = "0.4.3";
-        public static string _defaultImageTag = "2.8.3";
+        public static string _defaultImageTag = "2.8.4";
```
karok2m: Our latest release is still `2.8.3`. I would recommend to use that value. Otherwise deployment without an image tag will fail now. We usually update those values once a new version of Helm chart is released.

deploy/src/Microsoft.Azure.IIoT.Deployment/Resources/scripts/jumpbox.sh
```diff
@@ -323,7 +323,7 @@ if [ "$is_private_repo" = true ] ; then
     echo $setsvc
 
     # Install aiiot/azure-industrial-iot Helm chart
-    helm install --atomic azure-industrial-iot aiiot/azure-industrial-iot --namespace azure-industrial-iot --version $HELM_CHART_VERSION --timeout 30m0s \
```
karok2m: @marcschier, I would advice to keep the `--atomic` flag here. It makes sure that if Helm chart deployment did not succeed in `30` minutes then a failure is reported. Without it we will have scenarios when providing invalid Helm configuration details (such as invalid image, image tag or version) will not be reported as error and deployment tool will continue without notifying the caller that something went wrong.

docs/deploy/howto-add-aks-to-ps1.md
```diff
@@ -309,7 +309,7 @@ Note the following values in the YAML file:
 
 ```yaml
 image:
-  tag: 2.8.3
```
karok2m: We usually update those values only after we have a corresponding release. Please see #1748 for a sample of such change.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/SubscriptionServices.cs
cristipogacean: indeed, this looks like it will fix that log.

deploy/scripts/default.yml
marcschier: Latest is broken because it (correctly) validates its own certificate against it's host name, which is always wrong in how we deploy.  The fix (future) is to add the IP to the certificate somehow.  However, it is a larger work item.

deploy/src/Microsoft.Azure.IIoT.Deployment/Configuration/HelmSettings.cs
marcschier: Will do

deploy/src/Microsoft.Azure.IIoT.Deployment/Resources/scripts/jumpbox.sh
marcschier: What I see is that any error that occurs is rolled back and thus difficult to diagnose on the cluster, meaning we would have a hard time telling what went wrong in the field.  But no problem to keep status quo.

docs/deploy/howto-add-aks-to-ps1.md
marcschier: Will do

cristipogacean (dismissed):

cristipogacean (approved):

karok2m (approved):

PR #1830, opened 2022-10-11
Support data change trigger configuration in node

### LINKED ISSUE #1074: OPCPublisher: is it possible to monitor items on DataChangeTriggerType.StatusValueTimestamp ?
Hi!

Is it possible to monitor items on DataChangeTriggerType.StatusValueTimestamp ?

From the OPC UA Specification we have that the default behavior is to ["Report a notification if either the StatusCode or the value change. "](https://reference.opcfoundation.org/v104/Core/docs/Part4/7.17.2/) (i.e., _Microsoft.Azure.IIoT.OpcUa.Publisher.Models.DataChangeTriggerType.StatusValue_).

I was wondering if there's any mechanism available when configuring the OPCPublisher to define that we want to be notified on _SourceTimestamp_ changes also (i.e., DataChangeTriggerType.StatusValueTimestamp).

Many thanks!
Amparo

beandrad:
So that the OPC Publisher can be configured to receive a notification on Status, Status/Value or Status/Value/Timestamp change.

This change adds the field `"OpcNodes[].DataChangeTrigger"` that can take the value "Status", "StatusValue" or "StatusValueTimestamp". This allows the configuration of the monitored item data change trigger. It can be left null, in which case the default defined in the OPC UA specs is "StatusValue".

Fixes #1074


Ellerbach (dismissed):
Nice addition to fix this requested feature. Seems very smooth in terms of implementation and following the value change.

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher.csproj
```diff
@@ -7,6 +7,7 @@
     <Description>Azure Industrial IoT OPC Publisher API</Description>
   </PropertyGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\..\..\components\opc-ua\src\Microsoft.Azure.IIoT.OpcUa\src\Microsoft.Azure.IIoT.OpcUa.csproj" />
```
cristipogacean: We should only take dependencies on the API assemblies here, not to internal ones.

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/PublishedNodeApiModel.cs
```diff
@@ -77,5 +79,10 @@ public class PublishedNodeApiModel {
         [DataMember(Name = "queueSize", Order = 11,
             EmitDefaultValue = false)]
         public uint? QueueSize { get; set; }
+        
+        /// <summary> Data change filter </summary>
+        [DataMember(Name = "dataChangeFilter", Order = 12,
+            EmitDefaultValue = false)]
+        public DataChangeTriggerType? DataChangeFilter { get; set; }
```
cristipogacean: >DataChangeTriggerType [](http://example.com/codeflow?start=15&length=21)

Let's define an api model for this enum and not use the internal model.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/DataChangeTriggerType.cs
```diff
@@ -3,25 +3,31 @@
 //  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
 // ------------------------------------------------------------
 
+using System.Runtime.Serialization;
+
 namespace Microsoft.Azure.IIoT.OpcUa.Publisher.Models {
     /// <summary>
     /// Data change trigger
     /// </summary>
+    [DataContract]
```
cristipogacean: We should have this in the API model only. same for the attributes below.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Models/StandaloneCliModel.cs
```diff
@@ -56,6 +56,11 @@ public class StandaloneCliModel {
         /// </summary>
         public uint DefaultQueueSize { get; set; } = 1;
 
+        /// <summary>
+        /// set the default data change filter for monitored items
+        /// </summary>
+        public DataChangeTriggerType DefaultDataChangeFilter { get; set; } = DataChangeTriggerType.StatusValue;
```
cristipogacean: >DefaultDataChangeFilter [](http://example.com/codeflow?start=37&length=23)

We should have the this default configurable per environment variable. Like for example we do for the DefaultSamplingInterval

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesJobConverter.cs
```diff
@@ -326,6 +327,7 @@ private class OpcNodeModelComparer : IEqualityComparer<(string, OpcNodeModel)> {
                             QueueSize = node.QueueSize ?? standaloneCliModel.DefaultQueueSize,
                             // ToDo: Implement mechanism for SkipFirst.
                             SkipFirst = node.SkipFirst ?? standaloneCliModel.DefaultSkipFirst,
+                            DataChangeFilter = node.DataChangeFilter ?? DataChangeTriggerType.StatusValue
```
cristipogacean: >node.DataChangeFilter [](http://example.com/codeflow?start=47&length=21)

I think we should still allow `null` value here. This will represent the default behavior as specified here: [https://reference.opcfoundation.org/v104/Core/docs/Part4/7.17.2/](https://reference.opcfoundation.org/v104/Core/docs/Part4/7.17.2/). We should allow the user to decide if he/she goes for the default or StatusValue

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesJobConverter.cs
```diff
@@ -349,6 +351,7 @@ private class OpcNodeModelComparer : IEqualityComparer<(string, OpcNodeModel)> {
                                 QueueSize = node.QueueSize ?? standaloneCliModel.DefaultQueueSize,
                                 // ToDo: Implement mechanism for SkipFirst.
                                 SkipFirst = node.SkipFirst ?? standaloneCliModel.DefaultSkipFirst,
+                                DataChangeFilter = node.DataChangeFilter ?? DataChangeTriggerType.StatusValue
```
cristipogacean: >DataChangeFilter [](http://example.com/codeflow?start=56&length=16)

same as above

docs/modules/publishednodes_2.9.json
```diff
@@ -0,0 +1,124 @@
+// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
```
cristipogacean: Let's stick to the 2.8 for product version for now. We're not sure about the versioning of the next release.

docs/modules/publisher.md
```diff
@@ -118,8 +118,8 @@ The simplest way to configure OPC Publisher is via a configuration file. A basic
 ]
 ```
 
-Another example configuration files are provided via [`publishednodes_2.5.json`](publishednodes_2.5.json?raw=1)
-and [`publishednodes_2.8.json`](publishednodes_2.8.json?raw=1) files.
+Another example configuration files are provided via [`publishednodes_2.5.json`](publishednodes_2.5.json?raw=1),
+[`publishednodes_2.8.json`](publishednodes_2.8.json?raw=1) and [`publishednodes_2.9.json`](publishednodes_2.9.json?raw=1) files.
```
cristipogacean: >`publishednodes_2.9 [](http://example.com/codeflow?start=64&length=19)

let's keep version to 2.8 for now.

cristipogacean:
That's a really nice and helpful feature! Thank you very much for this contribution! 
I had a few comments that I hope you'll find reasonable.

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/PublishedNodeApiModel.cs
beandrad: There was already an enum defined in `api/src/Microsoft.Azure.IIoT.Api/src/Publisher/Models`; I moved it to api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models to avoid circular dependencies.

beandrad:
> That's a really nice and helpful feature! Thank you very much for this contribution! I had a few comments that I hope you'll find reasonable.

Thanks for the review @cristipogacean. I've made the requested updates; could you please have another look?

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/PublishedNodeApiModel.cs
marcschier: Can we call the property Data change trigger?  Data change filter is larger and includes dead band, so the name will be misleading.  Or we do the entire data change filter.

marcschier (changes requested):
See my comment re naming.  Plus an addition to our e2e tests would be good.

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/PublishedNodeApiModel.cs
beandrad: I've change it now; however, I don't understand why that name in the `PublishedDataSetVariableApiModel` (see code below), which doesn't seem to include dead band (as those are fields below). Should I also change it there and in the `OpcNodeModel`?

For reference:
 https://github.com/Azure/Industrial-IoT/blob/898257cf0ae44f8c4b2413154d0449f11198fe2c/api/src/Microsoft.Azure.IIoT.Api/src/Publisher/Models/PublishedDataSetVariableApiModel.cs#L75

https://github.com/Azure/Industrial-IoT/blob/898257cf0ae44f8c4b2413154d0449f11198fe2c/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesJobConverter.cs#L147

beandrad:
> See my comment re naming. Plus an addition to our e2e tests would be good.

Thanks for your review, @marcschier. I renamed the property to `DataChangeTrigger` and updated one of the existing e2e tests to check that setting the `DataChangeTrigger` in the published nodes file works. Unfortunately, I don't have access to the AzDO pipeline to provision the e2e test resources, so I run the scripts manually following the standalone pipeline definition (https://github.com/Azure/Industrial-IoT/blob/main/tools/e2etesting/pipeline_standalone.yml) . Apart from that I also had to manually build and push the opc publisher module docker image and make sure I was using this version for the e2e. I'm not sure if this is automated when these tests are run from the pipeline.

docs/modules/publisher-commandline.md
```diff
@@ -527,6 +535,10 @@ The `Alternative` field, where present, refers to the CLI argument applicable in
                                    can not be sent quick enough to IoTHub
                                    Min: 1024
                                    Default: 8192
+          --mc, --monitoreditemdatachangetrigger=VALUE
```
marcschier: Was this command line option part of OPC Publisher 2.5?

docs/modules/publisher.md
```diff
@@ -118,8 +118,7 @@ The simplest way to configure OPC Publisher is via a configuration file. A basic
 ]
 ```
 
-Another example configuration files are provided via [`publishednodes_2.5.json`](publishednodes_2.5.json?raw=1)
-and [`publishednodes_2.8.json`](publishednodes_2.8.json?raw=1) files.
+Another example configuration files are provided via [`publishednodes_2.5.json`](publishednodes_2.5.json?raw=1) and [`publishednodes_2.8.json`](publishednodes_2.8.json?raw=1).
```
marcschier: Nit: Other instead of Another, what do you think?

e2e-tests/IIoTPlatform-E2E-Tests/TestHelper.cs
```diff
@@ -3,6 +3,8 @@
 //  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
 // ------------------------------------------------------------
 
+using Microsoft.Azure.IIoT.OpcUa.Api.Publisher.Models;
```
marcschier: Move into the namespace declaration

e2e-tests/IIoTPlatform-E2E-Tests/Standalone/A_PublishSingleNodeStandaloneTestTheory.cs
```diff
@@ -3,6 +3,8 @@
 //  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
 // ------------------------------------------------------------
```
marcschier: Move into the namespace declaration and remove trailing whitespace.

e2e-tests/IIoTPlatform-E2E-Tests/TestHelper.cs
```diff
@@ -627,6 +629,7 @@ private static PrivateKeyFile GetPrivateSshKey(IIoTPlatformTestContext context)
                     opcNode.OpcPublishingInterval = opcPlcPublishingInterval / 2;
                     opcNode.OpcSamplingInterval = opcPlcPublishingInterval / 4;
                     opcNode.QueueSize = 4;
+                    opcNode.DataChangeTrigger = dataChangeTrigger.ToString();
```
marcschier: I think this will crash when null.

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/DataChangeTriggerType.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.Azure.IIoT.OpcUa.Api.Publisher.Models {
     using System.Runtime.Serialization;
-
+
```
marcschier: Remove trailing whitespace

api/src/Microsoft.Azure.IIoT.OpcUa.Api.Publisher/src/Models/PublishedNodeApiModel.cs
```diff
@@ -3,6 +3,7 @@
 //  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
 // ------------------------------------------------------------
 
+
```
marcschier: Remove additional line break.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Models/StandaloneCliModel.cs
```diff
@@ -56,6 +56,11 @@ public class StandaloneCliModel {
         /// </summary>
         public uint DefaultQueueSize { get; set; } = 1;
 
+        /// <summary>
+        /// set the default data change filter for monitored items
+        /// </summary>
+        public DataChangeTriggerType? DefaultDataChangeTrigger { get; set; } = null;
```
marcschier: Initialization is not required.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesJobConverter.cs
```diff
@@ -144,6 +144,7 @@ public class PublishedNodesJobConverter {
                                     QueueSize = node.Item2.QueueSize,
                                     // ToDo: Implement mechanism for SkipFirst.
                                     SkipFirst = node.Item2.SkipFirst,
+                                    DataChangeFilter = node.Item2.DataChangeTrigger
```
marcschier: I think this should assign to "DataChangeTrigger" also.  Could you rename DataChangeFilter to DataChangeTrigger here as well, so it is clearer?

marcschier (changes requested):

docs/modules/publisher.md
beandrad: agree! :)

docs/modules/publisher-commandline.md
beandrad: ups, my bad, wrong section! I'll update. I'll just remove for the time being as I'm not sure under which version this feature is going to be released.

e2e-tests/IIoTPlatform-E2E-Tests/TestHelper.cs
beandrad: It actually returns and empty string (see below); I can change it tho to be more explicit.
image

e2e-tests/IIoTPlatform-E2E-Tests/TestHelper.cs
beandrad: changed

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesJobConverter.cs
beandrad: makes sense; updated!

api/src/Microsoft.Azure.IIoT.Api/src/Publisher/Extensions/ModelExtensions.cs
```diff
@@ -1413,7 +1413,7 @@ public static class ModelExtensions {
                 PublishedVariableNodeId = model.PublishedVariableNodeId,
                 BrowsePath = model.BrowsePath,
                 Attribute = model.Attribute,
-                DataChangeFilter = (OpcUa.Publisher.Models.DataChangeTriggerType?)model.DataChangeFilter,
+                DataChangeTrigger = (OpcUa.Publisher.Models.DataChangeTriggerType?)model.DataChangeFilter,
```
marcschier: Here is another of these, can you change the DataChangeFilter which is a DataChangeTrigger to be named DataChangeTrigger?

marcschier (changes requested):

marcschier (approved):

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliConfigKeys.cs
```diff
@@ -76,6 +76,11 @@ public static class StandaloneCliConfigKeys {
         /// Key for the default queue size fro monitored items
         /// </summary>
         public const string DefaultQueueSize = "DefaultQueueSize";
+        
+        /// <summary>
+        /// Key for the default data change filter for monitored items.
+        /// </summary>
+        public const string DefaultDataChangeTrigger = "DefaulDataChangeTrigger";
```
alxy, line 83:
```suggestion
        public const string DefaultDataChangeTrigger = "DefaultDataChangeTrigger";
```

PR #1893, opened 2022-11-08
Add --bi / --batchtriggerinterval command line to define interval in milliseconds.

marcschier:
…milliseconds.


modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
```diff
@@ -192,7 +192,9 @@ public class StandaloneCliOptions : Dictionary<string, string>, IAgentConfigProv
                     { "it|issuercertstoretype=", "Legacy - do not use.", b => {legacyOptions.Add("it|issuercertstoretype"); } },
                     { $"bs|batchsize=|{StandaloneCliConfigKeys.BatchSize}=", "The size of message batching buffer.",
                         (int i) => this[StandaloneCliConfigKeys.BatchSize] = i.ToString() },
-                    { $"si|iothubsendinterval=|{StandaloneCliConfigKeys.BatchTriggerInterval}=", "The trigger batching interval in seconds.",
+                    { $"bi|batchtriggerinterval=|{StandaloneCliConfigKeys.BatchTriggerInterval}=", "The trigger batching interval in milliseconds.",
```
cristipogacean, line 198: I think the real problem here is the bug that BatchTriggerInterval is not able to take a TimeSpan as argument's value as described in our documentation. I would fix this bug rather than adding a new argument that uses integer milliseconds. the milliseconds issue would be solved by passing something like {00:00:00.100} then.

marcschier: I would say nothing prevents you from providing the BTI environment variable as timespan string, but most other durations in the command line options are in milliseconds, e.g., publishing interval, etc., and it is just easier to write --bi=100 rather than --bi={00:00:00.100}, the formatting you would need to look up in .net, lots of things you can get wrong there...

cristipogacean (approved):

PR #1894, opened 2022-11-09
handle session connecting when complex types dictionary loading takes long time

cristipogacean:
Ensure the session event handlers are not set until the loading of the complex types dictionaries are loaded. This way we avoid the session going into erroneous state triggered by the events while the session is still basically not ready.


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Services/DefaultSessionManager.cs
```diff
@@ -661,28 +661,31 @@ public class DefaultSessionManager : ISessionManager, IDisposable {
                     _applicationConfiguration, configuredEndpoint,
                     true, sessionName, _clientConfig.DefaultSessionTimeout,
                     userIdentity, null).ConfigureAwait(false);
-                session.Handle = wrapper;
```
marcschier, line 664: Is this set somewhere else?  Looks like below the Handle must be set to the wrapper, no?

cristipogacean: yes, it is set in the HandleInitAsync(). for some reason this was set twice.

marcschier (approved):

PR #1907, opened 2022-11-21
Ad hoc discovery fails with "capacity must be non negative" error found during bug bash.

marcschier:
Engineering tool passes negative values as defaults when running ad hoc discovery. Use Clone method to sanitize them to null like the rest of the code demands.

### DIFF
--- a/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Registry/Extensions/DiscoveryConfigModelEx.cs
+++ b/components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Registry/Extensions/DiscoveryConfigModelEx.cs
@@ -4,6 +4,7 @@
 // ------------------------------------------------------------
 
 namespace Microsoft.Azure.IIoT.OpcUa.Registry.Models {
+    using System;
     using System.Linq;
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Registry/Extensions/DiscoveryConfigModelEx.cs
```diff
@@ -22,16 +23,26 @@ public static class DiscoveryConfigModelEx {
...
+                IdleTimeBetweenScans = model.IdleTimeBetweenScans < TimeSpan.Zero ?
+                    null : model.IdleTimeBetweenScans,
+                MaxNetworkProbes = model.MaxNetworkProbes <= 0 ?
+                    null : model.MaxNetworkProbes,
+                MaxPortProbes = model.MaxPortProbes <= 0 ?
+                    null : model.MaxPortProbes,
+                MinPortProbesPercent = model.MinPortProbesPercent <= 0 ?
+                    null : model.MinPortProbesPercent,
+                NetworkProbeTimeout = model.NetworkProbeTimeout <= TimeSpan.Zero ?
+                    null : model.IdleTimeBetweenScans,
```
karok2m: Shouldn't we use `model.NetworkProbeTimeout` here if its value is greater than zero ?

marcschier: Oops, thanks. :-)

cristipogacean (approved):

karok2m (approved):

karok2m:
@marcschier, this change should be applied on `main` as well, right ?

PR #1920, opened 2022-12-06
Deadband configuration support

marcschier:


components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesJobConverter.cs
```diff
@@ -146,6 +146,8 @@ public class PublishedNodesJobConverter {
                                         // ToDo: Implement mechanism for SkipFirst.
                                         SkipFirst = node.Item2.SkipFirst,
                                         DataChangeTrigger = node.Item2.DataChangeTrigger,
+                                        DeadbandValue = node.Item2.DeadbandValue,
+                                        DeadbandType = node.Item2.DeadbandType
```
cristipogacean, line 150: nit: ending comma missing

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Storage/PublishedNodesJobConverter.cs
```diff
@@ -357,8 +359,10 @@ private class OpcNodeModelComparer : IEqualityComparer<(string, OpcNodeModel)> {
                             QueueSize = node.QueueSize ?? standaloneCliModel.DefaultQueueSize,
                             // ToDo: Implement mechanism for SkipFirst.
                             SkipFirst = node.SkipFirst ?? standaloneCliModel.DefaultSkipFirst,
-                            EventFilter = node.EventFilter,
-                            DataChangeTrigger = node.DataChangeTrigger
+                            DataChangeTrigger = node.DataChangeTrigger ?? standaloneCliModel.DefaultDataChangeTrigger,
+                            DeadbandType = node.DeadbandType,
+                            DeadbandValue = node.DeadbandValue,
+                            EventFilter = node.EventFilter
```
cristipogacean, line 365: nit: ending comma

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Extensions/StackTypesEx.cs
```diff
@@ -429,8 +430,6 @@ public static class StackTypesEx {
             switch (mode) {
                 case UaDataChangeTrigger.Status:
                     return DataChangeTriggerType.Status;
-                case UaDataChangeTrigger.StatusValue:
```
cristipogacean: why did we remove this?

cristipogacean (dismissed):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Extensions/StackTypesEx.cs
marcschier: StatusValue is the UA default value and null means "default" on our nullable enum version.

cristipogacean: That's per spec correct. Nevertheless, the user should have the flexibility to explicitly set explicitly StatusValue if they want to. In case the server is not compliant or it's behavior might differ from the null default and explicit StatusValue. This was an explicit requirement from a customer.

marcschier: Ah, ok, so what you are saying is you would like a way to force applying a data change filter, correct?  Yes, that would be a way.  Then we need to change the comparer implementation and treat null different from statusvalue.

cristipogacean (approved):

PR #1924, opened 2022-12-13
Implement skip first setting to skip first notification. 

marcschier:
Message is suppressed if there are no notifications.


modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
```diff
@@ -91,10 +91,9 @@ public class StandaloneCliOptions : Dictionary<string, string>, IAgentConfigProv
                     { $"hb|heartbeatinterval=|{StandaloneCliConfigKeys.HeartbeatIntervalDefault}=", "The publisher is using this as default value in seconds " +
                         "for the heartbeat interval setting of nodes without a heartbeat interval setting.",
                         (int i) => this[StandaloneCliConfigKeys.HeartbeatIntervalDefault] = TimeSpan.FromSeconds(i).ToString() },
-                    // ToDo: Bring back once SkipFirst mechanism is implemented.
-                    //{ "sf|skipfirstevent=", "The publisher is using this as default value for the skip first " +
-                    //    "event setting of nodes without a skip first event setting.",
-                    //    (bool b) => this[StandaloneCliConfigKeys.SkipFirstDefault] = b.ToString() },
+                    { $"sf|skipfirstevent=|{StandaloneCliConfigKeys.SkipFirstDefault}=", "The publisher is using this as default value for the skip first " +
```
cristipogacean: I find `skipfirstevent` a bit confusing now that we have support for OPC UA events. I know this is 2.5 backwards compatibility material, but maybe we could break it for the sake of clarity.

marcschier: Hehehe, yes, agreed, is it possible to hide it from help?  Will investigate...

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
```
cristipogacean: is this change intended?

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
```diff
@@ -222,46 +220,46 @@ public class StandaloneCliOptions : Dictionary<string, string>, IAgentConfigProv
...
                     { $"ri|enableroutinginfo=|{StandaloneCliConfigKeys.EnableRoutingInfo}=", "Enable adding routing info to telemetry. By default this is disabled.",
                         (bool b) => this[StandaloneCliConfigKeys.EnableRoutingInfo] = b.ToString() },
-                    { $"{StandaloneCliConfigKeys.UseReversibleEncoding}=", "Use reversible encoding in JSON encoders. Default is false.",
+                    { $"re|reversibleencoding|{StandaloneCliConfigKeys.UseReversibleEncoding}=", "Use reversible encoding in JSON encoders. Default is false.",
                         (bool b) => this[StandaloneCliConfigKeys.UseReversibleEncoding] = b.ToString() },
 
                     // testing purposes
                     { "sc|scaletestcount=", "The number of monitored item clones in scale tests.",
-                        (int i) => this[StandaloneCliConfigKeys.ScaleTestCount] = i.ToString() },
+                        (string i) => this[StandaloneCliConfigKeys.ScaleTestCount] = i.ToString(), true },
```
cristipogacean: what is the intention with this change? I think this was a numeric, not a bool argument

cristipogacean (dismissed):

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
marcschier: Intention is to hide the setting from the -h output.  the .ToString() before would convert to string, so the new line should remove the .ToString() since it already is a string.  Will fix this up in another PR.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Properties/launchSettings.json
marcschier: Removed, sorry snuck in.

cristipogacean (approved):

PR #1928, opened 2022-12-20
Send UA Metadata messages over PubSub

### LINKED ISSUE #1680: Support OPC UA PubSub metadata messages in OPC Publisher
Implement ua-metadata messages as defined in 7.2.3.4.2 of the OPC UA spec (see https://reference.opcfoundation.org/v104/Core/docs/Part14/7.2.3/). These should be sent before sending the corresponding ua-data (telemetry) message and sent in regular intervals to allow OPC UA PubSub subscribers to see them if the Broker doesn't support retaining messages. they should also be sent to a separate Broker topic, if possible, for easier message management in the Broker. For a reference implementation, see here: https://github.com/barnstee/UA-CloudPublisher/blob/94a22b29f345a97948dae5b443ac4603e201c2a9/PubSubTelemetryEncoder.cs#L47

marcschier:
* Collect metadata when configuration changes and version metadata using timestamp
* DataSetClass and DataSetFieldId configuration to metadata and dataset messages together using unique ids.
* Enable sending metadata updates periodically based on CLI configuration
* FIFO/LIFO server queue configuration using CLI, PN.json and DM API (DiscardNew)
* Enable sending keyframe based on CLI, PN.json and DM API settings.
* Remove dead and unnecessary code.


api/src/Microsoft.Azure.IIoT.Api/src/Publisher/Models/DataSetContentMask.cs
```diff
@@ -39,7 +39,7 @@ public enum DataSetContentMask {
         Status = 8,
 
         /// <summary>
-        /// Dataset writer id (json)
+        /// Dataset writer name (json)
```
marcschier: Undo  #Resolved

mregen:
>     public enum DataSetContentMask {

how come this definition is different than the spec? [https://reference.opcfoundation.org/Core/Part14/v105/docs/6.3.2.3.1](https://reference.opcfoundation.org/Core/Part14/v105/docs/6.3.2.3.1)

---
In reply to: [1383728797](https://github.com/Azure/Industrial-IoT/pull/1928#issuecomment-1383728797) [](http://example.com/codeflow?ancestors=1383728797)

---
In reply to: [1383728797](https://github.com/Azure/Industrial-IoT/pull/1928#issuecomment-1383728797) [](http://example.com/codeflow?ancestors=1383728797)

---
Refers to: api/src/Microsoft.Azure.IIoT.Api/src/Publisher/Models/DataSetContentMask.cs:15 in 59f97fd. [](commit_id = 59f97fd632145f47ee1e0613151f8f36bbadd4b2, deletion_comment = False)

marcschier:
>     public enum DataSetContentMask {

These and all the other models are close to the specification but not using the specification.  That is debateable, but the reason is historic. We cannot change at this point as these values are reflected in cosmosdb documents.  That said, the OPC UA masks make only sense in the context of the network message binding and user must understand which bits to set in which binding.  Something also debateable from a user experience.

---
In reply to: [1383728797](https://github.com/Azure/Industrial-IoT/pull/1928#issuecomment-1383728797) [](http://example.com/codeflow?ancestors=1383728797,1383728797)

---
Refers to: api/src/Microsoft.Azure.IIoT.Api/src/Publisher/Models/DataSetContentMask.cs:15 in 59f97fd. [](commit_id = 59f97fd632145f47ee1e0613151f8f36bbadd4b2, deletion_comment = False)

marcschier:
>             return new PublishedDataSetSettingsModel {

Undo

---
In reply to: [1384002800](https://github.com/Azure/Industrial-IoT/pull/1928#issuecomment-1384002800) [](http://example.com/codeflow?ancestors=1384002800)

---
Refers to: components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Extensions/PublishedDataSetSettingsModelEx.cs:21 in 59f97fd. [](commit_id = 59f97fd632145f47ee1e0613151f8f36bbadd4b2, deletion_comment = False)

api/src/Microsoft.Azure.IIoT.Api/src/Publisher/Extensions/ModelExtensions.cs
```diff
@@ -776,12 +719,11 @@ public static class ModelExtensions {
                 return null;
             }
             return new DataSetWriterApiModel {
-                DataSetWriterId = model.DataSetWriterId,
+                DataSetWriterName = model.DataSetWriterName,
```
cristipogacean, line 722: Do we drop the support for DataSetWriterId with this?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/StandaloneJobOrchestrator.cs
```diff
@@ -491,9 +491,9 @@ IJsonSerializer jsonSerializer
             job.WriterGroup.DataSetWriters.ForEach(d => {
                 d.DataSet.ExtensionFields ??= new Dictionary<string, string>();
                 d.DataSet.ExtensionFields["PublisherId"] = jobId;
-                d.DataSet.ExtensionFields["DataSetWriterId"] = d.DataSetWriterId;
+                d.DataSet.ExtensionFields["DataSetWriterId"] = d.DataSetWriterName;
```
cristipogacean: would it make sense to rename DataSetWriterId to DataSetWriterName overall? to avoid confusions?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Extensions/DataSetWriterModelEx.cs
```diff
@@ -40,11 +40,11 @@ public static class DataSetWriterModelEx {
             }
             var model = new SubscriptionModel {
                 Connection = dataSetWriter.DataSet.DataSetSource.Connection.Clone(),
-                Id = dataSetWriter.DataSetWriterId,
+                Id = dataSetWriter.DataSetWriterName,
                 MonitoredItems = monitoredItems,
                 ExtensionFields = dataSetWriter.DataSet.ExtensionFields,
                 Configuration = dataSetWriter.DataSet.DataSetSource
-                    .ToSubscriptionConfigurationModel()
+                    .ToSubscriptionConfigurationModel(dataSetWriter.DataSet.DataSetMetaData)
```
cristipogacean, line 47: do we need to check if DataSet is null here?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Extensions/PublishedDataSetEventModelEx.cs
```diff
@@ -52,7 +52,7 @@ public static class PublishedDataSetEventModelEx {
                 AttributeId = null,
                 IndexRange = null,
                 SamplingInterval = TimeSpan.Zero,
-                PendingAlarms = publishedEvent.PendingAlarms?.Clone() ?? null,
+                ConditionHandling = publishedEvent.ConditionHandling.Clone(),
```
cristipogacean, line 55: would it be possible that ConditionHandling is null here?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Models/StandaloneCliModel.cs
```diff
@@ -120,29 +163,33 @@ public class StandaloneCliModel {
...
-        public bool LegacyCompatibility { get; set; } = false;
+        public bool LegacyCompatibility { get; set; }
 
         /// <summary>
         /// Configuration flag for enabling/disabling runtime state reporting.
         /// </summary>
-        public bool EnableRuntimeStateReporting { get; set; } = false;
+        public bool EnableRuntimeStateReporting { get; set; }
+
+
```
cristipogacean: remove empty line

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Extensions/StackTypesEx.cs
```diff
@@ -637,26 +563,30 @@ public static class StackTypesEx {
...
         /// <param name="mask"></param>
+        /// <param name="fieldMask"></param>
         /// <param name="encoding"></param>
         /// <returns></returns>
-        public static uint ToStackType(this DataSetContentMask? mask, MessageEncoding? encoding) {
+        public static uint ToStackType(this DataSetContentMask? mask,
+            DataSetFieldContentMask? fieldMask, MessageEncoding? encoding) {
             if (mask == null) {
                 mask =
                     DataSetContentMask.DataSetWriterId |
```
cristipogacean, line 573: does this mean we encode both DataSetWriterId and DataSetWriterName? I thought we only use the latter.

api/src/Microsoft.Azure.IIoT.Api/src/Publisher/Extensions/ModelExtensions.cs
marcschier: No, default is DataSetWriterId "string" in the encoder/decoder, only if --strict is specified it is compliant.  Also, models still read and write from the "dataSetWriterId" json property to maintain backcompat.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/StandaloneJobOrchestrator.cs
marcschier: I think we should eventually. I want to make sure we do not break backcompat though.  Let me take an action item to follow up.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Extensions/DataSetWriterModelEx.cs
marcschier: Good one. I will add a Debug.Assert or null check.  But the above code is old and so assumption is that it is never null here.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Extensions/PublishedDataSetEventModelEx.cs
marcschier: Clone() is an extension method and checks for null, and in that case returns null.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Extensions/DataSetWriterModelEx.cs
marcschier: Checked on line 27.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Extensions/StackTypesEx.cs
marcschier: Yes, both are encoded, but we can update the Json MessageProfile(s) to remove the DataSetWriterId flag.  Then we need to also update the encoder to treat DataSetWriterName flag as DataSetWriterId.  Let me think about this a bit.  What should we do?

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Services/PublisherJobService.cs
```diff
@@ -349,24 +341,18 @@ public sealed class PublisherJobService : IPublishServices<string> {
             List<PublishedDataSetVariableModel> variables = null;
             foreach (var writer in publishJob.WriterGroup.DataSetWriters) {
                 if (writer.DataSet.DataSetSource.Connection.IsSameAs(connection) &&
-                    writer.DataSetWriterId == dataSetWriterId) {
+                    writer.DataSetWriterName == dataSetWriterId) {
```
cristipogacean: `writer.DataSetWriterName == dataSetWriterId` is a bit confusing

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/DataSetContentMask.cs
```diff
@@ -50,6 +50,36 @@ public enum DataSetContentMask {
...
+        /// Default non-reversible json
+        /// </summary>
+        DefaultJson = MessageType | DataSetWriterName |
+            SequenceNumber | DataSetWriterId |
+            Timestamp | MetaDataVersion | Status,
+
+        /// <summary>
+        /// Default Uadp
+        /// </summary>
+        DefaultUadp = MinorVersion | SequenceNumber |
```
cristipogacean, line 80: `MinorVersion` seem to be duplicate. 
I'm not sure I understand why DefaultUdp is different than the DefaultJson.

docs/modules/telemetry-messages-format.md
```diff
@@ -67,6 +98,57 @@ The following messages are emitted for data value changes in a subscription if P
...
+      "MessageId": "0",
+      "MessageType": "ua-metadata",
+      "PublisherId": "opc.tcp://localhost:57537/UA/SampleServer_A2425855",
+      "DataSetWriterId": 1,
+      "MetaData": {
+        "Namespaces": [
+          "http://opcfoundation.org/UA/",
+          "urn:localhost:OPCFoundation:CoreSampleServer",
+          "http://test.org/UA/Data/",
+          ..l
```
cristipogacean: is this intended `..l`?

cristipogacean (dismissed):

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Extensions/StackTypesEx.cs
marcschier: Ok, current encoder only writes the DataSetWriterName when in --strict mode. It then also writes the DataSetWriterId value.  When not in --strict mode, only DataSetWriterId with string is written.  I think that is ok, it provides a good level of backwards compatibility.  At the expends of having DataSetWriterId in --strict mode.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Services/PublisherJobService.cs
marcschier: Will change the variable name.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa/src/Publisher/Models/DataSetContentMask.cs
marcschier: There are different flags valid for UADP vs. Json.  E.g., UADP allows encoding just major or minor, while JSON requires both.

mregen:
>           "urn:mschier205:OPCFoundation:CoreSampleServer",

personal information

---
Refers to: docs/modules/telemetry-events-format.md:71 in ed3aaf4. [](commit_id = ed3aaf4f3ea5f4d34249bd35d2a1240a862183d4, deletion_comment = False)

mregen:
>           "DataAccess",

is this really a namespace?

---
Refers to: docs/modules/telemetry-events-format.md:84 in ed3aaf4. [](commit_id = ed3aaf4f3ea5f4d34249bd35d2a1240a862183d4, deletion_comment = False)

mregen:
>               "Type": "ByteString",

not a compliant type info. should be a number? or NodeId?

---
Refers to: docs/modules/telemetry-events-format.md:229 in ed3aaf4. [](commit_id = ed3aaf4f3ea5f4d34249bd35d2a1240a862183d4, deletion_comment = False)

mregen:
>                 "TypeId": "http://opcfoundation.org/SimpleEvents#i=183",

use nsu= with ; as separator?

---
Refers to: docs/modules/telemetry-events-format.md:246 in ed3aaf4. [](commit_id = ed3aaf4f3ea5f4d34249bd35d2a1240a862183d4, deletion_comment = False)

mregen:
>           "MessageType": "ua-condition",

ua-event?

---
Refers to: docs/modules/telemetry-events-format.md:291 in ed3aaf4. [](commit_id = ed3aaf4f3ea5f4d34249bd35d2a1240a862183d4, deletion_comment = False)

mregen:
>           "DataSetWriterId": "SIMPLE-EVENTS",

Id is a number?

---
In reply to: [1385219386](https://github.com/Azure/Industrial-IoT/pull/1928#issuecomment-1385219386) [](http://example.com/codeflow?ancestors=1385219386) #Resolved

---
Refers to: docs/modules/telemetry-events-format.md:25 in ed3aaf4. [](commit_id = ed3aaf4f3ea5f4d34249bd35d2a1240a862183d4, deletion_comment = False)

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Publisher/src/Services/PublisherJobService.cs
marcschier: Updated

docs/modules/telemetry-messages-format.md
marcschier: No, good one.

components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Edge.Publisher/src/Engine/StandaloneJobOrchestrator.cs
marcschier: I moved the filling of extra extension fields into the encoder.  Need to still support extension fields in PubSub messages and configuration from control plane, will put on backlog.

marcschier:
>                 "TypeId": "http://opcfoundation.org/SimpleEvents#i=183",

I will address the docs in another PR.  There will be a --strict version of the docs, and backcompat version (this one).

---
In reply to: [1385128952](https://github.com/Azure/Industrial-IoT/pull/1928#issuecomment-1385128952) [](http://example.com/codeflow?ancestors=1385128952,1385128952)

---
Refers to: docs/modules/telemetry-events-format.md:246 in ed3aaf4. [](commit_id = ed3aaf4f3ea5f4d34249bd35d2a1240a862183d4, deletion_comment = False)

marcschier:
>           "MessageType": "ua-condition",

Condition handling is special feature of publisher.  It is not compatible with ua-event, if you turn it on.  Since you can turn it on and off we need something in the message that shows the type of the message.  Maybe we should use keyframe though.

---
In reply to: [1385203203](https://github.com/Azure/Industrial-IoT/pull/1928#issuecomment-1385203203) [](http://example.com/codeflow?ancestors=1385203203,1385203203)

---
Refers to: docs/modules/telemetry-events-format.md:291 in ed3aaf4. [](commit_id = ed3aaf4f3ea5f4d34249bd35d2a1240a862183d4, deletion_comment = False)

marcschier:
>           "DataSetWriterId": "SIMPLE-EVENTS",

I will address the docs.  There will be a --strict version of the docs, and backcompat version (this one).

---
In reply to: [1385219386](https://github.com/Azure/Industrial-IoT/pull/1928#issuecomment-1385219386) [](http://example.com/codeflow?ancestors=1385219386,1385219386)

---
Refers to: docs/modules/telemetry-events-format.md:25 in ed3aaf4. [](commit_id = ed3aaf4f3ea5f4d34249bd35d2a1240a862183d4, deletion_comment = False)

marcschier:
>               "Type": "ByteString",

I will address the docs.  There will be a --strict version of the docs, and backcompat version (this one).

---
In reply to: [1385126686](https://github.com/Azure/Industrial-IoT/pull/1928#issuecomment-1385126686) [](http://example.com/codeflow?ancestors=1385126686,1385126686)

---
Refers to: docs/modules/telemetry-events-format.md:229 in ed3aaf4. [](commit_id = ed3aaf4f3ea5f4d34249bd35d2a1240a862183d4, deletion_comment = False)

cristipogacean (approved):

mregen:
>                         // is set must be fixed.

bug in UA stack?

---
Refers to: components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Stack/PubSub/UadpDiscoveryMessage.cs:122 in ed3aaf4. [](commit_id = ed3aaf4f3ea5f4d34249bd35d2a1240a862183d4, deletion_comment = False)

marcschier:
> // is set must be fixed

@mregen, yes, BinaryEncoder disposes the underlying stream even if leaveOpen is set to true in constructor.  I was planning to address this, but if you are in the code already, greatly appreciated.

mregen:
>                         // is set must be fixed.

yes, dispose always disposes the stream itself

---
In reply to: [1385355738](https://github.com/Azure/Industrial-IoT/pull/1928#issuecomment-1385355738) [](http://example.com/codeflow?ancestors=1385355738)

---
Refers to: components/opc-ua/src/Microsoft.Azure.IIoT.OpcUa.Protocol/src/Stack/PubSub/UadpDiscoveryMessage.cs:122 in ed3aaf4. [](commit_id = ed3aaf4f3ea5f4d34249bd35d2a1240a862183d4, deletion_comment = False)

PR #1944, opened 2023-01-22
Enable ability to specify a topic name for metadata messages

### LINKED ISSUE #1943: Ability to configure metadata queue name through commandline
**Is your feature request related to a problem? Please describe.**
Customers want to send metadata to a dedicated topic if connecting to an MQTT broker.  Today it goes to the telemetry path.  The suggestion in Part 14 is to use a subtopic, by default $metadata.  

**Describe the solution you'd like**
We want to be able to specify the name of the subtopic to use using a command line argument like --mdo --metadataoutput.  On MQTT the metadata should be send with RETAIN and ttl as described in the spec to a subtopic of the telemetry topic with the provided name.  With EdgeHub the output name should be used, or if --ri/--enableroutinginfo is enabled, <output>/<metdataoutput>.

**Describe alternatives you've considered**
A full topic path could be used, but it would apply to all telemetry paths, unless we also make it templated.  It would only apply to MQTT, while we have seen customers leveraging routing for custom handling of messages, which might be useful to do just for metadata.  We could also provide the topic/output in pn.json, but this would be difficult to configure, and likely always the same subtopic anyway, command line is far simpler.

Open questions: Should we add outputname support to mqtt adapter templating support and by default remove the [outputname} and remove "duplicate path separators" if no output name is provided?   Separate feature request?

marcschier:
* Allow configuration of a metadata queue name which will become an output name with IoT Hub routing and a sub topic in mqtt mode.
* Send metadata messages as RETAIN and with TTL (v5 only)
* Update OPC UA .net packages and enable full metadata support
* Integration tests for MQTT mode with configuration and telemetry / metadata queue name tests.
* Update MQTT.net to version 4.
* Support method calls through MQTT v5 and v311 (latter using IoT Hub request id mode and fixed topic path).
* Update existing tests.
* Break IClient interface into smaller items for ability to inject some parts multiple times (future)


api/src/Microsoft.Azure.IIoT.Api/src/Publisher/Clients/PublisherModuleControlClient.cs
```diff
@@ -123,9 +123,6 @@ public sealed class PublisherModuleControlClient : IPublisherControlApi {
         /// <returns></returns>
         private async Task<R> CallServiceOnPublisherAsync<T, R>(string service,
             string deviceId, string moduleId, T request) {
-            if (string.IsNullOrEmpty(deviceId)) {
```
marcschier, line 126: Removed to allow null device id for MQTT RPC (see tests). The actual transport implementation will throw if what it needs is not provided.

api/src/Microsoft.Azure.IIoT.Api/src/Publisher/Extensions/ModelExtensions.cs
```diff
@@ -880,7 +881,6 @@ public static class ModelExtensions {
                 DiagnosticsInterval = model.DiagnosticsInterval,
                 MaxOutgressMessages = model.MaxOutgressMessages,
                 UseStandardsCompliantEncoding = model.UseStandardsCompliantEncoding,
-                EnableRoutingInfo = model.EnableRoutingInfo,
```
marcschier, line 883: Replaced with "EnableOutputRouting". $$RoutingInfo is now handled in IOT SDK adapter.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Models/PublisherExtensions.cs
```diff
@@ -41,7 +41,7 @@ public static class PublisherExtensions {
                 DataSetClassId = model.DataSetClassId,
                 DataSetDescription = model.DataSetDescription,
                 DataSetKeyFrameCount = model.DataSetKeyFrameCount,
-                DataSetMetaDataSendInterval = model.DataSetMetaDataSendInterval,
+                MetaDataUpdateTime = model.DataSetMetaDataSendInterval,
```
cristipogacean: I'm not sure if we're missing the `MetaDataQueueName` here?

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Models/PublisherExtensions.cs
```diff
@@ -127,7 +127,7 @@ public static class PublisherExtensions {
                 DataSetWriterGroup = endpoint.DataSetWriterGroup,
                 DataSetDescription = endpoint.DataSetDescription,
                 DataSetKeyFrameCount = endpoint.DataSetKeyFrameCount,
-                DataSetMetaDataSendInterval = endpoint.DataSetMetaDataSendInterval,
+                DataSetMetaDataSendInterval = endpoint.MetaDataUpdateTime,
```
cristipogacean: I'm not sure if we're missing the `MetaDataQueueName` here?

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
```diff
@@ -123,18 +124,12 @@ public class StandaloneCliOptions : Dictionary<string, string>, IAgentConfigProv
...
                     "Disables sending any metadata when metadata version changes. This setting can be used to also override the messaging profile's default support for metadata sending. \nDefault: `False` if the messaging profile selected supports sending metadata, `True` otherwise.\n",
                     (bool? b) => this[StandaloneCliConfigKeys.DisableDataSetMetaData] = b?.ToString() ?? "True" },
-
-                // TODO: Default metadata output name
-
-                { $"ri|enableroutinginfo:|{StandaloneCliConfigKeys.EnableRoutingInfo}:",
-                    "Add the routing info to telemetry messages. The name of the property is `$$RoutingInfo` and the value is the `DataSetWriterGroup` for that particular message.\nWhen the `DataSetWriterGroup` is not configured, the `$$RoutingInfo` property will not be added to the message even if this argument is set.\nDefault: `False` (disabled).\n",
-                    (bool? b) => this[StandaloneCliConfigKeys.EnableRoutingInfo] = b?.ToString() ?? "True" },
                     { $"lc|legacycompatibility=|{StandaloneCliConfigKeys.LegacyCompatibility}=",
```
cristipogacean, line 133: I'm not sure if we're missing something like Default `MetaDataQueueName` here?

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
```diff
@@ -206,9 +207,11 @@ public class StandaloneCliOptions : Dictionary<string, string>, IAgentConfigProv
                 "---------------------------",
                 "",
 
-                { $"aa|autoaccept:|{StandaloneCliConfigKeys.AutoAcceptCerts}:",
-                    "The publisher trusts all servers it is establishing a connection to. WARNING: This setting should never be used in production environments!\n",
+                { $"aa|acceptuntrusted:|{StandaloneCliConfigKeys.AutoAcceptCerts}:",
```
cristipogacean, line 210: that looks like a breaking change to me. I think we should check if the Twin module has the same argument here to have consistency.

cristipogacean (dismissed):

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Models/PublisherExtensions.cs
marcschier: Totally correct.  Will fix.

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Models/PublisherExtensions.cs
marcschier: Totally correct.  Will fix

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
marcschier: Below in transport settings:  { $"mqn|metadataqueuename:|{StandaloneCliConfigKeys.DefaultDataSetMetaDataQueueName}:",

modules/src/Microsoft.Azure.IIoT.Modules.OpcUa.Publisher/src/Runtime/StandaloneCliOptions.cs
marcschier: autoaccept still works, it is just hidden.  Should we merge in twin then we get this for free.

cristipogacean (dismissed):

PR #1956, opened 2023-03-07
Fix issue crashing in diagnostics handler

marcschier:


docs/release-announcement.md
```diff
@@ -6,33 +6,45 @@ We are pleased to announce the release of version 2.8.4 of our Industrial IoT Pl
 
 ### IMPORTANT - PLEASE READ
```
cristipogacean, line 17: should we update this to 2.8.5? and have a separate chapter 2.8.4?

cristipogacean (dismissed):

docs/release-announcement.md
marcschier: Good catch.

cristipogacean (approved):

PR #1965, opened 2023-03-24
Avoid key-already-exists-exception in the Trigger logic + some async enhancements

dimtix:
1. In the currentlyMonitored items list can be more than one item with the same StartNodeId. E.g. by simultaneous monitoring of the events and coditions at the same node (two OpcNodes-configs with the same Id in pn.json in the same datasetWriter). Trigger logic was changed to avoid key-already-exists-exception in this case - only the first item will be set as trigger in this case.

3. The trigger logic do not needed, if threre are not Triggers configured at all.


marcschier:
Integration test is failing in several runs.

marcschier:
We will remove Trigger configuration in preview 2 as it never really worked or was tested.  This should address the crash on duplicate startnodeid. 

Is triggering required? 

(And thanks for all the great pull requests!)

dimtix:
Triggering is not required. But it craches at the moment without proposed changes.

PR #1970, opened 2023-04-24
Update publisher.md

eddraper-msft:
This was simplified into this working configuration earlier today


marcschier:
Already fixed in mainline

PR #2034, opened 2023-08-22
Update to latest nuget packages

marcschier:


src/Azure.IIoT.OpcUa/src/Encoders/JsonDecoderEx.cs
```diff
@@ -67,6 +67,12 @@ public void Dispose()
             }
         }
 
+        /// <inheritdoc/>
+        public void Close()
```
mregen, line 71: isn't this int ?

mregen: ok, for the encoder...

mregen (approved):

Issue #2109, opened 2023-11-15
Redis Dapr sample


xky0007:
**Describe the bug**
When using dapr with redis, it cannot publish data out.

**To Reproduce**
1. Set opc-publisher-service in docker compose file  
``` services:  
  opc-publisher-test:
    container_name: "opc-publisher-test"
    image: mcr.microsoft.com/iotedge/opc-publisher:2.9.2
    environment:
    - PublishedNodesFile=/appdata/publisher/pn.json
    - DaprConnectionString="PubSubComponent=redis-pubsub-cloud;HttpPort=7001;"
    - DefaultTransport=Dapr
    - MessagingMode=PubSub
    - LogLevel=Debug

    network_mode: "host"
    volumes:
    - ./data/opc-publisher-1:/appdata/publisher
    
publisher-dapr:
    image: "daprio/daprd:edge"
    command: [
      "./daprd",
      "-app-id","opc-publisher-test",
      #"-placement-host-address", "localhost:50006",
      "-components-path","/components",
      #"--app-protocol","http",
      "--dapr-http-port","7001",
      #"--app-port","7000"
    ]
    network_mode: "host"

    volumes:
      - "./dapr:/components"
    depends_on:
      - opc-publisher-test
```
2. dapr configuration
```
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: redis-pubsub-cloud
  namespace: default
spec:
  type: pubsub.redis
  version: v1
  metadata:
  - name: redisHost
    value: localhost:6379
  - name: redisPassword
    value: ""
```

**Expected behavior**

**Screenshots**
![image](https://github.com/Azure/Industrial-IoT/assets/30423949/f957ac89-8a5f-4a64-8dea-c80404b0a8c8)

```
opc-publisher-test                           | [08:45:03.7398] dbug: Azure.IIoT.OpcUa.Publisher.Stack.Services.OpcUaSubscription[0]
opc-publisher-test                           |       Keep alive for subscription opc.tcp://localhost:4844_346F0688_TCN-A1234J4321:TCN-A1234J4321-data:10 with sequenceNumber 2, publishTime 11/15/2023 08:45:03.
opc-publisher-test                           | [08:45:09.4540] dbug: Azure.IIoT.OpcUa.Publisher.Services.NetworkMessageSink[0]
[...]

PR #2115, opened 2023-11-18
Fix dapr configuration checks

marcschier:
Must check against empty or null string of pubsubname and storename before enabling clientts. Set values to null instead of empty string to be explicit.


marcschier:
#2109

Issue #2119, opened 2023-11-20
Dapr should work in http instead of https


xky0007:
**Describe the bug**

In the source code `internal sealed class Dapr : ConfigureOptionBase<DaprOptions>`

```
 // Permit the port to be set if provided, otherwise use defaults.
 if (properties.TryGetValue(GrpcPortKey, out var value) &&
     int.TryParse(value, CultureInfo.InvariantCulture, out var port))
 {
     options.GrpcEndpoint = "https://localhost:" + port;
 }

 if (properties.TryGetValue(HttpPortKey, out value) &&
     int.TryParse(value, CultureInfo.InvariantCulture, out port))
 {
     options.HttpEndpoint = "https://localhost:" + port;
 }
```

From the dapr API [doc](https://docs.dapr.io/reference/api/pubsub_api/), it should use http instead of https. And after I changed https to http, #2109 is also solved. (Also need to change to GRPC protocol #2118 )

Thanks.

marcschier:
Many thanks for the contributions, more are certainly welcome in this area in particular. I added a "Scheme" key to the connection string to select http or https, the default is now "http".  preview4 container image will be available later today.  Keep it coming :-)

PR #2121, opened 2023-11-21
Update Configuration.cs

xky0007:
Fix dapr should use http instead of https.

#2119


xky0007:
I saw there is already a pull request contains the problem. Closed.

PR #2128, opened 2023-12-02
Diagnostic notification logging filter

marcschier:


mregen (dismissed):

mregen (dismissed):
:shipit:

src/Azure.IIoT.OpcUa.Publisher/src/Services/RuntimeStateReporter.cs
```diff
@@ -269,8 +269,18 @@ private async Task UpdateApiKeyAndCertificateAsync()
...
+                var san = new SubjectAlternativeNameBuilder();
+                san.AddDnsName(dnsName);
+                await AddAddressesAsync(dnsName, san).ConfigureAwait(false);
 
+                var altDns = _identity?.ModuleId ?? _identity?.DeviceId;
+                if (!string.IsNullOrEmpty(altDns))
+                {
+                    san.AddDnsName(altDns);
+                    await AddAddressesAsync(altDns, san).ConfigureAwait(false);
+                }
```
mregen, line 293: I would also check the dns names on existing certs to see if they still match, unless it is safe to assume they are always the same

src/Azure.IIoT.OpcUa.Publisher/src/Services/RuntimeStateReporter.cs
```diff
@@ -289,6 +299,19 @@ private async Task UpdateApiKeyAndCertificateAsync()
                 "Stored new Certificate in {Store} store (and scheduled renewal after {Duration}).",
                 apiKeyStore.Name, renewalDuration);
             _certificateRenewals++;
+
+            async Task AddAddressesAsync(string hostName, SubjectAlternativeNameBuilder san)
+            {
+                try
+                {
+                    var addresses = await Dns.GetHostAddressesAsync(hostName).ConfigureAwait(false);
+                    addresses.ForEach(san.AddIpAddress);
```
mregen: >san.AddIpAddress [](http://example.com/codeflow?start=38&length=16)

does it filter duplicates?

src/Azure.IIoT.OpcUa.Publisher/src/Services/RuntimeStateReporter.cs
```diff
@@ -289,6 +299,19 @@ private async Task UpdateApiKeyAndCertificateAsync()
                 "Stored new Certificate in {Store} store (and scheduled renewal after {Duration}).",
                 apiKeyStore.Name, renewalDuration);
             _certificateRenewals++;
+
+            async Task AddAddressesAsync(string hostName, SubjectAlternativeNameBuilder san)
+            {
+                try
+                {
+                    var addresses = await Dns.GetHostAddressesAsync(hostName).ConfigureAwait(false);
```
mregen, line 319: >Dns.GetHostAddressesAsync(h [](http://example.com/codeflow?start=42&length=27)

does it also return Ipv6 temporary addresses, not sure if there is a need to filter

src/Azure.IIoT.OpcUa.Publisher/src/Services/RuntimeStateReporter.cs
marcschier: Good one, added filter.

src/Azure.IIoT.OpcUa.Publisher/src/Services/RuntimeStateReporter.cs
marcschier: Should it filter out all ipv6?  Or just some

mregen: if any I would only add global unicast IPv6 addresses (2000::/3) or no IPv6 at all.

mregen (dismissed):
:shipit:

PR #2352, opened 2024-10-02
Update dependencies and add response compression.

marcschier:


src/Azure.IIoT.OpcUa/src/Encoders/JsonDecoderEx.cs
```diff
@@ -92,6 +92,13 @@ public void SetMappingTables(NamespaceTable namespaceUris, StringTable serverUri
             // No op
         }
 
+        /// <inheritdoc/>
+        public IEncodeable DecodeMessage(Type expectedType)
+        {
+            throw ServiceResultException.Create(StatusCodes.BadNotSupported,
+                "Not supported in this decoder.");
+        }
```
mregen, line 100: this method is quite handy for fuzzing the decoder

PR #2478, opened 2026-06-17
Fix Windows test-host crash: serialize alarm condition branch access

marcschier:
## Problem

The Windows `test_windows_Release 2` job (Module.Tests) intermittently failed with a **native access violation** that crashed the test host. WinDbg `!analyze -v` on a first-chance dump pinned the fault to `Opc.Ua.ConditionState.IsBranch` dereferencing a **null reference**, reached during alarm condition processing in the **Alarms test server**.

## Root cause

The OPC UA stack (`OPCFoundation.NetStandard.Opc.Ua.Core` 1.5.378.145) stores condition branches in a plain, **non thread-safe `Dictionary`** (`m_branches`). It is:

- **enumerated** in `ConditionState.GetRetainState` (via `UpdateRetainState`) and `ConditionState.ConditionRefresh`, and
- **mutated** in `CreateBranch` / `ReplaceBranchEvent` / `RemoveBranchEvent`

with **no synchronization**. The Alarms server runs a live simulation that drives alarm state changes on background `Timer` threads. When the branch dictionary is enumerated while a concurrent mutation is in flight, the enumeration observes a torn entry and dereferences it → native AV. This only reproduces under the slow, constrained CI Windows container.

## Fix

We instantiate the alarm condition types in `SourceState.CreateAlarm`, so we can subclass them and serialize all branch access. `GuardedAlarmStates.cs` adds guarded subclasses of the four types we create (`AlarmConditionState`, `ExclusiveDeviationAlarmState`, `NonExclusiveLevelAlarmState`, `TripAlarmState`) that:

- lock `GetRetainState` and `CreateBranch` on the **owning node manager `Lock`** — the same lock every alarm mutation in this server already takes (re-entrant, no new lock ordering), and
- **propagate that lock to branches** the stack creates via `Activator.CreateInstance(GetType(), this)`.

Contained entirely to the test server; the address space and alarm behavior are unchanged (base logic is still invoked, just serialized).

## Validation

- `dotnet build` of `Azure.IIoT.OpcUa.Publisher.Testing.Servers` and `Azure.IIoT.OpcUa.Publisher.Module.Tests` — 0 errors.
- Ran the alarm/branch tests locally (branches are exercised by "pending conditions"): **9/9 passed** across JSON / MsgPack / MQTT / device-method variants (`CanSendPendingConditions*`).

This is the 4th distinct native-AV root-caused in the Module.Tests Windows job; prior fixes: #2469, #2474, #2476.

### DIFF
--- a/src/Azure.IIoT.OpcUa.Publisher.Testing/src/Alarms/GuardedAlarmStates.cs
+++ b/src/Azure.IIoT.OpcUa.Publisher.Testing/src/Alarms/GuardedAlarmStates.cs
@@ -1,0 +1,211 @@
+// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+namespace Alarms
+{
+    using Opc.Ua;
+
+    /// <summary>
+    /// The OPC UA stack stores condition branches in a plain, non thread-safe
+    /// <see cref="System.Collections.Generic.Dictionary{TKey, TValue}"/> that is
+    /// both enumerated (<c>ConditionState.GetRetainState</c>,
+    /// <c>ConditionState.ConditionRefresh</c>) and mutated
+    /// (<c>CreateBranch</c>, <c>ReplaceBranchEvent</c>, <c>RemoveBranchEvent</c>)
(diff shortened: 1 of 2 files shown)


marcschier:
Tracking the underlying OPC UA stack thread-safety bug in #2479. This PR is the contained workaround in the Alarms test server.

Issue #2479, opened 2026-06-17
OPC UA stack: ConditionState branch dictionary is not thread-safe (native AV under concurrent access)


marcschier:
## Summary

The OPC UA stack (`OPCFoundation.NetStandard.Opc.Ua.Core` 1.5.378.145) accesses a condition's branch collection (`ConditionState.m_branches`, a plain `Dictionary<string, ConditionState>`) from multiple threads **without synchronization**:

- **Readers:** `ConditionState.GetRetainState` (via `UpdateRetainState`), `ConditionState.ConditionRefresh`, `GetBranch`
- **Writers:** `ConditionState.CreateBranch`, `ReplaceBranchEvent`, `RemoveBranchEvent`, `ClearBranches`

When a condition's state changes on a background thread while the branch dictionary is concurrently enumerated, the enumeration can observe a torn/garbage entry and dereference it, crashing the **process** with a native access violation (observed as `Opc.Ua.ConditionState.IsBranch` on a `null` `this`).

## Impact

Surfaced as intermittent **test-host crashes** in the Windows `test_windows_Release 2` CI job. The Alarms test server (`src/Azure.IIoT.OpcUa.Publisher.Testing/src/Alarms`) runs a live simulation that drives alarm state changes on `System.Threading.Timer` threads, which races with condition read paths. Only reproduces under the slow, constrained CI Windows container.

## Workaround (this repo)

PR #2478 subclasses the alarm condition types created in `SourceState.CreateAlarm` (`GuardedAlarmStates.cs`) and serializes `GetRetainState`/`CreateBranch` on the owning node manager `Lock` (the lock every alarm mutation already takes), propagating the lock to stack-created branches.

## Follow-up

- [ ] Report the branch-dictionary thread-safety issue upstream to OPC Foundation (UA-.NETStandard).
- [ ] Once the stack provides thread-safe branch handling, remove the `GuardedAlarmStates` workaround.

## References

- Workaround PR: #2478
- Related test-host crash fixes: #2469, #2474, #2476

PR #2482, opened 2026-06-18
Fix Windows test-host crash: guard alarm retain-state against null BranchId

marcschier:
## Problem

Follow-up to #2478, #2480 and #2481. Build [168527043](https://msazure.visualstudio.com/One/_build/results?buildId=168527043) — which included **all three** prior fixes — still failed `test_windows_Release 2` with the same native test-host crash.

## What the dump showed

- **#2481 worked**: the dump showed only a **single** `BaseServerFixture..ctor` (no concurrent construction), confirming server construction is now serialized.
- But the fault persisted at **exactly the same point across all five failing builds**: `Opc.Ua.ConditionState.IsBranch` dereferencing a **null `BranchId`**, reached from `AlarmConditionState.GetRetainState` during the alarm server's construction (`SourceState..ctor → Refresh → … → SetEnableState → GetRetainState`).
- The concurrent `ConditionRefreshWorkerAsync` thread was **blocked/waiting** (correctly serialized by #2480), and no other thread was an obvious heap corruptor.
- Inspecting the faulting condition: a valid object whose `BranchId` is null while `EnabledState` is populated (so the retain logic runs, then crashes on `IsBranch`). The exact cause of the transient null `BranchId` couldn't be pinned — the .NET 10 DAC won't load for SOS — but the fault point is 100% consistent.

## Fix

Guard the retain-state evaluation in the guarded alarm states (`GuardedAlarmStates.cs`): if `BranchId` or `EnabledState` is not present, treat the condition as **not retained** instead of dereferencing null and crashing the test host. In normal operation both are always present, so behavior is unchanged.

This is a targeted defensive guard at the single, consistent fault point, on top of the structural fixes in #2478/#2480/#2481. Test-only.

## Validation

- Clean build of `Testing.Servers` + `Module.Tests` (0 errors).
- `PendingConditions` (exercise retained alarm conditions): **9/9 passed** locally — guard does not change behavior.

Upstream SDK thread-safety tracked in #2479.

### DIFF
--- a/src/Azure.IIoT.OpcUa.Publisher.Testing/src/Alarms/GuardedAlarmStates.cs
+++ b/src/Azure.IIoT.OpcUa.Publisher.Testing/src/Alarms/GuardedAlarmStates.cs
@@ -30,6 +30,27 @@ namespace Alarms
     }
 
     /// <summary>
+    /// Helpers for the guarded alarm states.
+    /// </summary>
+    internal static class BranchGuard
+    {
+        /// <summary>
+        /// The stack's retain-state evaluation dereferences BranchId (via IsBranch)
+        /// and EnabledState. Under heavy CI load these have been observed to be null
+        /// on a condition that is otherwise in use, which crashes the test host with
+        /// a native access violation. Skip the evaluation in that case - an
+        /// uninitialized condition is treated as not retained - rather than crash.
+        /// In normal operation BranchId/EnabledState are always present, so behavior
+        /// is unchanged.
+        /// </summary>
+        /// <param name="condition"></param>
+        public static bool IsInitialized(ConditionState condition)
+        {
+            return condition.BranchId != null && condition.EnabledState?.Id != null;
+        }
+    }
+
+    /// <summary>
     /// Guarded <see cref="AlarmConditionState"/>.
     /// </summary>
     internal sealed class GuardedAlarmConditionState : AlarmConditionState, IBranchGuarded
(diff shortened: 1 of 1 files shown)


marcschier:
### CI-fix loop result

**Relevant failure (this branch's purpose) — FIXED.** The Windows `test_windows_Release 2` / `Publisher.Module` native test-host crash this PR targets is gone: the full suite now runs to completion (`Passed: 1457, Skipped: 79, Total: 1537`) instead of aborting on the `ConditionState.IsBranch` access violation.

**Remaining red check is a pre-existing, unrelated flake.** `MqttConfigurationIntegrationTests.CanSend*ToTopicConfiguredWithMethodAsync` fails with `Furly.Exceptions.MethodCallException: Timed out calling method \ ... Broker localhost:489xx possibly unreachable` (MQTT method-call transport). Evidence it is **not** caused by this branch:
- `main` run [27735782499](https://github.com/Azure/Industrial-IoT/actions/runs/27735782499) (which does **not** contain this PR) fails the identical test `CanSendPendingConditionsToTopicConfiguredWithMethodAsync(useMqtt5: True)` with the same broker-unreachable error on the same port 48999.
- An Alarms-condition null-guard cannot cause an MQTT broker to be unreachable.
- Re-run cleared the (separate, intermittent) ubuntu crash; the Windows MQTT failure reproduced identically to main.

The MQTT-broker reachability flakiness is a separate issue affecting `main` and should be tracked/fixed independently; it is out of scope for this Alarms test-host-crash fix.

PR #2489, opened 2026-07-07
Log direct method call failures so they surface in support bundles

marcschier:
## Summary

Investigation of a customer incident (`UnpublishAllNodes_V1` returning HTTP 500) found that when a direct method call fails, **the failure is effectively invisible in the module logs and support bundles**. Direct method invocations bypass the ASP.NET request pipeline, and the Furly router only logs their exceptions via `MethodCallError` at **`LogLevel.Trace`** — which is never enabled in production. As a result, support bundles contained no trace of the failing call or its exception, making root-cause analysis impossible.

This PR adds durable, visible logging at the exception-filter choke points so future issues in this area are captured going forward.

## Changes

- **`RouterExceptionFilterAttribute`** (direct-method path — the single choke point every direct-method exception passes through): extracted the exception→status logic into a `Map` helper and added visible failure logging. The log level is derived from the mapped status: **≥500 → Error, 408/429/412 → Warning, other 4xx → Debug**. The filter is instantiated by the router via reflection (no DI), so the module host wires up a logger once during startup. **Status mapping is unchanged** (logging is additive).
- **`Startup.Configure`**: initializes the filter logger from `ILoggerFactory`.
- **`ControllerExceptionFilterAttribute`** (HTTP path): symmetric failure logging using the per-request DI logger already available via `HttpContext.RequestServices`.
- **`ConfigurationController`**: optional injected `ILogger<ConfigurationController>`; logs each configuration mutation (`PublishNodes`, `UnpublishNodes`, `UnpublishAllNodes`, `AddOrUpdateEndpoints`) at Information with the endpoint and node count, so control-plane activity — and the context of any subsequent failure — is recorded. The constructor parameter is optional, so existing call sites are unaffected.
- **Tests** (`tests/Filters/ExceptionFilterLoggingTests.cs`): 13 tests covering router status-mapping preservation, controller-filter level selection (500→Error, 404→Debug, 408→Warning), and ConfigurationController invocation logging.

## Validation

- `dotnet build -c Release` — module source and tests build with 0 errors.
- New tests: 13/13 passed.
- Existing controller suite: 781 passed / 0 failed (27 skipped) — no regressions.

## Notes / follow-ups (not in this PR)

- The behavioral hardening for `UnpublishAllNodes` (treat null/empty/whitespace `EndpointUrl` as "purge all") is already on `main` via #2477; backporting it to `release/2.9` is recommended separately.
- Optional future enhancement: a `Meter` counter for direct-method failures to surface them in Prometheus / the periodic diagnostics block.

### DIFF
--- a/src/Azure.IIoT.OpcUa.Publisher.Module/src/Controllers/ConfigurationController.cs
+++ b/src/Azure.IIoT.OpcUa.Publisher.Module/src/Controllers/ConfigurationController.cs
@@ -42,7 +42,7 @@ namespace Azure.IIoT.OpcUa.Publisher.Module.Controllers
     [Authorize]
     [Produces(ContentMimeType.Json, ContentMimeType.MsgPack)]
     [Consumes(ContentMimeType.Json, ContentMimeType.MsgPack)]
-    public partial class ConfigurationController : ControllerBase, IMethodController
+    public class ConfigurationController : ControllerBase, IMethodController
     {
         /// <summary>
         /// Create publisher methods controller
@@ -293,11 +293,8 @@ namespace Azure.IIoT.OpcUa.Publisher.Module.Controllers
             CancellationToken ct = default)
         {
             ArgumentNullException.ThrowIfNull(request);
-            if (_logger != null)
-            {
-                ConfigurationOperationInvoked(_logger, nameof(AddOrUpdateEndpointsAsync),
-                    "<multiple>", request.Count);
-            }
+            _logger?.ConfigurationOperationInvoked(nameof(AddOrUpdateEndpointsAsync),
+                "<multiple>", request.Count);
             await _configServices.AddOrUpdateEndpointsAsync(request, ct).ConfigureAwait(false);
             return new PublishedNodesResponseModel();
         }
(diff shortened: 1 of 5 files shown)


src/Azure.IIoT.OpcUa.Publisher.Module/src/Controllers/ConfigurationController.cs
```diff
@@ -418,6 +430,30 @@ public async Task<List<PublishDiagnosticInfoModel>> GetDiagnosticInfoAsync(
...
+        private void LogEntryOperation(string operation, PublishedNodesEntryModel? request)
+        {
+            if (_logger != null)
+            {
+                ConfigurationOperationInvoked(_logger, operation,
+                    request?.EndpointUrl ?? "<none>", request?.OpcNodes?.Count ?? 0);
+            }
+        }
+
+        [LoggerMessage(EventId = 1, Level = LogLevel.Information,
```
marcschier: Use the logging class style like in all other areas in the project.  In all classes in this PR.

marcschier: Done in ca9adc40d. Moved all logging to the project's source-generated logging-class style: separate internal static partial class *Logging with an EventClass base and [LoggerMessage] 	his ILogger extension methods — applied to RouterExceptionFilterAttribute, ControllerExceptionFilterAttribute, and ConfigurationController. Build + tests pass (13/13 logging tests).

PR #2520, opened 2026-08-07
Cover the legacy updated-timestamp heartbeat behaviour

marcschier:
A customer reported that after the heartbeat watchdog fix their value stream
was complete and gap free, yet a share of `SourceTimestamp` distances was still
far from the configured sampling interval, with some distances going backwards.
They then disclosed that they run `--hbb=WatchdogLKVWithUpdatedTimestamps`.

That is material, and it is the one thing the existing coverage did not
exercise.

## Why

`WatchdogLKVWithUpdatedTimestamps` is the only behavior in the product that
*synthesizes* a source timestamp. On heartbeat it re-sends the last known value
with

```
SourceTimestamp += (timerFireTime - valueReceiveTime)
```

The shift is measured on the OPC Publisher wall clock, while the timestamp it
is added to comes from the server clock. The constant offset between the two
clocks cancels, but network latency, publish cycle phase and processing delay
do not - they land directly in the emitted `SourceTimestamp`.

The consequence is that such a heartbeat is *not* a duplicate. Every other
heartbeat behavior re-sends the original timestamp, which is what makes the
"same timestamp twice in a row" signature in the troubleshooting guide work.
Here the timestamp is unique, and the `Heartbeat` indicator only exists in the
full featured message profiles - so on `Samples` or `PubSub` a consumer cannot
distinguish the message from a real sample.

There is a second consequence that was not previously understood. The shift is
measured from the moment the value was *received*, which already lags the
timestamp the server assigned to it. A heartbeat firing shortly before the next
real value therefore carries a timestamp *beyond* the one that value will
carry, and the observed timestamp sequence moves backwards. That is customer
symptom (d) - an "old" value arriving right after a newer one - produced
without a single value being lost or reordered.

## What was measured

Against the counter server, whose source timestamps are perfectly regular by
construction, so the source contributes exactly zero jitter and everything
reported is attributable to OPC Publisher.

**Heartbeat at the sampling interval** (50 nodes, 2 s sampling, 2 s heartbeat):

| Build | Heartbeats | Shifted timestamps | Message interval violations |
| --- | --- | --- | --- |
| With the watchdog grace period (#2517) | 0 | 0 | 0 |
| Grace period removed (pre #2517) | 946 | 946 | 900 |

**Slow source, so heartbeats are guaranteed to fire** (50 nodes, counter every
10 s, 2 s heartbeat, 300 value transitions):

| Behavior | Heartbeats | Shifted timestamps | Timestamp regressions |
| --- | --- | --- | --- |
| `WatchdogLKV` (default) | 1050 | 0 | 0 |
| `WatchdogLKVWithUpdatedTimestamps` | 1049 | 1049 | 99 |

In every case the value stream itself was complete and in order: zero missing,
zero out of order, zero value interval violations. Only the timestamps moved.

## Changes

No product code changes.

- `UpdatedTimestampHeartbeatDoesNotDisplaceTimestampsAsync` runs the customer
  scenario with `--hbb=WatchdogLKVWithUpdatedTimestamps` and asserts the
  watchdog does not fire while data flows. Verified both ways - clean with the
  grace period, 946 early heartbeats without it.
- `SlowSourceHeartbeatsKeepTimestampsOrderedAsync` and
  `SlowSourceUpdatedTimestampHeartbeatsShiftTimestampsAsync` use a counter that
  changes five times slower than the heartbeat, so heartbeats are guaranteed to
  fire and the shift arithmetic actually runs. This is deterministic rather
  than load dependent - the server simply does not produce a value. They pin
  the difference between the two behaviors, and cover the separate case of a
  slowly changing counter where heartbeats are expected and wanted.
- `RunScenarioAsync` asserts the resolved `DefaultHeartbeatBehavior` whenever a
  behavior is requested. An unrecognised command line option is only warned
  about, not rejected, so without this a renamed switch would leave the default
  behavior in place and the arms would stay green while covering nothing.
- `TelemetryQualityReport.SourceTimestampRegressions` - symptom (d) measured on
  timestamps rather than on values. The existing ordering check compares
  counter values and therefore cannot trip on a repeated value no matter what
  timestamp it carries, so nothing covered this.
- `TimestampJitterDiagnosticTests` reports the distribution of source timestamp
  distances rather than a pass or fail, so a future report of this shape can be
  characterised quickly. Opt in via `IIOT_JITTER_DIAG=1`.
- `soak.yml` gains the new scenarios. The pull request smoke job filters on
  `Category=LongRunning` and picks them up automatically.
- Docs: a warning on the legacy behavior in `readme.md`, and a new
  *Jittered timestamps without duplicates* section in `troubleshooting.md`
  covering the case where the completeness check passes but the spacing does
  not, with remedies in order of preference. Also a note at the top of the
  troubleshooting guide that it always describes the latest release, so that
  individual sections do not have to give version relative advice.


docs/opc-publisher/troubleshooting.md
```diff
@@ -177,6 +178,25 @@ To rule out that values were actually dropped or that the server queue overflowe
...
+- but a share of the `SourceTimestamp` distances is far from the sampling interval,
+- and occasionally a timestamp is *earlier* than the previous one, which looks like reordering.
+
+This is the [legacy behavior](./readme.md#legacy-behavior) working as designed: it shifts the re-sent timestamp forward by the time elapsed since the value was received, measured on the OPC Publisher clock. Every heartbeat therefore lands at an arbitrary point inside the sampling period and carries a unique timestamp.
+
+Measured against a perfectly regular simulated source at a two second sampling and two second heartbeat interval, this produced roughly 29 % of samples as heartbeats and roughly 29 % of `SourceTimestamp` distances outside a 2 % tolerance band, including negative distances - while the value sequence itself was complete and gap free.
+
+Remedies, in order of preference:
+
+1. Upgrade to a version that contains the [watchdog grace period](./readme.md#watchdog-grace-period). With it, a heartbeat configured at or below the publishing interval no longer races the value it is waiting for, and stops firing while data is flowing.
```
marcschier: This does not exist and it is always assumed that the user uses the latest version.  Add to the top of troubleshooting (if not exist) that user should use the last released version.

marcschier: Fixed in 1bd29b959.

Dropped the "upgrade to a version that contains the watchdog grace period" remedy, and added a note at the top of the troubleshooting guide instead:

> IMPORTANT: This documentation always describes the latest released version of OPC Publisher. Before investigating any issue, make sure you are running the [latest release](https://github.com/Azure/Industrial-IoT/releases). Issues are only diagnosed and fixed against it, and a problem you are seeing may already be resolved.

That belongs at the top rather than in this one section, since it applies to every symptom in the guide.

I removed the measurement of the pre grace period behaviour from the same section for the same reason - it described something the latest version does not do. Replaced it with what the current version actually does: the watchdog only fires once a value has missed the heartbeat interval plus the grace period, so while data flows normally no timestamp is displaced, and seeing this symptom points at late values rather than at a defect. Added a pointer to compare `ingressHeartbeats` against `ingressValueChanges` to quantify that.

The remaining remedies are now version independent: switch to the default `WatchdogLKV`, raise the heartbeat interval above the sampling interval, filter on the `Heartbeat` indicator, or analyze the message `Timestamp` instead.

